PR #2, opened 2016-11-30
Moved Che to demo.che.ci.centos.org

l0rd:
Change made as per @kbsingh request


Katka92:
comment from terminal

PR #23, opened 2017-03-16
Add support for cluster up and remove cluster-admin

vpavlin:
This PR adds support for using `oc cluster up` instead of `minishift`

It also changes use of `cluster-admin` to just `admin` for the project, which is enough to get workspace running and is close to what we will be able to get from OSO


ibuziuk:
Closing, since it is pretty much outdated

PR #29, opened 2017-04-06
137 integrate baysian lsp

tsmaeder:
This PR adds plugins for integratin the bayesian language server to this repository. The plugins are not yet integrated into any build that might go on here.


plugins/ls-bayesian-agent/pom.xml
```diff
@@ -30,4 +30,17 @@
             <artifactId>che-core-api-agent-shared</artifactId>
         </dependency>
     </dependencies>
+    <build>
+    	<plugins>
```
amisevsk: Mixed tabs and spaces here

sunix:
Outdated, replaced with https://github.com/redhat-developer/rh-che/pull/39

PR #32, opened 2017-04-13
Che 176 RedHat Che distribution build

davidfestal:


README.md
```diff
@@ -1,6 +1,79 @@
...
+
+##### Build using a local upstream Che
+
+This allows using a local che repository you already have on your machine.
+
+    mvn -DlocalCheRepository=<root of your local upstream Che Git repo> clean install
+
+##### Enabling / Disabling the Dashboard
+
+By default the Dashboard is part of the RedHat Che distribution.
```
davidfestal: Well, the default for the local developer would be *with* dashboard, because without a dashboard it's not easy to test.
But anyway, after discussing with @l0rd , it was decided that the CI build will create docker images for the 3 cases:
- Upstream che
- RH distribution with dashboard
- RH distribution without dashboard

assembly/assembly-ide-war/.classpath
```diff
@@ -6,24 +6,25 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry excluding="**" kind="src" output="target/classes" path="src/main/resources">
```
davidfestal: yeah, we'll remove them in a new PR as you proposed.

assembly/assembly-ide-war/.settings/org.eclipse.core.resources.prefs
```diff
@@ -1,5 +1,2 @@
 eclipse.preferences.version=1
-encoding//src/main/java=UTF-8
```
davidfestal: yeah, we'll remove them in a new PR as you proposed.

assembly/assembly-ide-war/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,5 +1,5 @@
 eclipse.preferences.version=1
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
```
davidfestal: yeah, we'll remove them in a new PR as you proposed.

assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -1,102 +1,130 @@
...
-            <artifactId>h2</artifactId>
+            <groupId>com.redhat.che</groupId>
+            <artifactId>che-keycloak-plugin-server</artifactId>
         </dependency>
         <dependency>
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che</groupId>
```
davidfestal: Well, no this dependency isn't enough, because the RH war overriden Java classes depend on these dependencies, and depending on the upstream `war` for overlays doesn't bring transitive dependencies.
We could use a `pom`  dependency on the upstream `war`  project, but this currently makes the dependency-checking unhappy (we already tried with @gorkem  So we kept it like that.

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterAnalyticsAddresser.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012-2016 Codenvy, S.A.
```
davidfestal: Well, depending on the upstream `war` for overlays doesn't bring transitive dependencies.
see [this comment](https://github.com/redhat-developer/rh-che/pull/32/files#r111462819)

assembly/pom.xml
```diff
@@ -1,34 +1,204 @@
...
+        <relativePath>../invoker/rh-changes</relativePath>
     </parent>
-    <groupId>com.redhat.che</groupId>
     <artifactId>almighty-ide-assembly-parent</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>${redhat.che.version}</version>
     <packaging>pom</packaging>
     <name>Almighty IDE assembly (parent)</name>
     <modules>
+        <module>dashboard</module>
```
davidfestal: The dashboard itself has to be patched (and even rebuilt fully with patched sources) because of the keycloak integration.

assembly/assembly-ide-war/pom.xml
```diff
@@ -463,4 +278,44 @@
             </plugin>
         </plugins>
     </build>
+    <profiles>
```
davidfestal: If I understand correctly, to do what you propose I would have to add directly all (but the `dashboard-client`) transitive dependencies of the upstream `assembly-ide-war` project to the main (non-profile) `<dependencies>` list. I don't see how it would make the build easier to read and maintain.

README.md
```diff
@@ -1,6 +1,79 @@
...
+
+This checks out and builds the upstream Che before building the RedHat distribution.
+
+    mvn clean install
+
+##### Default build - quick version (bypass upstream Che)
+
+This allows reusing a previously checked-out and built upstream Che.
+    
+    mvn --activate-profiles=-checkout-base-che clean install
```
davidfestal: Question of taste :-)

Well, the need for quotes around `-P!checkout-che` to avoid the following shell error:
```
bash: !checkout-base-che: event not found
``` 
looked also a bit weird in my eyes...

assembly/assembly-ide-war/.classpath
sunix: I suggest to revert these changes for this PR as git fixup and rebase autosquash before merging. It's just to have clean commits.

README.md
```diff
@@ -1,6 +1,79 @@
...
+- The Keycloak integration
+- The Bayesian integration (soon)
+
+#### How the RedHat Che distribution build works
+
+The default behavior is to:
+- checkout the upstream GitHub `che-dependencies` and `che` repositories into the
+`target/export`, based on a given fork (`eclipse` by default) and branch
+(`openshift-connector` by default)
+- build the upstream repositories first
```
sunix: For me, it doesn't make sense. It's like we don't have maven ... checkout and build the other repository should be an option not the default behaviour. The default behaviour is to rely on maven to get the dependency.

davidfestal: > For me, it doesn't make sense. It's like we don't have maven ... checkout and build the other repositories should be an option not the default behaviour. The default behaviour should be to rely on maven to get the dependencies.

well, it seems I wasn't clear enough in the `Readme`: what it means really is that *by default*, maven *automatically takes care of everything as part of the root maven build*: it checks out the upstream che, builds it as part of the invoker *setup* step, and then runs the RH assembly sub-build.
I'll update the `Readme` to clarify this point.

README.md
davidfestal: OK, right, since I also like the `!`, that makes it clear that we *don't want* to checkout the upstream Che again, let me change to use it, as well as the more usual `-P` option.

assembly/assembly-ide-war/pom.xml
sunix: in the following profile, you define 2 profiles where only one is necessary. And it's kind of duplicating things

davidfestal:
> We need to stop using Almighty and change those to Fabric8. When I started the original assembly almighty was a thing not anymore.

@gorkem I'll update the names with `fabric8`, also the version suffixes for the generated artifacts. I'll do it also in PR https://github.com/kbsingh/build-run-che/pull/54

README.md
```diff
@@ -1,6 +1,80 @@
...
+
+The RedHat Che distribution build is a *maven* build. By default, it *automatically* takes care 
+of *everything* as part of the root maven build, which means that it:
+- checks out the upstream GitHub `che-dependencies` and `che` repositories into the
+`target/export`, based on a given fork (`eclipse` by default) and branch
+(`openshift-connector` by default),
+- builds the upstream repositories first as a pre-step,
+- then builds the RedHat distribution maven sub-project based on this upstream build.
+
+However, by passing a given propetrty, it is also possible to bypass the checkout and build
```
l0rd: s/propetrty/property

PR #34, opened 2017-04-18
Added definition files for setting minishift quotas

sbryzak:
Signed-off-by: Shane Bryzak <sbryzak@redhat.com>

### DIFF
--- a/quotas/README.MD
+++ b/quotas/README.MD
@@ -1,0 +1,41 @@
+These definition files allow quotas to be set in a local minishift environment that mimic the quotas available on the free-int cluster.
+
+To apply the quotas:
+
+1. Login as the system user:
+
+  oc login -u system:admin
+  
+2. Apply the quotas:
+
+  oc create -f compute-resources.yaml -n eclipse-che
+  oc create -f compute-resources-time-bound.yaml -n eclipse-che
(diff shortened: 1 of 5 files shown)


amisevsk:
This could be useful for quickly setting up a test environment on minishift, but it does duplicate some files from the fabric8-online repo. Since the fabric8-online repo is already a dependency in the `create-all.sh` script, it might be easier to have a new script call `oc apply -f` on those files instead of maintaining two parallel copies.

@l0rd WDYT?

l0rd:
Yes @amisevsk is right.

@sbryzak have a look at https://github.com/fabric8io/fabric8-online/pull/121

But yes, I think we should document it in the root README.md. In my notes I have this snippet to setup Che namespace with quotas:
```
export CHE_OPENSHIFT_PROJECT=mario-loriedo-che
oc login -u developer -p developer
oc new-project ${CHE_OPENSHIFT_PROJECT}
oc login -u system:admin
oc apply -f ~/github/fabric8-online/packages/fabric8-online-che/target/classes/META-INF/fabric8/openshift.yml
```

@sbryzak can you adapt that and add it as a new section (for example `Setup Che namespace (optional)`) in root README.md

l0rd:
Fixed

PR #35, opened 2017-04-19
Replace $CHE_HOSTNAME with $OPENSHIFT_NAMESPACE_URL

l0rd:
This is related to https://github.com/eclipse/che/pull/4831


ibuziuk:
@l0rd +1, but I think README.md should be also updated

l0rd:
@ibuziuk yes you are right. I'm going to update the README as well

l0rd:
@ibuziuk README updated

PR #36, opened 2017-04-19
Use an env variable to set Che version used in create_all.sh

l0rd:
Add a new environment variable ($CHE_SERVER_IMAGE) to configure the version of Che image that will be deployed with `create_all.sh`.


l0rd:
We can close this one since it's already possible to set the Che image and tag when deploying via https://github.com/redhat-developer/rh-che/blob/master/dev-scripts/minishift_deploy.sh

PR #37, opened 2017-04-19
Use the right memory limits in create-all.sh

l0rd:


l0rd:
@amisevsk that's a good idea. That not something we need now (we want to make sure we test in OSO conditions) but that would be something we may add in the future. But the env var should be called `CHE_WORKSPACES_MEMORY_LIMIT` rather than `CHE_SERVER_MEMORY` right?

PR #39, opened 2017-04-19
137 integrate baysian lsp

sunix:
This PR adds plugins for integrating the bayesian language server to the RH Che assembly. this PR requires  https://github.com/eclipse/che/pull/4846

demo: https://youtu.be/tflGMIbyIpE

Replace https://github.com/redhat-developer/rh-che/pull/29


invoker/rh-changes/pom.xml
```diff
@@ -41,6 +41,18 @@
                 <type>tar.gz</type>
             </dependency>
             <dependency>
+                <groupId>org.eclipse.che.plugin</groupId>
```
davidfestal: Here the `groupId` should be  `com.redhat.che` IMO (see the review comment on https://github.com/redhat-developer/rh-che/pull/39/files#diff-a0951a27f025f1f7de5ead9e60153ab7).

As for the version, there's no reason why we should use the upstream Che version. No reason, as well, that it would follow the version of the RH distribution that derives from the upstream Che version. Beyesian plugins, like Keycloak plugins, are RH-only standalone components (added into the RH Che distribution) that can have their own version.
So it seems to me that we should add a property:
```
        <bayesian.plugins.version>1.0.0-SNAPSHOT</bayesian.plugins.version>
```
as done for the keycloak plugins, and use it here in the dependency management.

plugins/ls-bayesian-agent/src/main/resources/com.redhat.bayesian.lsp.script.sh
```diff
@@ -0,0 +1,155 @@
+unset PACKAGES
+unset SUDO
+unset PAYLOAD
+
+command -v bzip2 >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" bzip2"; }
+command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
+command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+
+test "$(id -u)" = 0 || SUDO="sudo"
```
l0rd: @sunix That won't work on OpenShiftOnline (`sudo` is not allowed). Please refer to agents scripts in `openshift-connector` branch (for example https://github.com/eclipse/che/blob/openshift-connector/agents/ls-typescript/src/main/resources/org.eclipse.che.ls.typescript.script.sh#L16).

l0rd:
@sunix we should make sure that this work on OpenShift before merging.

plugins/ls-bayesian-agent/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Codenvy, S.A. - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
```
sunix: should be fixed by https://github.com/redhat-developer/rh-che/pull/39/commits/5e3b835eea1471d38512803343364ac7651bf212

plugins/che-plugin-bayesian-lang-server/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Codenvy, S.A. - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
```
sunix: should be fixed by https://github.com/redhat-developer/rh-che/pull/39/commits/5e3b835eea1471d38512803343364ac7651bf212

plugins/ls-bayesian-agent/pom.xml
davidfestal: Thanks! I approved the changes.

plugins/che-plugin-bayesian-lang-server/pom.xml
davidfestal: Thanks! I approved the changes.

PR #42, opened 2017-04-23
Change the way how and when to install new version of bayesian lsp server

msrb:
- make sure the target directory is clean and there are no leftovers
from previous installations
- fetch the tarball directly from GitHub
- only upgrade if current version differs from the latest stable

Signed-off-by: Michal Srb <michal@redhat.com>


msrb:
tested locally on minishift: 
screenshot from 2017-04-23 12-30-35

msrb:
I think the most important change from this PR is the new location from where to fetch the tarball. Still not ideal, but tarballs at GitHub are actually hosted on AWS S3, so downtimes should be pretty rare. The previous location was only good for development/testing.

Note more I think about how the release process and delivery of the lsp server should look like, more I am inclined to actually embedding it inside `com.redhat.che:ls-bayesian-agent`. IMO, being able to control which version exactly is running on which deployment beats the ability to do independent releases. WDYT?

msrb:
After checking how other agents do this and discovering that all the shell scripts look pretty much alike, I am closing this PR in favour of much simpler https://github.com/redhat-developer/rh-che/pull/45.

PR #44, opened 2017-04-24
Include the checks / tests by default, but allow skipping them...

davidfestal:
... also add multi-thread builds as an `opt-in` option.


davidfestal:
@l0rd is it what you were expecting ?

PR #45, opened 2017-04-24
Download bayesian lsp server from GH

msrb:


msrb:
Note I've tested these changes locally on minishift.

PR #49, opened 2017-04-26
Set workspace timeout to 1h

snjeza:
https://issues.jboss.org/browse/CHE-233


l0rd:
Thank you @snjeza ! Can you please just commit the patch so we don't have to duplicate `che-machine-configuration.properties` in 2 distinct repositories?

We are already doing that for other files (c.f. https://github.com/redhat-developer/rh-che/blob/master/assembly/assembly-ide-war/src/main/patches/src/main/webapp/IDE.jsp.patch)

snjeza:
@l0rd I have updated the patch.

l0rd:
Hi @snjeza we could not merge this before summit demo. If you rebase it now we will be able to merge it.

PR #50, opened 2017-04-26
Change SSO addresses to prod

gorkem:
Changes the SSO address to prod and relax
the refresh timer


gorkem:
managed to push to the correct branch 😕

PR #51, opened 2017-04-26
Bump version of bayesian lsp server to 0.0.4

msrb:


msrb:
The only change is that if there is an environment variable "RECOMMENDER_API_TOKEN", then its value will be sent in "Authorization" header of the HTTP request.

PR #53, opened 2017-04-28
Provide the ability to skip modules of the RH distribution build

davidfestal:
https://issues.jboss.org/browse/CHE-241

This required moving some directories: `assembly` and `plugins` are now sub-directories instead of being at the root. But the resulting build is more consistent and robust, and allows rebuilding only some modules of the RH build, which should make the life of developers much easier.

I also completed the documentation with the new developer-friendly options 

I'll create the corresponding PR for the CI build to update the paths of the generated distribution artifacts: 


davidfestal:
@l0rd @gorkem I'm currently testing this in CI locally. I'll let you know as soon as it's OK

davidfestal:
@l0rd @gorkem I tested the CI with these PRs that built successfully and pushed the following tag 20fdeed-fabric8-296870d on https://hub.docker.com/r/dfestal/che-server/tags/. I tested the pushed image and it seems to work well also.

l0rd:
LGTM @davidfestal let's wait summit to merge that and the correspondent PR on build-run-che

sunix:
If we want to keep the invoker, couldn't we use alternate invoker.properties files http://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html#invokerPropertiesFile ? If I understand correctly, this adds the ablility to have multiple invoker.properties files: developers could either choose to use predefined (CI or defaultdev) or customize its own to skip tests etc ...
To me, adding more custom options will make the README.md even bigger and harder to read.

davidfestal:
@sunix 

> To me, adding more custom options will make the README.md even bigger and harder to read.

Did you have a look at the changes to the readme in this PR ? First I tried to make it clearer what the default developer command line /options are. Then I added the options to build only sub-modules, so it's bigger yes, but it doesn't seem to me this makes the overall Readme *harder to read*.

> couldn't we use alternate invoker.properties files http://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html#invokerPropertiesFile ? 

we already use the invoker properties file to forward the standard reactor options from the main build to the `rh-che` sub-build. And the properties to pass have the same name / behaviour as the standard multi-module reactor options. So they are not hard to remember. And I added the *small* variant also ;-)

> developers could either choose to use predefined (CI or defaultdev) or customize its own to skip tests etc ...

I don't find this *easier for the developer* to have to *learn* the syntax of this invoker properties file than simply use properties that have the same name as standard maven reactor options.

sunix:
Can we also skip upstream che modules during the build ?

>Did you have a look at the changes to the readme in this PR ? First I tried to make it clearer what the default developer command line /options are. Then I added the options to build only sub-modules, so it's bigger yes, but it didn't seem to me this has made the overall Readme *harder to read*.

Yes, the longer it is, the harder it is to read and figure out how to build.

>> developers could either choose to use predefined (CI or defaultdev) or customize its own to skip tests etc ...
> I don't find this *easier for the developer* to have to *learn* the syntax of this invoker properties file than simply use properties that have the same name as standard maven reactor options.

Using predefined as template don't need any learning curves.

davidfestal:
> Can we also skip upstream che modules during the build ?

I've had this idea also and would probably take some look into it when working on https://issues.jboss.org/browse/CHE-232

PR #54, opened 2017-05-01
Start using own stacks

gorkem:
Adds a stacks component to rh-che that overrides the
upstream stacks.


invoker/rh-changes/pom.xml
```diff
@@ -82,6 +82,12 @@
                 <version>${redhat.che.version}</version>
                 <type>war</type>
             </dependency>
+            <dependency>
+                <groupId>com.redhat.che</groupId>
+                <artifactId>fabric8-ide-stacks</artifactId>
+                <version>${redhat.che.version}</version>
```
davidfestal: Is there a reason why the stacks version should be `${redhat.che.version}` ?
I mean `${redhat.che.version}` is following the upstream Che version with a suffix (`fabric8` or `fabric8-without-dashboard`). It has sense for the RH components that *derive from* the upstream components (assemblies, dashboard) but with modifications / additions.
On the contrary, components that exist on their own (such as  additional RH plugins) can have their own versioning.  How are these stacks going to be managed ? are they going to be regularly updated from their upstream counterparts, or are they going to have their own life ?

l0rd:
That's the good approach but I have some comments regarding this implementation:

1. We are including `che-in-che` stack and I'm not sure that's we want that: currently this stack is not going to run on OSO 
2. Instead of duplicating files from eclipse/che (files that we will need to sync manually) shouldn't we just patch `stacks.json` as it's already done for other upstream files (e.g. [subversionApiITest](https://github.com/redhat-developer/rh-che/blob/master/invoker/base-che/src/main/patches/subversionApiITest.patch))

What do you think?

invoker/rh-changes/pom.xml
l0rd: @davidfestal we should consider it as a customization of upstream che. So it makes sense to use upstream version. 

But I do think that since it's a customization of a `json` file it could be implemented as a patch rather than a maven module (c.f. [my comment](https://github.com/redhat-developer/rh-che/pull/54#issuecomment-298862849)).

gorkem:
I have left `che-in-che` because dog fooding is a priority. I do not think we need to patch stacks.json but have our own. Also I am not a big fan of the patching that we are doing during builds, it is error prone IMHO.

l0rd:
I agree that using patches is error prone. But having duplicated code can lead to more subtle errors that can be discovered long time after they have been introduced. 

For example if a tag, a version of a component or the name of an agent of these stacks are changed upstream we will need to remember to manually apply the change downstream. If we forget we won't  notice it at build time: that's something that will show up at runtime only. And if we don't test every stack on OSIO (we currently don't) the first ones to notice these subtle errors will be users. In production.

With patches the build will get broken every time an update breaks a patch but at least we know it as soon as we try to build rh-che.

l0rd:
> I have left che-in-che because dog fooding is a priority. 
In makes sense to add `che-in-che` stack in that case. We will need to add the image to https://github.com/redhat-developer/che-dockerfiles

sunix:
In a long term view and implementation. I would see a stack-fabric8.json that would override upstream one in a smart way, pretty much like an Eclipse extension point:
1: all the stacks that are in stack-fabric8.json would override the stacks with the same stackid,
2. other stacks would stay unchange
3. new stacks in stack-fabric8.json would be added to the list

Same kind of behaviour would happen to the stacks.json that is created once in Che data folder: would get updated following these kind of rules.

gorkem:
@sunix Unless we have a good way to test and verify all the upstream image changes on openshift.io we should not include any stack from there.

invoker/rh-changes/pom.xml
gorkem: @davidfestal what is the property that I should use here?

davidfestal: since it's a customization of upstream che, I think using `${redhat.che.version}` is good.

sunix:
@gorkem it looks good to me then :)

sunix:
I've reverted the stacks.json on upstream che and it doesn't seems to work. (taking upstream stacks.json).
You can try:

    mvn clean install -DwithoutKeycloak -Pfast -PmultiThread -Dche-fork=sunix -Dche-branch=fix-route-change-rebased-without-stacks

I tested on minishift

l0rd:
When fixing this PR we need to take into account that we want to build 2 versions of rh-che:

- fabric8-ide: redhat maintained stacks only but no bayesian agent 
- osio-ide: fabric8-ide + bayesian agent started by default

Building fabric8-ide is important to be able to deploy Che on non-osio OpenShift clusters (i.e. minishift)

sunix:
@l0rd I think creating new variant of the assembly should be part of another/a new task

l0rd:
@sunix I agree. But we discussed about different solutions to fix this PR. And when choosing the solution we should take into account that the patch of `stacks.json` will be probably be splitted in 2 steps:

1. For fabric8-ide: filtering out non supported stacks
2. For osio-ide: adding bayesian agents 

Does that makes sense?

l0rd:
Has already been merged via https://github.com/redhat-developer/rh-che/commit/39ebdb17fa3b3d8890882546f3b2ce2b6d3f27e2#diff-d1904fe33bc14912fc2e4d5a15a5bc99

PR #55, opened 2017-05-02
Update URL for fabric8-analytics LSP server

msrb:


msrb:
The LSP code now lives in https://github.com/fabric8-analytics/fabric8-analytics-lsp-server. All further development will take place there. The tarball is exactly the same as the previous one.

msrb:
Rebased :wink:

PR #56, opened 2017-05-03
Add script for building redhat dist image

amisevsk:
Add script `rh-build.sh` which builds docker plugin from main repo, runs mvn install in root `rh-che` repo, and then builds a docker image.

Also modified create-all and build.sh to support `CHE_IMAGE_TAG`s other than "nightly".


sunix:
+1 for modified create-all and build.sh to support CHE_IMAGE_TAGs other than "nightly".
-1 for adding another script to build:
we have the new maven invoker that is kind of scripting che and rh-che maven projects. What your script is doing could be done in this maven project ( except for the docker part) I would use a profile in maven that would have all your options.

amisevsk:
@sunix I don't know that a profile would be really useful here: the point of the script is to ideally build a minimal set of components, and only works if you are working on the docker/openshift plugin specifically. Once development focus in upstream Che shifts away from just OpenShift connector, the script will be useless. There are too many different permutations of requirements to codify all of them as profiles. The real issue is that there's no good way to automatically do all the necessary component builds when working on one specific area.

@davidfestal Yeah this PR is more of a holdover to just enable somewhat easier development. The other `build.sh` is really useful for small changes to e.g. OpenShiftConnector (since it takes ~30 seconds on my local machine). Once we have common scripts this will not be needed anymore.

sunix:
@amisevsk 
1. rh-che has the ability to rebuild che upstream. the build script does the same, it's kind of duplicating things
2. in upstream che, use `mvn clean install -Pfast` does the same thing your `mvnche` does
3. in rh-che, I'd like to add to add a new profile `-Pdev` that would do what `mvnrh`does and be activated by default. With the invoker the profile could even add the `-Pfast` to upstream che
4. if you still want to run mvn in build.sh, i'm ok but i would try to add the minimum of logic there. Build command in build.sh should look like `mvn clean install -Pfast`.

amisevsk:
@sunix 
1. I don't see where invoking `mvn clean install` in `rh-che` rebuilds the docker plugin. 
2. IIRC `-Pfast` was slightly different from the `mvnche` command we're using; that's why it's in the `build.sh` script as well. I could be wrong though.
3. The main purpose of `build.sh` is to rebuild one specific set of modules. Ideally, running rh-build.sh would rebuild rebuild the docker plugin from upstream, then wsmaster assembly, then main assembly. This is not a full build, on purpose, since this subset of modules is much quicker to build than a default `mvn clean install -Pfast`. 
4. I use this script because as far as I can tell, there is currently no simple path to developing changes in upstream che at the moment. Building just upstream results in a broken che, since workspace launch depends on the bayesian plugin, and building just rh-che does not pull in upstream changes. Also, I want a local docker image with my changes that I can use in minishift and don't want to manually link our assembly into the upstream repo. I'm not saying this is a permanent script, but I don't see how anyone is making progress in upstream Che right now without a script similar to this one.

sunix:
1. it is done here: https://github.com/redhat-developer/rh-che#build-using-a-local-upstream-che (see  https://github.com/redhat-developer/rh-che#build-using-a-local-upstream-che
2. it is in `build.sh` because at the time, the `-Pfast` wasn't in upstream yet. it has been introduced in January or February 2017 as far as I remember
3. The problem with build.sh is that it builds only plugin-docker modules + assembly. But in the openshift-connector branch, we have more plugins and modules involved. So what you are saying is that build.sh is just for people working only on plugin-openshift-client ?
4. I'm just saying that what you do in the script (pulling, building a local upstream che) could be done at the rh-che maven project level and so it makes sense to do it there and not adding one more additionnal layer. We are already doing maven in maven with this build, and with this script, it looks like maven in maven in bash.

scripts/rh-build.sh
```diff
@@ -0,0 +1,71 @@
...
+cd ${SCRIPT_DIR}/../
+mvnrh "$@" install
+
+cd ${GITHUB_REPO}/dockerfiles/che/
+mv Dockerfile Dockerfile.alpine
+cp Dockerfile.centos Dockerfile
+
+# Link custom assembly and build docker image
+source ${GITHUB_REPO}/dockerfiles/build.include
+BUILD_ASSEMBLY_ZIP=$(echo "${SCRIPT_DIR}"/../assembly/assembly-main/target/eclipse-che-*fabric8*.tar.gz)
```
davidfestal: @amisevsk  after PR #53, you should update the path here  to `${SCRIPT_DIR}"/../target/builds/fabric8/fabric8-che/assembly/assembly-main/target/eclipse-che-*fabric8*.tar.gz`

amisevsk: Thanks, will do!

amisevsk:
Closing PR as it seems unnecessary at the moment.

PR #57, opened 2017-05-04
Add binding to OpenShiftWorkspaceFilesCleaner in WsMasterModule

amisevsk:
Looks like we missed a binding in WsMasterModule on upstream Che. OpenShiftWorkspaceFilesCleaner is necessary to avoid NPEs and properly delete files in PVs.

See: https://issues.jboss.org/browse/CHE-181


amisevsk:
@davidfestal Fixed! Thanks.

PR #61, opened 2017-05-05
README improvements: TOC, TLDR for build

l0rd:


README.md
```diff
@@ -1,37 +1,63 @@
...
-- the ability to disable the Dashboard (and remove the *Go to Dashboard* button from the Che IDE)
-- The Keycloak integration
-- The Bayesian integration (soon)
+- The ability to disable the Dashboard (and remove the *Go to Dashboard* button from the Che IDE)
+- Keycloak integration
+- [fabric8-analytics Language Server](https://github.com/fabric8-analytics/fabric8-analytics-lsp-server) 
 
 #### How the RedHat Che distribution build works
 
-The RedHat Che distribution build is a *maven* build. By default, it *automatically* takes care
```
davidfestal: At the beginning, the initial test was somewhat similar to the one you propose. Then I tried to make it clearer and more explicit due to a comment:
https://github.com/redhat-developer/rh-che/commit/bbeb548236003ce0904133f1e039bb6a4b382d4e
It seemed it was needed to make even clearer clear that the checkout of the upstream project is *automatic*, and that this is *only by default*.
   
Don't know if this insistance is still relevant.

Anyway for me it's OK in boith cases, since we tell it just after (*However ...*)

l0rd: @sunix is it clear for you?

README.md
```diff
@@ -1,37 +1,63 @@
...
 
+### TL;DR
+
+There is a good chance that you want to build rh-che with the dashboard but without keycloak support (because it's easier to test) and as fast as possible :bowtie::
+
+```
+git clone https://github.com/redhat-developer/rh-che
+cd rh-che
+mvn -DwithoutKeycloak `# disable Keycloak support` \
+    -Pfast            `# skip tests and other verifications` \
```
amisevsk: What about `-PmultiThread`?

README.md
```diff
@@ -1,37 +1,63 @@
...
+    -DlocalCheRepository=${UPSTREAM_CHE_PATH}  `# get already built upstream Che` \
+    clean install
+```
+
 #### What is the RedHat Che distribution
 
-The RedHat distribution of Eclipse Che is a RedHat-specific packaging of the Che assemblies
-that allows adding some RedHat specific plugins / behaviors upton the standard upstream Che
+The RedHat distribution of Eclipse Che is a RedHat-specific packaging of Che assemblies
+that adds some RedHat specific plugins / behaviors upton the standard upstream Che
```
amisevsk: that adds some RedHat specific plugins / behaviors **on top of** the standard upstream Che

README.md
sunix: @l0rd if we just describe what the build does it is clearer. I don't see the point in adding more information that we already know: 

> ... is a *maven* build

Seems obvious for any java developer

> it *automatically* takes care of *everything* as part of the root maven build, which means that it

This is adding more confusion. @davidfestal what would add more useful information could be how the build works with the maven invoker plugin.

So :+1: for your change @l0rd

PR #63, opened 2017-05-06
CHE-250 - Use -XX:+UseStringDeduplication VM argument

snjeza:
https://issues.jboss.org/browse/CHE-250 - Use -XX:+UseStringDeduplication VM argument when starting che server and workspace

### DIFF
--- a/scripts/create-all.sh
+++ b/scripts/create-all.sh
@@ -44,8 +44,8 @@ oc create configmap che \
       --from-literal=workspaces-memory-limit="1300Mi" \
       --from-literal=workspaces-memory-request="500Mi" \
       --from-literal=enable-workspaces-autostart="false" \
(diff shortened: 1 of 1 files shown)


snjeza:
@gorkem 
I have tested on minishift 1.0.0, oc 1.5.0 and used 'docker stats' to review the memory usage.
I have imported the console-java-simple and che-starter project, build, run and test them.
The memory usage is changed when running an action in a workspace, but it is stabilized when a workspace is not used for some time.
The following are the results:
a) a che server with the -XX:+UseG1GC -XX:+UseStringDeduplication argument uses about 570 MB of RAM, a workspace uses less then 500MB
b) without the -XX:+UseG1GC -XX:+UseStringDeduplication argument, a che server uses about 670 MB, a workspace uses more then 630MB

The argument is added only to the ws agent, not to other agents in a workspace. The argument is used when building, running or testing projects.

snjeza:
@l0rd https://github.com/fabric8io/fabric8-online/pull/220

PR #65, opened 2017-05-11
Exported docker image name for CI jobs

mlabuda:
This PR creates a file with docker image name stored in variable. It is used for CI jobs and passing vars between jobs.

### DIFF
--- a/cico_build.sh
+++ b/cico_build.sh
@@ -21,6 +21,7 @@ systemctl start docker
 
 currentDir=`pwd`
 
+chmod a+x ..
 chown -R ${BuildUser}:${BuildUser} ${currentDir}
 
 buildCommand="cd ${currentDir} && ./build_che.sh $*"


cico_build.sh
```diff
@@ -75,7 +75,12 @@ if [ $? -eq 0 ]; then
     # We are not pushing the nightly tag because we don't need it and CI has an issue
     # when publishing > 1 tag at a time 
     # docker push ${DOCKER_HUB_NAMESPACE}/che-server:${NIGHTLY}
-    docker push ${DOCKER_HUB_NAMESPACE}/che-server:${TAG}
+    export CHE_SERVER_DOCKER_IMAGE=${DOCKER_HUB_NAMESPACE}/che-server:${TAG}
+    cd
+    env_var_file=che.image.env
+    touch $env_var_file
+    echo 'export CHE_SERVER_DOCKER_IMAGE='${CHE_SERVER_DOCKER_IMAGE} > $env_var_file
```
l0rd: I'm afraid that  because of the `cd`, at next loop iteration, commands that depend on path (line 62 and 63) won't work anymore. And these 5 lines can be replaced with one line only:
```sh
echo "export CHE_SERVER_DOCKER_IMAGE=${DOCKER_HUB_NAMESPACE}/che-server:${TAG}" > ~/che.image.env
```

mlabuda: updated

mlabuda:
Namespace changed, but are the creds same for both repos?

PR #68, opened 2017-05-12
Replaced deploying of Che by verification steps of Che deployment to OSIO

mlabuda:


cico_deploy.sh
```diff
@@ -6,56 +6,55 @@ set +e
...
-    echo 'Deploy Failed!'
-    exit 1
-  fi
-
-  # Create route
-  oc expose service che-host --hostname=${CHE_OPENSHIFT_HOSTNAME}
-
+OSO_TOKEN=$(curl -sSL -H "X-Vault-Token: ${RHCHEBOT_DOCKER_HUB_PASSWORD}" -H "Content-Type: application/json" \
+	-X GET http://li546-232.members.linode.com:8200/v1/secret/os_token | cut -d\" -f18)
+OSO_USER=mario-loriedo
```
l0rd: We should change the value to `mloriedo`. The test account that we are going to use is `mloriedo`

mlabuda:
Ok, done. Please review.

PR #77, opened 2017-05-23
Update README.md: skip upstream build upstream che

sunix:
Update sample build command, adding one more option to skip build of upstream che project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ UPSTREAM_CHE_PATH=/path/to/upstream/che
 mvn -DwithoutKeycloak                          `# disable Keycloak support` \
     -Pfast                                     `# skip tests and other verifications` \
     -PmultiThread                              `# enable maven multi threading` \
-    -DlocalCheRepository=${UPSTREAM_CHE_PATH}  `# get already built upstream Che` \
+    -DlocalCheRepository=${UPSTREAM_CHE_PATH}  `# get already built upstream Che local folder` \ 
+    -P '!checkout-base-che'                    `# skip checkout and build upstream Che` \
     clean install
 ```
 


README.md
```diff
@@ -28,7 +28,8 @@ UPSTREAM_CHE_PATH=/path/to/upstream/che
 mvn -DwithoutKeycloak                          `# disable Keycloak support` \
     -Pfast                                     `# skip tests and other verifications` \
     -PmultiThread                              `# enable maven multi threading` \
-    -DlocalCheRepository=${UPSTREAM_CHE_PATH}  `# get already built upstream Che` \
+    -DlocalCheRepository=${UPSTREAM_CHE_PATH}  `# get already built upstream Che local folder` \
```
davidfestal: I don't see why the `-P '!checkout-base-che'` is necessary here, since when using a local upstream Che repo (with `-DlocalCheRepository`) the upstream Che will never be checked-out from Git (to `target/export`) and rebuilt.
These 2 options are related to 2 different use-cases that are mutually exclusive.

l0rd: @sunix I was wondering too why we need to add  `!checkout-base-che`. When I run the old command che wasn't checked out. Can you be more explicit? Is it when che branch isn't `openshift-connector`?

sunix: 1. Specifying localCheRepository is still building upstream che in localCheRepository folder
2. Disabling the profile checkout-base-che still requires localCheRepository to be set ....

sunix: checkout-base-che is not only checking out but also building ....

davidfestal: > 1. Specifying localCheRepository is still building upstream che in localCheRepository folder

I never experienced this, and don't see, in the code, how it could behave this way. Could you give the precise command you run, in order to trigger the upstream che build in your local upstream Che repository ?

> 2. Disabling the profile checkout-base-che still requires localCheRepository to be set ....

Things are the other way round. Both main use cases are *with or without* `localCheRepository`:
    - when you build *without* `localCheRepository`, it checks out the upstream Che in `target/export` *and* builds it. But  you can remove the `checkout-base-che` profile to *skip the upstream checkout and build* step and only build the RH distribution, using the upstream that was already checked-out and built by a previous build in `target/export` 
    - when you build *with* `localCheRepository`, it *only builds the RH distribution* but, when it needs to access the upstream Che sources (e.g. GWT application or Dashboard for example), or the upstream Che generated artifacts (e.g. assembly-main), it will get them from the local upstream Che repository defined by the `localCheRepository` setting.

> checkout-base-che is not only checking out but also building ....

This profile is automatically selected when the `localCheRepository` property is not provided (the upstream repository should be checked-out when it is not provided). It is only useful when *removing it* to skip the checkout and build (see the previous remark).

sunix: OK it's too complicated for me, what is the command so it skips the build of upstream che ?

amisevsk: @sunix To my understanding, if you use `localCheRepository`, upstream won't be built and the build will use just artifacts that are in that folder (those generated by a previous build you did in upstream). This is how it works for me locally, but I could be wrong though.

davidfestal: @amisevsk yes, that's it.

sunix: @amisevsk @davidfestal which folders in localCheRepository ? Normally we should get already built artifacts from local maven repository (or remote nexus repo if not found). I don't quite understand the goal of `localCheRepository`

davidfestal: explained in comment https://github.com/redhat-developer/rh-che/pull/77#discussion_r118252410

> when you build with `localCheRepository`, it only builds the RH distribution but, when it needs to __*access the upstream Che sources (e.g. GWT application or Dashboard for example)*__, or the upstream Che generated artifacts (e.g. __*assembly-main*__), it will get them from the local upstream Che repository defined by the `localCheRepository` setting.

The RH build needs need both generated artifacts __*and*__ the upstream __*sources*__ according to the modules.

l0rd:
@sunix can we close this one?

sunix:
not needed anymore thanks

PR #78, opened 2017-05-23
Dev-oriented scripts to build and deploy the RH distribution to minishift, based on CICO script

davidfestal:
Developer-oriented scripts that allow building / deploying / rolling out the RH distribution to a local minishift installation:

- `dev_build.sh`:
    - run the RH maven build with the options passed as arguments,
    - generate the docker images for the upstream and the RH
distribution and tag them appropriately in the *local docker
environment*.

- `dev_build_and_deploy.sh`:
    - run the RH maven build with the options passed as arguments,
    - generate the docker images for the upstream and the RH
distribution and tag them appropriately in the *minishift docker
environment*
    - delete all the Che-related resources in minishift
    - create all the require resources in minishift based on the
generated RH docker image.
    
- `dev_build_and_rollupdate.sh`:
    - run the RH maven build with the options passed as arguments,
    - generate the docker images for the upstream and the RH
distribution and tag them appropriately in the *minishift docker
environment*
    - roll-update the installed Che in minishift with the generated RH
docker image.

PR #81, opened 2017-05-30
Add instructions to build rh-che docker image

l0rd:
In README.md we currently have only instructions to build che maven assembly. Nothing is said about how to deploy the assembly to OpenShift. This PR goal is to fill this gap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,12 @@
 
 ## Table Of Content
 
-* [How to build the RedHat Che distribution](#how-to-build-the-redhat-che-distribution)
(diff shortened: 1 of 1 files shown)


l0rd:
@davidfestal not sure if this doc will be needed after we merge #78. That depends on how much you are going to cleanup README.md. In any case let's merge #78 and then discuss that.

davidfestal:
I'm mostly changing the second part of the Readme. So it will be complementary IMO

humaton:
If this was part of the README week ago we would save some time. Current build instructions are not complete.

l0rd:
This PR can be closed because it has been outdated by other commits

PR #85, opened 2017-06-01
Remove Subversion support from the RH distribution

davidfestal:
Fix for https://issues.jboss.org/browse/CHE-245

### DIFF
--- a/builds/fabric8-che/assembly/assembly-ide-war/pom.xml
+++ b/builds/fabric8-che/assembly/assembly-ide-war/pom.xml
@@ -23,6 +23,12 @@
         <original-project-name>assembly-ide-war</original-project-name>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.redhat.che</groupId>
+            <artifactId>che-nosvn-plugin-ide</artifactId>
+        </dependency>
+    </dependencies>
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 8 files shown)


sunix:
To me, we should make upstream Che subversion plugin as easy to remove as removing a plugin. Upstream Che is a product, but it is overall a platform.

l0rd:
@davidfestal how hard you think it would be make svn a real plugin in Che upstream? Something we could just remove in our distribution, without the need of a fake plugin.

@sunix I agree with you but we must take into account how urgent it is to remove svn in osio

sunix:
@l0rd yes agree, we should create a task for this and plan it

davidfestal:
@l0rd well, The `SubversionCredentialsDialog` interface is injected into the standard project wizard (through constructor), in case Subversion is used. Adding a default `NoSvnSubversionCredentialsDialog` simply allows Google Gin no to fail. I assume that moving  this class (without binding it) into the upstream Che (in the SVN core API module) would avoid the need to provide it in a fake plugin. However the Gin binding of the `SubversionCredentialsDialog` to the `NoSvnSubversionCredentialsDialog` would still be needed in the RH assembly code.

Alternatively, removing the use of this interface from the project wizard code in the upstream che would certainly require a broader refactoring.

@sunix @l0rd I also think that should be done as part of a new issue. And when working on it, we might be able to remove the fake nosvn plugin from the RH assembly.

davidfestal:
@l0rd @gorkem Is it OK to merge this then ?

gorkem:
@davidfestal can you raise this up on upstream.

davidfestal:
> @davidfestal can you raise this up on upstream.

https://github.com/eclipse/che/issues/5271

davidfestal:
@gorkem @l0rd are you OK to merge this ?

l0rd:
Ok for me

davidfestal:
@gorkem is it OK for you too ?

gorkem:
@davidfestal yes, please go ahead.

sunix:
https://github.com/eclipse/che/pull/4490

PR #87, opened 2017-06-02
Make the Maven build compatible with M2Eclipse

davidfestal:
#### Description of the problem 

The RH distribution build has several strong requirements such as:
- Producing several variants of the RH Che distribution with a single build, while ensuring that the build process and intermediate artifacts of each variant are fully isolated,
- Deriving the version number of the produced RH Che distribution assembly artifacts (for all variants) from the version of the upstream Che,
- Reusing ( and modifying / completing ) *both the source and the generated artifacts* of some upstream Che maven modules (e.g. the `assembly-ide-war` GWT application) to provide the variants of the RH Che distribution artifacts,
- Checking out the a given branch of the upstream Che version and building it when the build is used by CI.

To meet these requirements, the maven build has a specific structure (using the maven `invoker` plugin) that, until now, was *not compatible with the M2Eclipse integration*. This prevented using JDT services (completion, cross-references, etc ...) on modules that are also Java projects (such as the plugins or the `assembly-wsmaster-war`)

#### Description of the Fix

This PR introduces slight changes that allow importing the RH distribution maven sub-modules *as Maven projects in Eclipse*. For projects that are Java projects, M2Eclipse can now correctly calculate the classpath based on the Maven POM file, and all the services of JDT are available again.
However the main structure of the build is not impacted, so that it doesn't impact / break the requirements described [above](description-of-the-problem).  
 
### DIFF
--- a/builds/fabric8-che/pom.xml
+++ b/builds/fabric8-che/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
-        <version>@che.version@</version>
+        <version>5.13.0-SNAPSHOT</version>
         <relativePath>@baseCheRepository@</relativePath>
     </parent>
     <groupId>com.redhat.che</groupId>
@@ -197,6 +197,14 @@
                 </executions>
             </plugin>
             <plugin>
+              <groupId>com.google.code.sortpom</groupId>
+              <artifactId>maven-sortpom-plugin</artifactId>
+              <inherited>false</inherited>
+              <configuration>
+                <sortProperties>false</sortProperties>
+              </configuration>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-dependency-plugin</artifactId>
                 <executions>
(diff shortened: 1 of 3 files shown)


davidfestal:
@mario could you review this PR please ?

l0rd:
@davidfestal I think that we should review the list of *strong requirements*. I'm not sure we really need all of them.

By the way I'm ok with this PR. Before merging I would like @ibuziuk and @sunix to have a look at that because I think they asked for this change.

davidfestal:
> I think that we should review the list of *strong requirements*. I'm not sure we really need all of them.

Sure, I should have said *strong initial requirements*. as soon as we are based on the upstream master and can have access to there  snapshot Mavenrepository, point 3 and 4 wouldn't be strong constraints anymore.
However requirements 1 or 2 are still relevant as far as I understand, and are really met by the current build structure.

cc @sunix , cc @ibuziuk

davidfestal:
So finally I was able to also test this fix against the Che maven support (in OSIO):
On a `centos-rhche-in-che` workspace (on a stack I just created) on OSIO, I was able to convert one of the RH build sub-modules (`keycloak-plugin-ide`) to a Java/Maven project and have the classpath correctly filled and the Java completion + error markers working.

amisevsk:
I've tested importing into eclipse, and it works great! Thanks @davidfestal

builds/fabric8-che/assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -22,14 +22,14 @@
...
         <dependency>
-            <groupId>aopalliance</groupId>
-            <artifactId>aopalliance</artifactId>
-        </dependency>
-        <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
         </dependency>
         <dependency>
+            <groupId>aopalliance</groupId>
```
davidfestal: This is related to a pom sorting error I encountered during the work on this PR.

builds/fabric8-che/pom.xml
```diff
@@ -172,6 +172,14 @@
...
                 </executions>
             </plugin>
+            <plugin>
+              <groupId>com.google.code.sortpom</groupId>
+              <artifactId>maven-sortpom-plugin</artifactId>
+              <inherited>false</inherited>
+              <configuration>
+                <sortProperties>false</sortProperties>
+              </configuration>
+            </plugin>
```
davidfestal: it is

builds/fabric8-che/pom.xml
```diff
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
-        <version>@che.version@</version>
```
davidfestal: Do you have any idea when you will be able to merge #89 ? it seems it depends on the merge of the `openshift-connector-rebased` branch of `eclipse/che`, which might take some more time.

What is the exact problem with this line ? Is it the fact that, on your `openshift-connector-rebased` branch, you will need to change the `5.6.0-openshift-connector-SNAPSHOT` version to make it consistent with the upstream master version (which you already did for the root `pom.xml` [here](https://github.com/redhat-developer/rh-che/pull/89/files#diff-600376dffeb79835ede4a0b285078036R25) ?

sunix: Yes

davidfestal: OK, so what is the risk or complexity ? Just one line to change in your existing PR, or am I missing something ? 

Anyway, any idea about when you should be able to merge #89 ?
It would be useful for developers to have this PR merged while working on the remaining subtasks of [CHE-229](https://issues.jboss.org/browse/CHE-229) in the meantime.

ibuziuk: < It would be useful for developers to have this PR merged while working on the remaining 
+1, without David's fix it's utterly difficult to develop plugins for rh-che

davidfestal:
@sunix let me merge this in `openshift-connector-rebased` as we discussed this morning.

PR #89, opened 2017-06-02
[WIP] Openshift connector rebased

sunix:
1. This was first a rebased of https://github.com/redhat-developer/rh-che/pull/54 with

-  structure fix after @davidfestal work
- fix stacks.json not included in assembly package

2. Fix to make rh-che work with the latest upstream che openshift-connector rebased


l0rd:
@sunix If I get it right this PR should not be merged now. We should probably wait for `openshift-connector-rebased` branch of `eclipse/che` to be merged to master. If this is the case please add [WIP] or something similar in the PR title.

builds/fabric8-che/assembly/assembly-wsagent-war/src/main/java/org/eclipse/che/wsagent/server/WsAgentAnalyticsAddresser.java
```diff
@@ -1,46 +0,0 @@
...
- * Copyright (c) 2012-2017 Codenvy, S.A.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
- *
- * Contributors:
- *   Codenvy, S.A. - initial API and implementation
- *******************************************************************************/
-package org.eclipse.che.wsagent.server;
```
davidfestal: If I remember correctly this file was added here with the idea to change, in the future, the analytics URL to something more related to the RH distribution, or even disable the mechanism totally.

I assume now we automatically get the default one from [here](https://github.com/eclipse/che/blob/master/wsagent/che-wsagent-core/src/main/java/org/eclipse/che/wsagent/server/WsAgentAnalyticsAddresser.java).

@gorkem any idea about this ?

sunix: As there is no change at the moment, we don't add it. If we need in the future something like this, let's add it at this time.
And replacing the class is not a good way to do anyway.
At the moment this is adding one more class to maintain for nothing.

davidfestal: > And replacing the class is not a good way to do anyway.
> At the moment this is adding one more class to maintain for nothing.

if you say so, and if @gorkem agrees, fine by me.

builds/fabric8-che/pom.xml
```diff
@@ -172,6 +190,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
davidfestal: Why do you fully skip the dependency analysis ? We have a build options to skip it fully from the command line.
And else there is a more fine-grained way to skip some dependency that might be a problem (for false-positive cases for example).

See here for the way to remove dependency errors individually: https://github.com/redhat-developer/rh-che/blob/07556f724354bf008e0c09587fc44b2e40e7e000/builds/fabric8-che/assembly/pom.xml#L190

sunix: To me dependency analysis is not something we need in assembly projects, and it doesn't work well with overlay. It is useful in classic jar projects.

builds/fabric8-che/assembly/assembly-wsagent-war/pom.xml
```diff
@@ -145,71 +55,4 @@
...
-                <artifactId>maven-patch-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>patch-sources</id>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-war-plugin</artifactId>
```
sunix: About the war overlays, as far as I know, when you add another war as a the dependency of a war project, it will use that one to perform the overlay. So adding overlay configuration is not needed.

sunix: about the `copy-web-resources` execution, I reincluded that. Thanks for reporting that.

davidfestal: What about the question about Keycloak ? The inherited war configuration changes the `warSourceDirectory` to `${project.build.directory}/patched/src/main/webapp`.

That's why the `copy-web-resources` execution of the `maven-resources-plugin` plugin was useful, for example to include `WEB-INF/classes/codenvy/rh-che-machine-configuration.properties` or `WEB-INF/keycloak.json`. As far as I can see in the build result they are not in the generated war.

davidfestal: > about the copy-web-resources execution, I reincluded that. Thanks for reporting that.

OK ! Sorry ! I didn't see your last message before answering :-)

builds/fabric8-che/pom.xml
davidfestal: OK. As you like.

sunix:
replaced by https://github.com/redhat-developer/rh-che/pull/117

PR #90, opened 2017-06-05
Add cleanup of previous build to fix CI 

l0rd:
CI is currently failing. This should fix the problem

### DIFF
--- a/cico_deploy.sh
+++ b/cico_deploy.sh
@@ -2,69 +2,88 @@
 set -u
 set +e
 
+echo ""
+echo "#####################################"
+echo "     STARTING BUILD VALIDATION"
+echo "#####################################"
+echo ""
+
 # Source build variables
 cat jenkins-env | grep PASS > inherit-env
 . inherit-env
 . config 
 . ~/che_image_tag.env
 
+echo "VALIDATION: Validating image ${DOCKER_HUB_NAMESPACE}/che-server:${CHE_SERVER_DOCKER_IMAGE_TAG} =="
+
 # Install oc client
 yum install -y centos-release-openshift-origin
 yum install -y origin-clients
 
+OSO_USER=mloriedo
 OSO_TOKEN=$(curl -sSL -H "X-Vault-Token: ${DOCKER_HUB_PASSWORD}" -H "Content-Type: application/json" \
 	-X GET http://li546-232.members.linode.com:8200/v1/secret/os_token | cut -d\" -f18)
 if [[ -z "$OSO_TOKEN" ]]; then
-    echo "OSO token is empty, cannot proceed with verification"
+    echo "VALIDATION: Error OSO token is empty, cannot proceed with verification"
     exit 1
 fi
-OSO_USER=mloriedo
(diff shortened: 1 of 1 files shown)


cico_deploy.sh
```diff
@@ -2,69 +2,88 @@
...
-OSIO_VERSION=$(curl -sSL http://central.maven.org/maven2/io/fabric8/online/apps/che/maven-metadata.xml | grep latest | sed -e 's,.*<latest>\([^<]*\)</latest>.*,\1,g')
 CHE_OPENSHIFT_HOSTNAME=${OSO_USER}-che.${OSO_DOMAIN}
 CHE_OPENSHIFT_PROJECT=${OSO_USER}-che
 
 oc login ${OSO_API_ENDPOINT} --token=${OSO_TOKEN}
 oc project ${CHE_OPENSHIFT_PROJECT}
-echo "Getting version of OSIO and applying template"
+
+echo "VALIDATION: Cleaning-up previous deployment"
+oc delete all -l project=che
```
mlabuda: In tests I am using 'oc apply --force=true -f -' instead of resource clean up and it works for me.

cico_deploy.sh
```diff
@@ -2,69 +2,88 @@
...
+
+echo "VALIDATION: Deploying che-server"
 curl -sSL http://central.maven.org/maven2/io/fabric8/online/apps/che/${OSIO_VERSION}/che-${OSIO_VERSION}-openshift.yml | \
     sed "s/    hostname-http:.*/    hostname-http: ${CHE_OPENSHIFT_HOSTNAME}/" | \
     sed "s/          image:.*/          image: ${DOCKER_HUB_NAMESPACE}\/che-server:${CHE_SERVER_DOCKER_IMAGE_TAG}/" | \
```
mlabuda: I think this is the real problem, because it uses RH_CHE_DOCKER_HUB_NAMESPACE var and that one points to docker.io/rhchestage. That mean we have one more slash in sed command and that is causing problem. See log of latest failing job https://ci.centos.org/view/Devtools/job/devtools-build-run-che-build-master/322/console : 

```sed: -e expression #1, char 49: unknown option to `s'```
`error: no objects passed to apply`
`No resources found.`
`Checking state of Che pod.`

cico_deploy.sh
l0rd: @mlabuda have you tested when `che-server` is idling?

cico_deploy.sh
l0rd: Good catch. Let me try to fix that.

cico_deploy.sh
mlabuda: Yeah, it works. There is one more issue - when applying a new template we check whether a pod is running immediately after applying. Sometimes could happen that old pod is still running and condition thus pass on first check but then it waits on second. As a workaround I added sleep to tests as a quickest workaround. But it would be better, to check current version of deployment and expect a version +1.

l0rd:
@mlabuda I've updated this PR please review

PR #98, opened 2017-06-08
CHE-272: Moving setting GitHub access token functionality to rh-che

ibuziuk:


ibuziuk:
@l0rd @sunix after openshift-connector will be merged to master github force activation property should be moved from super() call in constructor - https://github.com/redhat-developer/rh-che/pull/98/commits/1abbede6904cbf8ed6f8953466f4098685889faf#diff-1bf44ef91bbc868d5235bdea421f5341R40

Or we can update `openshift-connector`  branch now and remove API for setting github access token from upstream. It will require:
1. Merging this PR
2. Removing   `che.oauth.github.forceactivation` support from upstream `openshift-connector` branch
3. Providing new version of che-server to fabric8 http://central.maven.org/maven2/io/fabric8/online/apps/che/ and updating tenants
4. Updating che-starter for using new API and deploying new version on prod cluster

Not sure if it really worth doing it now though

ibuziuk:
I have slept on it and decided that it is important to remove setting github access token functionality from openshift-connector branch now to make transition to master smooth. It will be required to update che-starter with support of two versions of setting access token API (old & new one)

ibuziuk:
PR has been updated
- PR for removing  setting GitHub access token API from `openshift-connector` - https://github.com/eclipse/che/pull/5313
- PR for che-starter - https://github.com/redhat-developer/che-starter/pull/189

builds/fabric8-che/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/rh-che.properties
```diff
@@ -2,3 +2,6 @@
 # Endpoints for obtiaining Github / OpenShift Online tokens based on Keycloak token
 che.keycloak.oso.endpoint=NULL
 che.keycloak.github.endpoint=NULL
+
+# Register GitHub access token without client id and secret
+oauth.github.forceactivation=false
```
sunix: maybe should be true by default in our case (rh-che)? so we don't need to customize it in our template

builds/fabric8-che/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/oauth/OpenShiftGitHubOAuthAuthenticator.java
```diff
@@ -0,0 +1,57 @@
...
+                                             @Named("che.oauth.github.forceactivation") boolean forceActivation) throws IOException {
+
+        super(clientId, clientSecret, redirectUris, authUri, tokenUri);
+
+        if (forceActivation &&
+            !isNullOrEmpty(authUri) &&
+            !isNullOrEmpty(tokenUri) &&
+            redirectUris != null && redirectUris.length != 0) {
+
+            configure("NULL", "NULL", redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
```
sunix: maybe we could just remove this "forceActivation" as in rh-che it would be always activated

ibuziuk: yup, technically we can and probably should do this

ibuziuk:
PR has been updated

builds/fabric8-che/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/oauth/OpenShiftGitHubOAuthAuthenticator.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.api.client.auth.oauth2.TokenResponse;
+import com.google.api.client.util.store.MemoryDataStoreFactory;
+
+@Singleton
+public class OpenShiftGitHubOAuthAuthenticator extends GitHubOAuthAuthenticator {
+
+    @Inject
+    public OpenShiftGitHubOAuthAuthenticator(@Nullable @Named("che.oauth.github.clientid") String clientId,
+                                             @Nullable @Named("che.oauth.github.clientsecret") String clientSecret,
```
sunix: `clientId` and `clientSecret` are not used anymore right ? and can be removed

ibuziuk: good catch

ibuziuk:
@sunix PR has been updated

PR #110, opened 2017-06-15
Fix the `build-openshift-connector.sh` script

davidfestal:
This script doesn't work correctly for now and isn't consistent with the documentation in the Readme.
This PR fixes the problem.


davidfestal:
Some precisions about what wasn't working:
- the following command line `mvnche -pl "plugins/plugin-docker,assembly/assembly-wsmaster-war,assembly/assembly-main" install` in script `build-openshift-connector.sh` was only building the `plugins/plugin-docker` parent module, without building the sub-modules. So finally no source code was built !
- the `build-openshift-connector.sh` was not calling the fabric8 build (+ docker image creating / tag) after building the openshift connector upstream modules, though mentioned in the readme:
> - `dev-scripts/build_openshift_connector.sh`:
>    - runs the upstream Che maven build on the openshift connector related modules
>    - updates the RH distribution by running RH maven build with the options passed as arguments,
>    - generates the docker images for the upstream and the RH distribution and tag them appropriately in the *local docker environment*.
- There was not way to pass the maven `clean` phase to the openshift connector upstream build.

PR #114, opened 2017-06-21
Merge openshift-connector `stacks.json` with the one that is on master

davidfestal:
This fixes CHE-269

### DIFF
--- a/builds/fabric8-che/assembly/fabric8-stacks/src/main/resources/stacks.json
+++ b/builds/fabric8-che/assembly/fabric8-stacks/src/main/resources/stacks.json
@@ -55,13 +55,19 @@
         "commands": [{
             "commandLine": "scl enable rh-maven33 'mvn clean install -f ${current.project.path}'",
             "name": "build",
-            "type": "mvn"
-        }, {
+            "type": "mvn",
+            "attributes": {
+                "previewUrl": "",
+                "goal": "Build"
+            }
+        },
+        {
             "commandLine": "cd ${current.project.path} && scl enable rh-maven33 'java -jar target/*-fat.jar'",
             "name": "run",
             "type": "custom",
             "attributes": {
-                "previewUrl": "http://${server.port.8080}"
+                "previewUrl": "http://${server.port.8080}",
+                "goal": "Run"
             }
         }]
     },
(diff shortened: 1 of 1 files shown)


davidfestal:
I merged the `master` and `openshift-connector` branch (while making the required changes: removing the exec and ssh agents mainly), and add the content of the file to the new `stacks.json` on `rh-che`.

But I assume some still won't work correctly (such as the android stack I tested) since they eclipse docker requires a user with `sudo` rights.

@l0rd should we remove all the stacks that aren't supported in openshift.io (=> all the stacks that involve an `eclipse/xxx` docker image I assume) ?

l0rd:
@davidfestal https://issues.jboss.org/browse/CHE-269 was more about fixing stacks.json in eclipse/che (and if necessary in redhat-developer/rh-che too):

> Currently [file stacks.json on branch openshift-connector-rebased](https://github.com/eclipse/che/blob/openshift-connector/ide/che-core-ide-stacks/src/main/resources/stacks.json) is identical (or pretty close) to master. But in [openshift-connector branch stacks.json](https://github.com/eclipse/che/blob/openshift-connector/ide/che-core-ide-stacks/src/main/resources/stacks.json) there are some differences that still need to be merged to the rebased branch. For example the stack "CentOS nodejs" only exist in openshift-connector and should be ported to openshift-connector-rebased.

l0rd:
And yes `stacks.json` in rh-che should have:
- only supported stacks (java-centos, vertx, spring boot, nodejs-centos, wildfly, java-myslq-centos)
- the addition of the bayesian agent for every stack
- the removal of ssh agent for every stack

davidfestal:
let me remove the additional non-rhche stacks and this should be OK then. right @l0rd ?

davidfestal:
@l0rd 

>  But in openshift-connector branch stacks.json there are some differences (apart from `rhche` and agents) that still need to be merged to the rebased branch. For example the stack "CentOS nodejs" only exist in openshift-connector and should be ported to openshift-connector-rebased.

I reviewed the differences, and no significant difference exists now between the `openshift-connector` new RH-supported stacks in `rh-che` in the corresponding stacks in master (commit https://github.com/eclipse/che/commit/c8a8381b5d633b8e1115ee9df899d739cf16593e, just before PR https://github.com/eclipse/che/pull/5234). The reason is that some of our last changes in stacks were made as PR against master. And none is missing since @dharmit recently added the `Centos NodeJs` stack to master also.

davidfestal:
@l0rd, @gorkem  is it OK for you also ?

Issue #109, opened 2017-06-15
Workspaces are snapshotted on stop in OSIO


l0rd:
That's not the normal Eclipse Che behavior and it's probably an OpenShiftConnector bug. 
That has consequences in terms useless CPU and Storage consumption.

EDIT: The OpenShift connector doesn't really do the snapshot of the workspace so there is no resource consumptions issue. That's more about the message shown to the user being confusing. We should not say that the workspace has been snapshotted if it's not the case.

sunix:
when iddling, we should be careful to keep builds artifact such as /home/user/.m2/repository and others user folders.

l0rd:
@sunix that's a different issue. Maven local repository should be persisted in a PV. Snapshotting it is not a good idea: every snapshot would be of hundreds MB and there could be many thousands snapshots in a single day...

l0rd:
But yeah I agree that from a UX perspective persisting maven repository is even more important than avoiding snapshotting.

l0rd:
@amisevsk suggest that in OpenShift we do not really do a snapshots even if a pop up say that the image is being snapshotted. So the fix here would be not to say to users that a snapshot has been done when it's not the case. I have edited the description.

amisevsk:
@l0rd I think snapshotting workspaces is just the default behavior in Che. We should try setting [`che.workspace.auto_snapshot=false`](https://github.com/eclipse/che/blob/b95d74c5bbc8a7b8e87ce4af16392b8acca2134b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L37) and see if this resolves the problem.

For context, on OSO, "snapshots" are just ImageStreamTags that are labelled according to the snapshot that *would* have been generated. This is done because if Che thinks it has a snapshot of a workspace, it will try to launch that image (e.g. `rhche/centos_jdk8:4ce024f` instead of `rhche/centos_jdk8:latest`).

PR #116, opened 2017-06-21
Add configmap entry to disable workspace snapshots

amisevsk:
Che is by default taking "snapshots" of workspaces on shutdown. Adding the env var `CHE_WORKSPACE_AUTO__SNAPSHOT=false` disables this.

Not an urgent merge, since I don't think the pseudo-snapshots we're taking actually uses many resources.

Issue: https://github.com/redhat-developer/rh-che/issues/109
Depends on: https://github.com/fabric8io/fabric8-online/pull/245

### DIFF
--- a/dev-scripts/minishift_deploy.sh
+++ b/dev-scripts/minishift_deploy.sh
@@ -75,7 +75,8 @@ oc create configmap che \
       --from-literal=che-server-timeout-ms="0" \
       --from-literal=che-openshift-precreate-subpaths="true" \
       --from-literal=keycloak-oso-endpoint="https://sso.openshift.io/auth/realms/fabric8/broker/openshift-v3/token" \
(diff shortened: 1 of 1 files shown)


sunix:
this is only update developer script. Do we have this issue in production ?

amisevsk:
@sunix We do; I've tried to make the changes [here](https://github.com/fabric8io/fabric8-online/pull/245).

As I noted in the issue, we're not *actually* creating snapshots, but making this change will avoid creating unnecessary ImageStreamTags in OpenShift. It's not urgent to merge now, since we're focussing on bringing our branch into master and the impact of this issue is low.

amisevsk:
@sunix I've rebased onto master since https://github.com/fabric8io/fabric8-online/pull/245 is now merged.

PR #120, opened 2017-06-26
extends lsp server to support python ecosystem

invincibleJai:
* Moves Project specific logs in it's own directory (`/home/user/che`)
* Python ecosystem support for recommendations like maven and npm environment.


invincibleJai:
@sunix yes it is included in current one.

Issue #115, opened 2017-06-21
503s are back


l0rd:
This issue was originally reported by @jfchevrette [via email](https://gist.github.com/l0rd/12a6f7fc6397b4156eb24e9a6d21d886#file-jf-email-md)

The suspect is that an error 503 is returned when che-server is idled and its route receives an HTTP request. Here is a short video that shows the behavior (note that 503s are preceded by a 504 (timeout) and that 302 is normal because we are not authenticated):

![che-demo](https://user-images.githubusercontent.com/606959/27389508-e40028c4-569e-11e7-9eec-05c4285f626e.gif)

There are a couple of work in progress that may have an effect on this issue though:
1. Platform should soon consume che-starter responses asynchronously (c.f. https://github.com/redhat-developer/che-starter/issues/163)
2. A new PR for new implementation of Che workspace idling may be merged soon (https://github.com/redhat-developer/che-starter/issues/163)

l0rd:
A comment about 503 errors we had in the past: https://gist.github.com/l0rd/12a6f7fc6397b4156eb24e9a6d21d886#file-503-issues-we-had-before-summit-md

l0rd:
cc @ibuziuk

mlabuda:
After some failure in Che functional tests I dedicated to look why I have those failures. Basically Che starter does not wait until Che server pod is really running. Che server pod is in starting state but not running and Che starter proceed, that is the reason why some "Service" throws 503 in Che starter log

ibuziuk:
@mlabuda closing. realated PRs to `fabric8-online`  and `che-starter` have been merged and promoted to prod.

PR #159, opened 2017-06-29
rh-che #115: 'api/system/state' endpoint should not be secured by keycloak in order to be used in health check

ibuziuk:


builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -15,14 +15,22 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
             throws IOException, ServletException {
         HttpServletRequest request = (HttpServletRequest) req;
         String auth = request.getHeader("Authorization");
-        if(auth == null ){
+        String uri = request.getRequestURI();
+
+        if (uri.endsWith("/api/system/state")) {
+            System.out.println("Che server status endpoint should not be secured: " + uri);
```
benoitf: System.out ?

ibuziuk: @benoitf it was initially used in the class for logging, so I just decided to use it as well

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -15,14 +15,22 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
             throws IOException, ServletException {
         HttpServletRequest request = (HttpServletRequest) req;
         String auth = request.getHeader("Authorization");
-        if(auth == null ){
+        String uri = request.getRequestURI();
+
+        if (uri.endsWith("/api/system/state")) {
```
amisevsk: WDYT about adding this check below with the other skipped paths?

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
amisevsk: I agree that we should be using the standard logging framework, but this isn't necessarily the place to change it.

As it stands I feel like the filter logs too much anyways, many of these messages should be moved to the `debug` level.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
ibuziuk: @amisevsk well, I deliberately moved this check to the new block, so that the code would  be more readable, otherwise it  would be lost in numerous websocket checks (IMO it would make more sense to extract those into a separate method e.g.  `isWebSocketRequest()`)

l0rd: +1 to have two separate methods: `isWebSocketRequest()` and `isNonSecuredEndpoint()`

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
l0rd: I agree that this should be fixed but not in this PR. I've opened https://github.com/redhat-developer/rh-che/issues/164 to track the issue.

ibuziuk:
@l0rd ok, merging now and will provide a separate PR with refactoring when proper workspace security will be in place

PR #165, opened 2017-07-03
Use the correct upstream version

benoitf:
The upstream version is now 5.15.0


benoitf:
@sunix @davidfestal

benoitf:
I think yes, we should have only one place to toggle that che version
It's error prone when you've two locations to update one version

we should only have to set https://github.com/redhat-developer/rh-che/blob/0ec6cb3975a75e99948a33be41273e91a751603e/pom.xml#L25

Else it's hard to understand quickly while some of artifacts are not using the correct versions

PR #170, opened 2017-07-03
Refactoring in 'che-keycloak-plugin-server'

ibuziuk:
Need to apply this one in order to prevent console pollution once new health / readiness probes[1] will be in place 

[1] https://github.com/fabric8io/fabric8-online/pull/257

### DIFF
--- a/builds/fabric8-che/plugins/keycloak-plugin-server/pom.xml
+++ b/builds/fabric8-che/plugins/keycloak-plugin-server/pom.xml
@@ -50,6 +50,11 @@
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
     <build>
         <sourceDirectory>${src.java.dir}</sourceDirectory>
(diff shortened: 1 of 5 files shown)


ibuziuk:
cc: @benoitf

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -9,33 +20,50 @@
...
-            chain.doFilter(req, res);
-        }
+        String authHeader = request.getHeader("Authorization");
+        String requestURI = request.getRequestURI();
+        String requestScheme = req.getScheme();
 
-        if (auth == null) {
-            System.out.println("No auth header for " + request.getRequestURI());
+        if (authHeader == null) {
+            LOG.debug("No auth header for " + requestURI);
```
benoitf: we should use comma instead of + ( so we compute string only if debug is enabled)

ibuziuk:
@benoitf PR has been updated

ibuziuk:
@benoitf yeah, lets' do it in another PR once M2Eclipse [1] will be in master - for now working with plugins' code is frustrating - switching / cherry-picking between branches :/ 

[1] https://github.com/redhat-developer/rh-che/pull/87

builds/fabric8-che/plugins/keycloak-plugin-server/pom.xml
```diff
@@ -46,6 +46,10 @@
             <artifactId>keycloak-servlet-filter-adapter</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
```
benoitf: we could use scope provided as well

ibuziuk: fixed

Issue #139, opened 2017-06-28
Fix workspace keycloak authentication


ibuziuk:
see - https://issues.jboss.org/browse/CHE-291

davidfestal:
Fixed in PR https://github.com/redhat-developer/rh-che/issues/173

amisevsk:
Closing issue as fixed, since https://github.com/redhat-developer/rh-che/pull/117 was merged into master. Feel free to reopen, but I think the remaining issues are better served by separate issues.

amisevsk:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix workspace keycloak authentication](https://trello.com/c/PjqGPNSr/4-fix-workspace-keycloak-authentication)

PR #172, opened 2017-07-06
Fix workspace keycloak authentication

amisevsk:
Adds functionality to our Keycloak integration to also check which user is associated with a provided keycloak token, and blocks access to Che if that user is not the owner of the namespace in which Che is deployed

Changes:
- Add endpoint `/api/token/user`, which returns true if the token provided in the authorization header corresponds to the user who owns the namespace.
- Implement an extension of `KeycloakAuthenticatorValve` to perform this check using the above endpoint
- Update `KeycloakAuthenticationFilter` to do the same
- Testing Che locally is no longer possible with keycloak enabled, as keycloak integration now depends on external keycloak server for OSO tokens, which won't match local OSO tokens.

These changes don't include any updates to restrict access to the dashboard war (I couldn't figure out a straightforward way to implement this, but also didn't look too hard). As a result, forbidden users get stuck on a loading screen when trying to access the dashboard. The updated valve does block access to workspaces, however.

I consider this a first pass at including this functionality in keycloak, so suggestions for improvement are *very* welcome.

Fixes issue #139.

### DIFF
--- a/builds/fabric8-che/assembly/assembly-ide-war/pom.xml
+++ b/builds/fabric8-che/assembly/assembly-ide-war/pom.xml
@@ -352,7 +352,6 @@
                             <groupId>org.eclipse.che.plugin</groupId>
                         </exclusion>
                     </exclusions>
-                    
                 </dependency>
             </dependencies>
         </profile>
--- a/builds/fabric8-che/assembly/assembly-ide-war/src/main/patches/src/main/webapp/META-INF/context.xml.patch
+++ b/builds/fabric8-che/assembly/assembly-ide-war/src/main/patches/src/main/webapp/META-INF/context.xml.patch
@@ -4,5 +4,5 @@
  -->
  <Context allowCasualMultipartParsing="true">
      <Valve className="org.apache.catalina.valves.rewrite.RewriteValve"/>
-+    <Valve className="org.keycloak.adapters.tomcat.KeycloakAuthenticatorValve"/>
++    <Valve className="com.redhat.che.valve.UserAuthValve"/>
  </Context>
--- a/builds/fabric8-che/assembly/assembly-main/pom.xml
+++ b/builds/fabric8-che/assembly/assembly-main/pom.xml
@@ -32,6 +32,10 @@
             <artifactId>fabric8-ide-assembly-wsmaster-war</artifactId>
             <type>war</type>
         </dependency>
+        <dependency>
+            <groupId>com.redhat.che</groupId>
+            <artifactId>user-auth-valve</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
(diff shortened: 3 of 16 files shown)


amisevsk:
Moved PR to #173, so that it's from within the repo and targeting the openshift-connector-rebased branch.

ibuziuk:
Reopening. @davidfestal don't we need this fix in master ?

davidfestal:
well, I don't know: we just discussed yesterday at standup that we should to this on the `openshift-connector-branch`. But maybe I didn't understand correctly. @l0rd ?

davidfestal:
anyway, now we have both PRs so we can do both.

l0rd:
Yes let's do both. By the way as soon as #173 and the other PR @davidfestal is working on will be merged to openshift-connector-rebased branch we will merge the branch to master (hopefully by the end of the week).

ibuziuk:
> Testing Che locally is no longer possible with keycloak enabled, as keycloak integration now depends on external keycloak server for OSO tokens, which won't match local OSO tokens.

@davidfestal @amisevsk could you please clarify this statement? I thought keycloak integration has been always depended on external keycloak server and it was possible to test locally (does it mean that this PR can not be tested on minishift) ?

davidfestal:
> @davidfestal @amisevsk could you please clarify this statement

@ibuziuk I'll test this as soo as I've created my PR for Keycloak enablement first.

amisevsk:
@ibuziuk The issue is that the way I've had to add user checking depends on the external server to get the Openshift token for a specific keycloak token. This openshift token is used to check which username is assigned to it. A local minishift deploy will not recognize the token, making identifying the user impossible.

To clarify, what this PR does, basically:
1. A request comes, we use the keycloak servlet filter and valve to prompt the user for login
2. We get a new request with a keycloak token. We send this token to the sso.openshift.io endpoint (using the code added by @ibuziuk to get the OSO token
3. We send this OSO token to `/oapi/v1/user/~` on the openshift instance that contains che server to get username (this is done via OpenShiftClient).
4. We compare the username obtained in 3 against the project owner (currently, this is done by looking at the OpenShift project name, since our scheme uses `<username>-che`.

Locally, the token we get in 2. does not mean anything when sent in 3, so user auth fails every time. One way we could fix this is to add a property that points to OSIO's api, and then configuring step 3 to use that endpoint instead of the local one (for local development only -- it wouldn't be necessary when deployed). Then we could use `OPENSHIFT_USERNAME=<osio-username>` and it *should* work. I can maybe find time to implement this, if you think it is important for right now.

ibuziuk:
Workaround for invalid / expired oso tokens is relinking oso account via 

> https://sso.openshift.io/auth/realms/fabric8/account/identity

ibuziuk:
can not build the branch without `-Dfast` profile: 
```
[INFO] [WARNING] Used undeclared dependencies found:
[INFO] [WARNING]    com.google.guava:guava:jar:20.0:compile
[INFO] [WARNING]    javax.annotation:javax.annotation-api:jar:1.2:compile
[INFO] [INFO] Add the following to your pom to correct the missing dependencies: 
[INFO] [INFO] 
[INFO] <dependency>
[INFO]   <groupId>com.google.guava</groupId>
[INFO]   <artifactId>guava</artifactId>
[INFO]   <version>20.0</version>
[INFO] </dependency>
[INFO] <dependency>
[INFO]   <groupId>javax.annotation</groupId>
[INFO]   <artifactId>javax.annotation-api</artifactId>
[INFO]   <version>1.2</version>
[INFO] </dependency>
[INFO] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.10:analyze-only (analyze) on project che-keycloak-plugin-server: Dependency problems found -

```

l0rd:
#173 has been merged to master. I'm closing this PR.

PR #173, opened 2017-07-06
Improve Keycloak user validation to restrict access to Che

amisevsk:
**Note**: this PR is the same as #172 except made from the redhat-developer repo and rebased onto the `openshift-connector-rebased` branch. However, I haven't had a chance to test it with code changes that are present on this branch, so there may be issues I'm unaware of at the moment.

---

Adds functionality to our Keycloak integration to also check which user is associated with a provided keycloak token, and blocks access to Che if that user is not the owner of the namespace in which Che is deployed

Changes:

- Add endpoint `/api/token/user`, which returns true if the token provided in the authorization header corresponds to the user who owns the namespace.
- Implement an extension of `KeycloakAuthenticatorValve` to perform this check using the above endpoint
- Update `KeycloakAuthenticationFilter` to do the same
- Testing Che locally is no longer possible with keycloak enabled, as keycloak integration now depends on external keycloak server for OSO tokens, which won't match local OSO tokens.

These changes don't include any updates to restrict access to the dashboard war (I couldn't figure out a straightforward way to implement this, but also didn't look too hard). As a result, forbidden users get stuck on a loading screen when trying to access the dashboard. The updated valve does block access to workspaces, however.

I consider this a first pass at including this functionality in keycloak, so suggestions for improvement are very welcome.

Fixes issue #139.

PR #174, opened 2017-07-06
CHE-283 : Add a property to disable keycloak authentication in rh-che assembly

davidfestal:
This PR introduces the new `che.keycloak.disabled` property in the `rh-che.properties` file.

When this property is set to `true`, all the related Keycloak actions, both client side and server-side are bypassed, and no authentication is required.

### DIFF
--- a/builds/fabric8-che/assembly/assembly-ide-war/src/main/patches/src/main/webapp/IDE.jsp.patch
+++ b/builds/fabric8-che/assembly/assembly-ide-war/src/main/patches/src/main/webapp/IDE.jsp.patch
@@ -2,7 +2,7 @@
 +++ src/main/webapp/IDE.jsp
 @@ -55,7 +55,37 @@
          };
- 
+
      </script>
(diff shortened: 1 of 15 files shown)


ibuziuk:
@davidfestal Does this PR also includes the work that is done in - https://github.com/redhat-developer/rh-che/pull/173 (there are common parts in both but some functionality is missing) ?

davidfestal:
> @davidfestal Does this PR also includes the work that is done in - #173 (there are common parts in both but some functionality is missing) ?

No. I'm testing #173 currently (with a small modification to allow testing Keycloak on Minishift). Then when we merge #173 to `openshift-connector-rebased`, I'll rebase my PR on #173 .

But you have the opportunity to review the main changes of this PR in the meantime :-)

ibuziuk:
@davidfestal ok, thanks for clarifications. 

`(with a small modification to allow testing Keycloak on Minishift)`

Could you please explain how to test #173 on Minishift ?

davidfestal:
@gorkem @l0rd @ibuziuk could you review this PR please ?

PR #182, opened 2017-07-10
Revert "Skip functional test in CI because they are broken"

mlabuda:
Reverts redhat-developer/rh-che#171, tests now should run successfuly

### DIFF
--- a/cico_deploy.sh
+++ b/cico_deploy.sh
@@ -16,14 +16,13 @@ echo "export CHE_SERVER_DOCKER_IMAGE_TAG=$CHE_SERVER_DOCKER_IMAGE_TAG" >> $confi
 echo "export DOCKER_HUB_NAMESPACE=${DOCKER_HUB_NAMESPACE}" >> $config_file
 
 # Triggers update of tenant and execution of functional tests
-echo "CHE VALIDATION: Verification skipped until job devtools-che-functional-tests get fixed"
-# git clone https://github.com/redhat-developer/che-functional-tests.git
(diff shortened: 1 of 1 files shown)


l0rd:
Let's merge that when che-starter is updated in prodution (c.f. https://github.com/openshiftio/saas/pull/256)

ibuziuk:
@l0rd @mlabuda che-starter has been updated on prod, I think we are good to merge

l0rd:
@mlabuda another thing I would like to make sure before merging this PR is that, if the functional tests fail, we (the Che team) need to be able to access the following logs:

- openshift events
- che-server
- wsagent
- bayesian

Not having access to those logs would make almost impossible to understand what is wrong and  address the failure.

ibuziuk:
@rhopp @Katka92 any updates on this issue ? Is it planned to include verification [1] step before pushing images ?

[1] https://github.com/redhat-developer/rh-che/blob/master/.ci/cico_deploy.sh#L32

rhopp:
3 of last ~30 runs failed [1]. Every one of them with different error/failure. We need to investigate those before we can include the verification step before pushing images.

[1] - https://ci.centos.org/view/Devtools/job/devtools-che-functional-tests-periodical/

ibuziuk:
@rhopp thanks for clarifications. Please, let us know if some action is required from dev side. It would be really great to enable verification step in the short run

rhopp:
I've created issue here: https://github.com/redhat-developer/che-functional-tests/issues/139
Closing this PR as I will be providing another PR.

Issue #180, opened 2017-07-07
Can not open workspace from Codebases view (ERR_TOO_MANY_REDIRECTS)


ibuziuk:
The following error occurs when trying to open a workspace from "Codebases" view on openshift.io: 
```
This page isn’t working
sso.openshift.io redirected you too many times.
Try clearing your cookies.
ERR_TOO_MANY_REDIRECTS
```
Workaround: clearing cookies / incognito mode sometimes also helps

alexeykazakov:
I see there is "response_mode=fragment" in HTTP request from Che to sso.openshift.io (Keycloak). Most likely it's from JavaScript Keycloak Adapter. I guess this forces Keycloak to use fragments (#) instead of URL parameters (?) in redirect URLs. Fragments can be handled by JavaScrtipt only. If Tomcat (Java) Adapter is involved here too then it won't handle fragments correctly.

ibuziuk:
Might be related issue - https://github.com/fabric8-ui/fabric8-ui/issues/1539

davidfestal:
let me try to explicitly disabling the `fragment` response mode in the dashboard code.

alexeykazakov:
I actaully see problem only when response_mode=fragment is not used. In both cases Keycloak uses fragments (have no idea why) but it works only if response_mode=fragment
I guess because it's keycloak.js that sends requests with  response_mode=fragment 
but when it's not working the request comes from Tomcat Adapter or whatever is used by Che

l0rd:
The fix for this has been deployed on OSIO (tenant version 1.0.208)

PR #186, opened 2017-07-11
Fix for the "too many redirects" error when using Keycloak

davidfestal:
The fixes issue https://github.com/redhat-developer/rh-che/issues/180

With this issue it was impossible to open a workspace from Codebases view unless you clean your browser cache and cookies.


davidfestal:
After rebasing on the last `openshift-connector-rebased` branch and testing again, it seems the fix is not sufficient I got the error (or a similar error again).

davidfestal:
> I do not think response mode:query will do any good here.

@gorkem well, as we discussed yesterday, it seems it *is* necessary to have the bug fixed in all use-cases.

Issue #137, opened 2017-06-27
Setting header Authorization: Internal allows bypassing keycloak on Che.


ibuziuk:
see https://issues.jboss.org/browse/CHE-299

Making API requests with the header `Authorization: Internal` allows one to bypass keycloak validation; you can delete someone's workspaces from a plain bash terminal if you know their base url.

ibuziuk:
@gorkem @amisevsk do you happen to remember why it was initially implemented to skip requests with `Authorization: Internal` in https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java#L60 ?

amisevsk:
@ibuziuk It was necessary because internal requests don't have an authorization header and so get blocked. The header is added by [KeycloakHttpJsonRequestFactory](https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakHttpJsonRequestFactory.java) which is bound for the rh-che distribution. 

Two approaches I was considering was detecting that the source of the requests is internal, or alternatively, setting passing the service account token as a header and using that to confirm it came internally.

ibuziuk:
@amisevsk thanks for clarifications, using service account token looks like a valid solution

ibuziuk:
PR is pushed to master

PR #187, opened 2017-07-11
rh-che #137: Using service account token in 'Authorization' header for internal requests coming from wsagent to wsmaster

ibuziuk:


ibuziuk:
@l0rd @davidfestal @sunix @benoitf guys, if you have some time could you take a look at this non-working fix. The PR does is meant to do the following ( the 3rd step is failing though):

1. Obtain token once in `ServiceAccountTokenProvider` - https://github.com/redhat-developer/rh-che/pull/187/commits/279e1532185e4c26a1bdf085ef40774772961f40#diff-50201ff67e43cab317d59cdcffa4d0bfR27

2. Set this token in `Athorization` header in `KeycloakJsonRequestFactory`  - https://github.com/redhat-developer/rh-che/pull/187/commits/279e1532185e4c26a1bdf085ef40774772961f40#diff-caee93a9d7fa021f67722b3c7669ab33R36

3. Check this token in `KeycloakAuthenticationFilter` - https://github.com/redhat-developer/rh-che/pull/187/commits/279e1532185e4c26a1bdf085ef40774772961f40#diff-3491cb55057a810ce78e6db30ea72245R69

I have added some logging which claims the following:

1. Obtaining the token happens only once
3. Service account tokens from `KeycloakJsonRequestFactory` and `KeycloakAuthenticationFilter` are the same.
2. Token that is set in `KeycloakJsonRequestFactory` is correct (service account token)
4. During the comparison in `KeycloakAuthenticationFilter` the `authHeader` is different from what was set in `KeycloakJsonRequestFactory` 

if I change the implementation of `ServiceAccountTokenProvider` the following:

`this.token = "Bearer " + RandomStringUtils.randomAlphanumeric(64);
`
in `KeycloakAuthenticationFilter` when the comparison of header / token happens those are completely different. Basically, the fix works only if the token in `ServiceAccountTokenProvider` is hardcoded in init method.
Any ideas of what might be wrong ?

ibuziuk:
Reimplementing provider via static block results in the same problem: 

```
public final class ServiceAccountTokenProvider {
    private static final Logger LOG = LoggerFactory.getLogger(ServiceAccountTokenProvider.class);

    private static final String token;

    static {
        try (OpenShiftClient client = new DefaultOpenShiftClient()) {
            token = client.getConfiguration().getOauthToken();
            LOG.info("token from static block {}", token);
        }
    }

    private ServiceAccountTokenProvider() {
    }

    public static String getToken() {
        return token;
    }
}
```
Also works fine with hardcoded token

ibuziuk:
Trying to dig this issue further and start inspecting ws-agent logs. Basically, the code with obtaining service account token is called twice on wsmaster and wsagent side and the tokens are different (which explains the failure) :

wsmaster logs:

`2017-07-11 18:11:39,115[ost-startStop-1]  [INFO ] [.s.ServiceAccountTokenProvider 33]   - Token from PostConstructor block <TOKEN_1>`

wsagent logs:

`017-07-11 18:30:31,528[ost-startStop-1]  [INFO ] [.s.ServiceAccountTokenProvider 33]   - Token from PostConstructor block <TOKEN_2>`

So obtaining the token is called twice by wsmaster & wsagent which produce two different tokens for some reason.

Here is the output of the `oc` commands:

`oc describe serviceaccount che`

```
Name:		che
Namespace:	eclipse-che
Labels:		<none>

Image pull secrets:	che-dockercfg-gwvvh

Mountable secrets: 	che-token-86tiw
                   	        che-dockercfg-gwvvh

Tokens:            	che-token-86tiw
                   	        che-token-x4ufk
```

<TOKEN_1> from wsmaster = `oc describe secret che-token-86tiw`
<TOKEN_2> from wsagent does not match any service account tokens, but it matches `authHeader` from wsmaster logs when token / header comparison happens in   `KeycloakAuthenticationFilter` [1]

[1] https://github.com/redhat-developer/rh-che/pull/187/files#diff-3491cb55057a810ce78e6db30ea72245R69

ibuziuk:
closing in favour of https://github.com/redhat-developer/rh-che/pull/196

PR #196, opened 2017-07-13
rh-che #137: Using service account token in 'Authorization' header for internal requests coming from wsagent to wsmaster

ibuziuk:


ibuziuk:
More details can be found in previous PR - https://github.com/redhat-developer/rh-che/pull/187

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -83,13 +87,55 @@ private boolean isSystemStateRequest(String requestURI) {
...
+     * @param token
+     * @return true if openshift token is valid, false otherwise
+     */
+    private boolean isValidToken(final String token) {
+        LOG.debug("Validating workspace agent token");
+        Config config = new ConfigBuilder().withOauthToken(token).build();
+        try (OpenShiftClient client = new DefaultOpenShiftClient(config)) {
+            String namespace = client.getConfiguration().getNamespace();
+            LOG.debug("Validating the token against namespace '{}'", namespace);
+            return namespace.contains("-che");
```
amisevsk: Will this have the issue that I can use my token, prepended with Wsagent, to access someone else's workspace? 

e.g. If I get my OSO token, afaict the code above will return true if I pass the auth header `Wsagent <my token>`, since my namespace will be something like `<username>-che`?

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -83,13 +87,55 @@ private boolean isSystemStateRequest(String requestURI) {
...
+        } catch (Exception e) {
+            LOG.debug("The token is not valid", token);
```
amisevsk: Missing `{}` in the error message.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
ibuziuk: ah, damn good point - if service account is going to be used for wsagent communication some jwt parsing should be involved for verification.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
ibuziuk: fixed

ibuziuk:
```
Will this have the issue that I can use my token, prepended with Wsagent, to access someone else's workspace?

e.g. If I get my OSO token, afaict the code above will return true if I pass the auth header Wsagent <my token>, since my namespace will be something like <username>-che?
```

@amisevsk actually no, it will not be possible to hack wsagent API this way ;-) Basically, jwt tokens are used for authorization and the parsed version of it looks the following way ( the easiest way to parse the token is via https://jwt.io/ ):

```
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "eclipse-che",
  "kubernetes.io/serviceaccount/secret.name": "default-token-v32hgs",
  "kubernetes.io/serviceaccount/service-account.name": "default",
  "kubernetes.io/serviceaccount/service-account.uid": "1033c10e-6aec-11e2-9347-3aa8a044c2f1",
  "sub": "system:serviceaccount:eclipse-che:default"
}
```

So, if you try to hack my che using your token, it will fail in `isValid()` method [1], because you will not be able to get the namespace using your `secret.name` which does not exist in my service account.  

[1] https://github.com/redhat-developer/rh-che/pull/196/files#diff-3491cb55057a810ce78e6db30ea72245R132

ibuziuk:
@amisevsk PR has been updated. I slept on it and figure out that my last comment was utterly wrong - now there is also namespace validation in place.

*NOTE:* this PR can be built only using `-Dskip-enforce=true` flag, I am going to open a CQ  for `io.jsonwebtoken` lib after review

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -83,13 +92,55 @@ private boolean isSystemStateRequest(String requestURI) {
...
+     * @param token
+     * @return true if openshift token is valid and matches service account namespace, false otherwise
+     */
+    private boolean isValidToken(final String token) {
+        LOG.debug("Validating workspace agent token");
+        Config config = new ConfigBuilder().withOauthToken(token).build();
+        try (OpenShiftClient client = new DefaultOpenShiftClient(config)) {
+            String namespace = client.getConfiguration().getNamespace();
+            LOG.debug("Validating the token against namespace '{}'", namespace);
+            return serviceAccount.getNamespace().equals(namespace);
```
amisevsk: I'm not sure I understand the need for ServiceAccount and ServiceAccountTokenParser. Is it possible to use the block 

```java
    try (OpenShiftClient client = new DefaultOpenShiftClient()) {
        this.namespace = client.getNamespace();
    }
    //...
    return this.namespace.equals(tokenNamespace);
```
as a (maybe static block) in `KeycloakAuthenticationFilter`? I understand that the SA token returned is going to be different between wsagent and wsmaster, since they are running in separate pods (and so the token passed to `isValidToken()` won't match), but the namespace should be the same. The above block would get the namespace of the current pod where we're running, and line 137 above would get the namespace of the token passed in. This *should* still avoid the issue of me accessing your workspace, since my token would lead to a different namespace.

I haven't tested it though, so I may have missed something. I've been burned by OS before.

ibuziuk:
@amisevsk another great catch. I have updated PR and now there is no need for jwt token parsing.

ibuziuk:
@l0rd are you ok to merge this one ?

PR #202, opened 2017-07-19
Add swagger api doc to TokenController

### LINKED ISSUE #178: Add Javadoc in class `TokenController`
The class `/che-keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/contoller/TokenController.java` misses javadoc comments on the endpoint methods.

Just a reminder of PR comment https://github.com/redhat-developer/rh-che/pull/173#discussion_r126131357 that will finally be fixed in a new PR.

amisevsk:
Fixes #178


sunix:
could the description be something like that ?

> which returns true if the token provided in the authorization header corresponds to the user who owns the namespace.

?

builds/fabric8-che/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/contoller/TokenController.java
```diff
@@ -117,6 +121,7 @@ public Response getOpenShiftToken(@HeaderParam(HttpHeaders.AUTHORIZATION) String
 
     @GET
     @Path("/user")
+    @ApiOperation(value = "Check if passed keycloak token matches token assigned to OSIO user who owns current namespace.")
```
sunix: could the description be something like that ?
https://github.com/redhat-developer/rh-che/pull/173#issue-240872944

> which returns true if the token provided in the authorization header corresponds to the user who owns the namespace.

?

amisevsk: That does sound better -- thanks.

Issue #94, opened 2017-06-07
Che: context assist HTTP 401


joshuawilson:
_From @burrsutter on June 6, 2017 23:49_

https://screencast.com/t/549bv3F2

Using the Vert.x Basic template and 
router.
looking at the methods of the Vert.x router - they all pop-up HTTP 401

_Copied from original issue: openshiftio/openshift.io#201_

joshuawilson:
This issue was moved to fabric8io/fabric8-ui#1461

l0rd:
The (temporary) solution is to remove authentication for some http requests. But we need to open a issue on upstream Che to solve the problem and in any case that issue does not exist when we you a language server.

davidfestal:
The underlying root cause is the use of GWT `Frame` objects to manage the popup of Java Quick documentation and Java completion documentation.
The Frame object creates an new html frame in which the client-side Keycloak-related Javascript code cannot be injected. And thus the http requests sent from these frames don't get the right Keycloak authorization token, and are refused on the server-side by the Keycloak Authentication filter.

FTR I created the following issue in upstream Che: https://github.com/eclipse/che/issues/5745

davidfestal:
This is fixed in production since version `1.0.234` of OSIO Che

PR #214, opened 2017-07-19
Fix for the context assist HTTP 401/403 errors in Che

davidfestal:
The fixed issue is https://github.com/redhat-developer/rh-che/issues/94

However this should be considered a temporary fix. The real underlying problem lies in upstream Che as mentioned in the issue. 

### DIFF
--- a/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
+++ b/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
@@ -12,6 +12,8 @@
 package com.redhat.che.keycloak.server;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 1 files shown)


builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -70,7 +72,8 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
         }
 
         if (isSystemStateRequest(requestURI) || isWebsocketRequest(requestURI, requestScheme)
-                || isKeycloakSettingsRequest(requestURI) || isWorkspaceAgentRequest(authHeader)) {
+                || isKeycloakSettingsRequest(requestURI) || isWorkspaceAgentRequest(authHeader)
+                || isRequestFromGwtFrame(request.getRequestURI())) {
```
ibuziuk: there is already "requestURI", no need to obtain it again

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -143,4 +146,15 @@ private boolean isValidToken(final String token) {
         }
     }
 
+    
+    private boolean isRequestFromGwtFrame(String requestURI) {
+        URI uri;
+        try {
+            uri = new URI(requestURI);
+            return "/api/java/javadoc/find".equals(uri.getPath()) // Java Quick Documentation popup
```
ibuziuk: Do we really need to create URI object?
Previously we used "endwith()" approach

davidfestal: > Do we really need to create URI object?
> Previously we used "endwith()" approach

There are parameters in the expected URIs, and using `String.contains()` didn't seem precise enough to me.

Issue #200, opened 2017-07-18
Rebase eclipse/che `openshift-connector-wip` branch


sunix:
using https://github.com/sunix/che/commits/openshift-connector-wip ( `openshift-connector-wip rebased to master)  and https://github.com/sunix/rh-che/tree/fix-patch-dashboard (fix patch issue after rebase)
I couldn't start a workspace. It seems the workspace starts but once the IDE is trying to connect to it, it fails and ask to stop the workspace.
Without keycloak:
![eclipse che wksp-huh2 - mozilla firefox_042](https://user-images.githubusercontent.com/650571/28359811-c2288fb2-6c73-11e7-972c-f385a9cd13a0.png)

With keycloak (after few minutes) it seems nothing happen, no error but finally got this one:
![eclipse che wksp-ac0i - google chrome_043](https://user-images.githubusercontent.com/650571/28359928-2026b288-6c74-11e7-9b0f-62bf73d74346.png)

PR #219, opened 2017-07-21
Update the RH distribution after rebasing the `openshift-connector-wip` branches to master

davidfestal:


davidfestal:
This will fix issue #200

Issue #218, opened 2017-07-20
Recommendations on pom.xml are currently not working


mlabuda:
I have a Che server running on image rhche/che-server:ba40d4c-fabric8-5bd0962 (15 hours old) and template 1.0.231. There is a env var RECOMMENDER_API_TOKEN, also there is com.redhat.bayesian.lsp agent present in workspace but there is no server.js running in workspace nor working Bayesian. Che server in log contains info about launching bayesian LSP
`2017-07-20 09:10:38,699[aceSharedPool-4] [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 692] - Launching 'com.redhat.bayesian.lsp' agent at workspace workspacex66f4i8nrf48xrqq`

l0rd:
@mlabuda I've changed the title of this issue because it looks like the issue is only for recommendation with `pom.xml` files. You can find more details below: 

**WHEN IS BAYESIAN NODE APP STARTED?** 
Bayesian node app is started only when a file (`package.json`, `pom.xml` or `requirement.txt`) is opened. That explains why you don't see the application after workspace startup. You need to open one of these files first.

**ISSUE WITH pom.xml files**
Currently recommendation agent is not activated for `pom.xml` files. That's is a side effect of the rebase we did last week on top of last version of `eclipse/che` that didn't support multiple LS on a single file. That problem has been solved and merged on `eclipse/che` and will be fixed on `rhche` with #200  

Said that Bayesian LS is working fine on OSIO for a `package.json` file:

![image](https://user-images.githubusercontent.com/606959/28425329-3df54e62-6d70-11e7-9c86-6a208a667a89.png)

cc @invincibleJai @samuzzal-choudhury

invincibleJai:
thanks @l0rd @mlabuda for putting more light on it. I tested requirements.txt as well , it's working as expected.

krishnapaparaju:
@l0rd  @invincibleJai  thanks

davidfestal:
Created PR https://github.com/redhat-developer/rh-che/pull/223. let's wait for the merge and promotion to prod before closing the issue.

invincibleJai:
@davidfestal Thanks, now bayesian LSP is working with CHE for pom.xml tested in osio
(with Image: rhche/che-server:8c76cd6-fabric8-4d2c60e) 

[...]

PR #223, opened 2017-07-24
Now enable bayesian LS on pom.xml (after the rebase on master)

davidfestal:
This is related to redhat-developer/rh-che#218

With this change, I tested (on OSIO) that the Bayesian plugin now displays recommendations on `pom.xml` files as it does on `package.json` files. 


builds/fabric8-che/plugins/che-plugin-bayesian-lang-server/src/main/java/org/eclipse/che/plugin/languageserver/bayesian/server/launcher/BayesianLanguageServerLauncher.java
```diff
@@ -87,8 +87,10 @@ private static LanguageServerDescription createServerDescription() {
                                                                               null),//
                                                            new DocumentFilter(JsonModule.LANGUAGE_ID,//
                                                                               "package\\.json",//
-                                                                              null)//
-                                             // TODO pom.xml
+                                                                              null),
+                                                           new DocumentFilter("pom",
```
sunix: Isn't it a constant you could reuse ?

davidfestal: If such a constant had been available in the maven language server I wold have taken it, but unfortunately the related code in the maven language server launcher is the following:

```java
    @Override
    public LanguageServerDescription getDescription() {
        return new LanguageServerDescription("org.eclipse.che.plugin.maven", Collections.singletonList("pom"), Collections.emptyList());
    }
```

On the other hand `pom` is quite and expected string for a pom file ;-)

tsmaeder: It probably makes sense to extract the launguage id "pom" to a constant when we start using it outside of the maven module.

davidfestal: Well, in this case we'll have to wait for the upstream PR to be merged first.
I'd prefer doing this in a second PR synchronized with a related upstream Che PR.
Is it OK for you ?

sunix: yes do 2 separates PR (che and rh-che) for that thanks

PR #227, opened 2017-07-26
Fix spring-boot run command

l0rd:
### What does this PR do?
Improve the spring-boot stack run command to hot re-deploy the app

https://github.com/openshiftio/openshift.io/issues/415


ibuziuk:
@davidfestal does redeploy works for you ? was not able to make it work on osio :/

l0rd:
@ibuziuk not sure what doesn't work here but maybe this comment may help: https://github.com/openshiftio/openshift.io/issues/416#issuecomment-318117751. https://github.com/openshiftio/openshift.io/issues/416 was closed but the problem was not fixed "entirely" and we still need to fix it. You may want to open a new issue about that in rh-che repo.

ibuziuk:
@l0rd thanks for the pointers - I have already opened - https://github.com/redhat-developer/rh-che/issues/254

Issue #133, opened 2017-06-27
Init tenant should set the correct Keycloak URL


ibuziuk:
see - https://issues.jboss.org/browse/CHE-282

With commmit in fabric8-online[1] we have introduced two new configs for che-server:

- keycloak-oso-endpoint
- keycloak-github-endpoint

These are currently hardcoded in fabric8io configuration with KC production URLs. That's an issue because we may need to use a distinct instance of KC based on which cluster Che will be deployed to. These 2 config maps should be configured by the tenant init script as it's already done for hostname-http

[1] https://github.com/redhat-developer/rh-che/commit/148fc268b6a7b26574220ad1de715ff6140fb5ce

ibuziuk:
Might be fixed via `api/keycloak/settings/`

davidfestal:
This endpoint will also be useful to provide some of the Keycloak configuration settings that, for now, are hard-wired in the client-side or server-side code of the Keycloak integration:
- WsMaster web application:
    - server-side configuration for the Keycloak servlet filter: [keycloak.json](https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/keycloak.json)

- WsAgent web application:
    - server-side configuration for the Keycloak servlet filter: [keycloak.json](https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/keycloak.json)

- GWT IDE web application:
    - server-side configuration for the Tomcat valve in front of the GWT application: [keycloak.json](
https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/assembly/assembly-ide-war/src/main/webapp/WEB-INF/keycloak.json)
    - client-side configuration in Javascript code: [IDE.jsp.patch](https://github.com/redhat-developer/rh-che/blob/d89bed0e4f1a4cfc30e11b09235731120ccc75c1/builds/fabric8-che/assembly/assembly-ide-war/src/main/patches/src/main/webapp/IDE.jsp.patch#L15)

The settings that should be configurable are mainly:
- the Keycloak server URL:
    - `url` in client-side Javascript configuration
    - `auth-server-url` in server-side Java configuration

- the Keycloak realm:
    - `realm` in client-side Javascript configuration
    - `realm` in server-side Java configuration

- the Keycloak client ID:
[...]

PR #235, opened 2017-07-31
Configure Keycloak  in the wsmaster properties, and make settings available through REST

davidfestal:
This pull request allows configuring the main Keycloak settings (realm, auth URL, etc... as well as endpoints to retrieve associated tokens such as git) in the wsmaster properties. The settings are centralized in the wsmaster, and pulled through the `wsmaster/api/keycloak/settings` endpoint from the other places that need them: user-auth tomcat valve, wsagent, and client-side JavaScript code.

There is no need for `Keycloak.json` files anymore, since Keycloak configurations are now managed centrally by `KeycloakConfigResolver` classes for both the *UI* web applications (GWT application) and the *services* web applications (wsmaster and wsagent). Using `KeycloakConfigResolver`s allows centralizing the Keycloak settings in a single place in both the source code and the deployment, but was required anyway for the tomcat valve, due to specifics of the valve lifecycle.

This is the implementation of https://github.com/redhat-developer/rh-che/issues/133#issuecomment-316663409


builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -65,13 +66,14 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
         HttpServletRequest request = (HttpServletRequest) req;
         String authHeader = request.getHeader("Authorization");
         String requestURI = request.getRequestURI();
+        String upgrade = request.getHeader("upgrade");
```
ibuziuk: will it be possible to hack / bypass keycloak auth by setting `Upgrade: WebSocket` header to the malicious request ?

ibuziuk:
@davidfestal tested on minishift - works just fine. So, I'm +1 but need some clarifications about potential security leak via `Upgrade: WebSocket` header

amisevsk:
@davidfestal I tried testing on minishift with keycloak disabled but for some reason the dashboard refuses to load (building from master works no problem). I'll try again later.

ibuziuk:
@amisevsk good catch, I can confirm that with keycloak disabled dashboard is not loading but `api/keycloak/settings` endpoint works correctly. Dashboard loads just fine with keycloak enabled though..

davidfestal:
@amisevsk @ibuziuk I fixed the problem with the dashboard when keycloak is disabled in commit https://github.com/redhat-developer/rh-che/pull/235/commits/65933c6ab371ae40219c5d431fab9f839181563b

davidfestal:
> I'm +1 but need some clarifications about potential security leak via Upgrade: WebSocket header

You're right. I just tested it and it is a potential security leak. I'm rolling back this specific change, and testing again. it seems that after rebasing on the very last `openshift-connector-wip` my change might not necessary anymore.

davidfestal:
@ibuziuk I fixed the potential security leak.

ibuziuk:
@davidfestal thanks! lemme test it again and I will approve

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/ServicesKeycloakConfigResolver.java
```diff
@@ -0,0 +1,29 @@
...
+        config.setCors(true);
+        config.setBearerOnly(true);
+        config.setPublicClient(false);
+        config.setConnectionPoolSize(20);
+        config.setDisableTrustManager(true);
+        config.setEnableBasicAuth(false);
+        config.setEnableBasicAuth(false);
+        config.setExposeToken(true);
+        Map<String, Object> credentials = new HashMap<>();
+        credentials.put("secret", "08a8bcd1-f362-446a-9d2b-d34b8d464185");
```
gorkem: I think we need to verify the secret settings with @alexeykazakov

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static void pullFromApiEndpointIfNecessary(String apiEndpoint) {
+        if (settings == null) {
+            URL url;
+            HttpURLConnection conn;
+            try {
+                url = new URL(KeycloakConstants.getEndpoint(apiEndpoint));
+                if (LOG.isLoggable(Level.INFO)) {
```
gorkem: Are these isLoggable() checks necessary?

builds/fabric8-che/assembly/assembly-wsagent-war/src/main/java/org/eclipse/che/wsagent/server/RedHatCheWsAgentServletModule.java
```diff
@@ -28,8 +36,13 @@ protected void configureServlets() {
...
-        bind(Boolean.class).annotatedWith(Names.named("che.keycloak.disabled")).toProvider(KeycloakPropertiesProvider.class);
+        bind(Boolean.class).annotatedWith(Names.named(KeycloakConstants.DISABLED_SETTING)).toProvider(KeycloakDisabledPropertyProvider.class);
+        bind(String.class).annotatedWith(Names.named(KeycloakConstants.AUTH_SERVER_URL_SETTING)).toProvider(KeycloakAuthServerUrlPropertyProvider.class);
+        bind(String.class).annotatedWith(Names.named(KeycloakConstants.CLIENT_ID_SETTING)).toProvider(KeycloakClientIdPropertyProvider.class);
+        bind(String.class).annotatedWith(Names.named(KeycloakConstants.REALM_SETTING)).toProvider(KeycloakRealmPropertyProvider.class);
         bind(KeycloakAuthenticationFilter.class).in(Singleton.class);
-        filter("/*").through(KeycloakAuthenticationFilter.class);
+        final Map<String, String> keycloakFilterParams = new HashMap<>();
+        keycloakFilterParams.put("keycloak.config.resolver", ServicesKeycloakConfigResolver.class.getName());
+        filter("/*").through(KeycloakAuthenticationFilter.class, keycloakFilterParams);
```
davidfestal: Well, I cannot inject the `che.keycloak.disabled` property value and use it here, since it is provided by the `KeycloakDisabledPropertyProvider` just bound above. And this provider (indirectly) pulls the setting from the wsagent.

That's why it seems simpler to take care of the `disabled` setting in the `KeycloakAuthenticationFilter` itself IMO.

gorkem: makes sense

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/ServicesKeycloakConfigResolver.java
davidfestal: @gorkem  I've asked him the question. However, since I simply added the secret value we were already using before the changes,  do you think we could merge this PR and manage the secret change later ?

builds/fabric8-che/assembly/dashboard/src/main/patches/src/app/index.module.ts.patch
```diff
@@ -105,7 +110,7 @@
  
  // add interceptors
  initModule.factory('ETagInterceptor', ($window, $cookies, $q) => {
-@@ -361,6 +445,7 @@
+@@ -361,6 +450,7 @@
```
davidfestal: @gorkem Do you think we could leave this for a future PR ? in order to avoid adding changes that are not related to the PR main purpose ?

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/ServicesKeycloakConfigResolver.java
alexeykazakov: The public client (used by Che) doesn't have any secret.

builds/fabric8-che/assembly/assembly-wsagent-war/src/main/java/org/eclipse/che/wsagent/server/RedHatCheWsAgentServletModule.java
benoitf: FYI if we only need a singleton map, instead of using new HashMap and map.put, you can use
https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#singletonMap(K,%20V)

builds/fabric8-che/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/rh-che.properties
```diff
@@ -3,3 +3,6 @@
 che.keycloak.oso.endpoint=NULL
 che.keycloak.github.endpoint=NULL
 che.keycloak.disabled=true
+che.keycloak.auth-server-url=https://sso.openshift.io/auth
```
benoitf: FYI you should use _ as string separator instead of -
like on https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L104

This is because we allow overriding properties with env variables
and we can use same pattern than on Che

builds/fabric8-che/assembly/dashboard/src/main/patches/src/app/index.module.ts.patch
```diff
@@ -38,12 +43,12 @@
...
++  promise.then((keycloakSettings) => {
++    console.log('keycloakSettings : ' + keycloakSettings.toString() );
++    if (keycloakSettings['che.keycloak.disabled'] == 'true') {
++      keycloakInit(true, keycloakSettings);
 +    } else {
-+      keycloakInit(false);
++      keycloakInit(false, keycloakSettings);
 +    }
 +  }, (error) => {
 +    console.log(error);
```
benoitf: maybe add for what the error is 
console.log('Error when initializing keycloak', error);

builds/fabric8-che/assembly/dashboard/src/main/patches/src/app/index.module.ts.patch
```diff
@@ -38,12 +43,12 @@
 +    
 +  }
 +
-+  promise.then((resp) => {
-+    console.log('resp : ' + resp.toString() );
-+    if (resp['che.keycloak.disabled'] == 'true') {
-+      keycloakInit(true);
++  promise.then((keycloakSettings) => {
++    console.log('keycloakSettings : ' + keycloakSettings.toString() );
```
benoitf: I think that using console.log('keycloakSettings : ', keycloakSettings) should work as well but I'm not 100% sure

builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
```diff
@@ -113,36 +122,16 @@ private String getToken(Request request) {
...
-            }
+        if (keycloakDisabled != null) {
+            return keycloakDisabled;
+        }
+        
+        Map<String, String> keycloakSettings = KeycloakSettings.get();
+        if (keycloakSettings == null) {
+            LOG.warn("KeycloakSettings = null => don't disable Keycloak");
+        } else {
+            keycloakDisabled = "true".equals(keycloakSettings.get(DISABLED_SETTING));
```
benoitf: could we use https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#parseBoolean(java.lang.String)

builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
```diff
@@ -41,15 +46,13 @@
...
-    private static final String USER_VALIDATOR_ENDPOINT = "http://che-host:8080/api/token/user";
+    private static final String API_ENDPOINT = "http://che-host:8080/api";
+    private static final String USER_VALIDATOR_ENDPOINT = API_ENDPOINT + "/token/user";
     private static final String AUTHORIZATION_HEADER = "Authorization";
-    private static final String KEYCLOAK_SETTINGS_ENDPOINT = "http://che-host:8080/api/keycloak/settings";
-    private static final Pattern DISABLED_SETTING_PATTERN = Pattern.compile(".*\"che\\.keycloak\\.disabled\":\"([^ \"]+)\".*");
     
-    private boolean keycloakDisabledRetrieved = false;
-    private boolean keycloakDisabled = false;
+    private Boolean keycloakDisabled = null;
```
benoitf: I think it's not thread safe (AtomicBoolean ?)

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakSettingsRetriever.java
```diff
@@ -0,0 +1,31 @@
...
+@Singleton
+public class KeycloakSettingsRetriever {
+    private static final Logger LOG = LoggerFactory.getLogger(KeycloakSettingsRetriever.class);
+    
+    private final String apiEndpoint;
+    
+    @Inject
+    public KeycloakSettingsRetriever(@Named("che.api") String apiEndpoint) {
+        this.apiEndpoint = apiEndpoint;
+        LOG.info("Endpoint = " + apiEndpoint);
```
benoitf: LOG.info("Endpoint = {0}", apiEndpoint)

benoitf: BTW is it info or more debug ?

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakUserChecker.java
```diff
@@ -39,7 +40,7 @@
     private static final String ENDPOINT = "http://che-host:8080/api/token/user";
 
     @Inject
-    private KeycloakHttpJsonRequestFactory requestFactory;
+    private DefaultHttpJsonRequestFactory requestFactory;
```
benoitf: IMHO you should use interface there : HttpJsonRequestFactory

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
```diff
@@ -0,0 +1,48 @@
+package com.redhat.che.keycloak.shared;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Map;
+
+import java.util.logging.Logger;
```
benoitf: why do we use java util logging instead of SLF4J ?

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static void pullFromApiEndpointIfNecessary(String apiEndpoint) {
+        if (settings == null) {
+            URL url;
+            HttpURLConnection conn;
+            try {
+                url = new URL(KeycloakConstants.getEndpoint(apiEndpoint));
+                LOG.info("Pulling Keycloak settings from URL :" + url);
```
benoitf: is it info trace ?

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/ServicesKeycloakConfigResolver.java
```diff
@@ -0,0 +1,29 @@
+package com.redhat.che.keycloak.shared;
```
benoitf: AFAIK it is missing copyright headers

builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
davidfestal: Well, keycloak settings are expected to be the same for all requests, so in the worst concurrency case, the settings retrieval would be done several times, but with the same value assigned to the `keycloakDisabled` variable. I don't see a real problem here TBH.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakSettingsRetriever.java
davidfestal: `info` is OK since it is normally done only once at start.

builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
benoitf: yes I agree with you that the same value will be assigned but it looks odd to grab the settings several time. (so maybe it's not around Boolean/AtomicBoolean but how we may initialize the value several times.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakUserChecker.java
davidfestal: We explicitely use `DefaultHttpJsonRequestFactory` because we don't want to use the `KeycloakHttpJsonRequestFactory` that is bound into Guice by the fabric8 assembly. We don't want to secure the related request through Keycloak since this request is to retrieve Keycloak settings.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
davidfestal: Because this component is also used in the Tomcat valve that doesn't use SLF4J.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
davidfestal: `info` is OK since it is normally done only once at start.

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakSettingsRetriever.java
benoitf: but is it a useful info for "normal" operation ?

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakUserChecker.java
benoitf: so I think it should be clearly stated in the comment ? Because I guess people will try to remove implem and use interface.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
benoitf: IMHO I think my point is only that this info is useless (only useful when you want to debug it). Not if the traces is printed a lot of time or not.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
benoitf: in Tomcat that runs che ?
because AFAIK tomcat/che integrates SLF4J
https://github.com/eclipse/che-lib/tree/master/che-tomcat8-slf4j-logback

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
benoitf: but note that it's only my feeling, not requiring a change.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
davidfestal: yes, but the valve doesn't live in a web application context, and at start there were problems at runtime when using SLF4J (don't remember exactly the details)

davidfestal:
> someone from @skabashnyuk team should have a look at the PR.

I tried to add @skabashnyuk or @mshaposhnik as reviewers but I cannot.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
benoitf: that's strange because logback libraries are added on the whole server
https://github.com/eclipse/che-lib/blob/master/che-tomcat8-slf4j-logback/src/assembly/assembly.xml#L30-L31

builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
davidfestal: I fixed it by extracting the related code in a synchronized function.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/ServicesKeycloakConfigResolver.java
davidfestal: I added them

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
davidfestal: I'd prefer letting it as is if you don't mind.

builds/fabric8-che/plugins/keycloak-plugin-shared/src/main/java/com/redhat/che/keycloak/shared/KeycloakSettings.java
davidfestal: well, strange yes. Maybe it would work now, but when I first worked on this, it was producing an error. Do you really want me to change / test this now, or couldn't we keep this for another issue ?

skabashnyuk: I wouldn't recommend using SLF4J in valve if possible. Because it can produce some strange issues with class loaders. now we package logback in each war too.

benoitf: no I don't really want you to change/test it now. I would say it would be use that time (or later) to use a filter (instead of valve) instead of changing java.util.logger to SLF4J

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakUserChecker.java
```diff
@@ -39,7 +40,7 @@
     private static final String ENDPOINT = "http://che-host:8080/api/token/user";
 
     @Inject
-    private KeycloakHttpJsonRequestFactory requestFactory;
+    private HttpJsonRequestFactory requestFactory;
```
sunix: is it in purpose ?

PR #236, opened 2017-07-31
Simplify deployment on minishift and osio

l0rd:
With this PR we introduce a new script to deploy Che on Openshift (both minishift and openshift.io are supported). There is no need to use `gofabric8 volumes` and `git clone` fabric8-online anymore.

To deploy on minishift:
`./openshift_deploy.sh`

To deploy to osio:
`export OPENSHIFT_FLAVOR=osio && ./openshift_deploy.sh`


README.md
```diff
@@ -145,73 +145,64 @@ in the Minishift docker environment, as detailed in next section.
...
     
 ### Deployment prerequisites
 
-* Get [minishift](https://github.com/minishift/minishift#installation) (we have tested with v1.0.0beta4)
-  * Minishift v1.0.0beta4 includes the `oc` binary, and this should be added to your `$PATH`. The binary is located
-    in the `.minishift/cache/oc/<oc version>/` directory (e.g. on Linux, it is in `~/.minishift/cache/oc/v1.4.1/`).
-* Get [gofabric8](https://github.com/fabric8io/gofabric8#getting-started) (we have tested with v0.4.121)
-* Clone fabric8-online git repository:
+* [minishift](https://github.com/minishift/minishift#installation) (v1.0.0 or greater)
+* `oc` the OpenShift command line client tool. The binary is included in minishift distribution and can be found in folder `.minishift/cache/oc/<oc version>/`. You need to add that path to your `$PATH` environment variable.
```
amisevsk: Might be worth noting that you can add `oc` to your path through minishift via `eval $(minishift oc-env)`.

README.md
```diff
@@ -145,73 +145,64 @@ in the Minishift docker environment, as detailed in next section.
...
+### Deployment Options
 
-```bash
-dev-scripts/minishift_deploy.sh
-```
-By default, the docker image of the deployed to Minishift is the last built `nightly-fabric8` image.
+You can set different deployment options using environment variables:
 
-However, you can override this by providing the `CHE_IMAGE_TAG` variable:
+* `OPENSHIFT_FLAVOR`: possilbe values are `minishift` and `osio` (default is `minishift`)
```
amisevsk: Typo: "**possible** values are"

README.md
```diff
@@ -145,73 +145,64 @@ in the Minishift docker environment, as detailed in next section.
...
-By default, Che is deployed on minishift with the Keycloak integration *disabled*.
-
-However, to explicitly enable the Keycloak integration, you can override the `CHE_KEYCLOAK_DISABLED` variable:
+#### openshift.io only options
 
-```bash
-bash -c "export CHE_KEYCLOAK_DISABLED=false ; dev-scripts/minishift_deploy.sh"
-```
+* `OPENSHIFT_ID`: Red Hat Developers username with some exceptions:
+  * If the username is an e-mail address the OPENSHIFT_ID will the part before the @ symbol (e.g. for cbrown@redhat.com `OPENSHIFT_ID` will be cbrown)
```
amisevsk: This might cause some confusion -- I believe that in some cases RHD will append a number to your username if there is a conflict, but I'm not sure. I think KB's username is \<username\>-1

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,233 @@
...
+#   $ SCRIPT_URL=https://raw.githubusercontent.com/redhat-developer/rh-che/master/dev-scripts/openshift_deploy.sh
+#   $ export OPENSHIFT_FLAVOR=osio
+#   $ curl -fsSL ${SCRIPT_URL} -o get-che.sh
+#   $ sh get-che.sh
+#
+
+set -e
+
+# --------------
+# Pring Che logo
```
amisevsk: Typo: "**Print** Che logo"

Fancy addition 👍

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,233 @@
...
+else
+  oc login "${OPENSHIFT_ENDPOINT}" --token="${OPENSHIFT_TOKEN}"  > /dev/null
+fi
+echo "done!"
+
+# --------------------------
+# Create project (if needed)
+# --------------------------
+DEFAULT_CHE_OPENSHIFT_PROJECT="eclipse-che"
+CHE_OPENSHIFT_PROJECT=${CHE_OPENSHIFT_PROJECT:-${DEFAULT_CHE_OPENSHIFT_PROJECT}}
```
amisevsk: This check should happen earlier -- it duplicates some code in the `# Set osio configuration` section, and the `# Set minishift configuration` section depends on `CHE_OPENSHIFT_PROJECT` for `DEFAULT_OPENSHIFT_NAMESPACE_URL`

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,233 @@
...
+# Set configuration common to both minishift and openshift
+# --------------------------------------------------------
+
+DEFAULT_CHE_IMAGE_REPO="docker.io/rhchestage/che-server"
+CHE_IMAGE_REPO=${CHE_IMAGE_REPO:-${DEFAULT_CHE_IMAGE_REPO}}
+DEFAULT_CHE_IMAGE_TAG="nightly-fabric8"
+CHE_IMAGE_TAG=${CHE_IMAGE_TAG:-${DEFAULT_CHE_IMAGE_TAG}}
+DEFAULT_CHE_LOG_LEVEL="DEBUG"
+CHE_LOG_LEVEL=${CHE_LOG_LEVEL:-${DEFAULT_CHE_LOG_LEVEL}}
+DEFAULT_CHE_DEBUGGING_ENABLED="true"
```
amisevsk: Missing line? `CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}`

README.md
l0rd: Done

README.md
l0rd: Fixed thanks

README.md
l0rd: I haven't found a better way to describe that. Maybe we should just try to infer it without asking the user to provide it. We could list the projects and select the one that ends with "-che". What do you think?

dev-scripts/openshift_deploy.sh
l0rd: Fixed thanks

dev-scripts/openshift_deploy.sh
l0rd: Good catch @amisevsk. I've fixed that thanks.

dev-scripts/openshift_deploy.sh
l0rd: Yep you are 100% right. I've moved this code in the minishift config section.

l0rd:
Thanks @amisevsk for you review. I've addressed everything except the description for the `OPENSHIFT_ID` parameter because I don't have a better solution (except avoiding that parameter at all)

README.md
```diff
@@ -145,73 +145,64 @@ in the Minishift docker environment, as detailed in next section.
...
 ```
-__Important note:__ the `FABRIC8_ONLINE_PATH` should always end with a `/`.
 
-* For all the scripts that deploy or rollout to Minishift, you should have minishift running.
-This can be done with the following command:
+### Deploy to openshift.io 
 
 ```bash
-minishift start
+SCRIPT_URL=https://raw.githubusercontent.com/redhat-developer/rh-che/master/dev-scripts/openshift_deploy.sh
```
davidfestal: why do we need to download the deployment script that we have locally ?
why not:
```
export OPENSHIFT_FLAVOR=osio && dev-scripts/openshift-deploy.sh
```
at least in the readme ?

The download variant is interesting when not having the `rh-che` gitHub repo locally, but it's not the main case IMO.

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,234 @@
...
+  minishift status | grep -q "Running" ||(echo "Minishift is not running. Aborting"; exit 1)
+  echo "done!"  
+
+  DEFAULT_OPENSHIFT_ENDPOINT="https://$(minishift ip):8443/"
+  OPENSHIFT_ENDPOINT=${OPENSHIFT_ENDPOINT:-${DEFAULT_OPENSHIFT_ENDPOINT}}
+  DEFAULT_OPENSHIFT_USERNAME="developer"
+  OPENSHIFT_USERNAME=${OPENSHIFT_USERNAME:-${DEFAULT_OPENSHIFT_USERNAME}}
+  DEFAULT_OPENSHIFT_PASSWORD="developer"
+  OPENSHIFT_PASSWORD=${OPENSHIFT_PASSWORD:-${DEFAULT_OPENSHIFT_PASSWORD}}
+  DEFAULT_OPENSHIFT_NAMESPACE_URL="${CHE_OPENSHIFT_PROJECT}.$(minishift ip).nip.io"
```
davidfestal: shouldn't this line be *after* line 88 (where `CHE_OPENSHIFT_PROJECT` is defined) ?

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,234 @@
...
+
+# --------------------------------------------------------
+# Set configuration common to both minishift and openshift
+# --------------------------------------------------------
+
+DEFAULT_CHE_IMAGE_REPO="docker.io/rhchestage/che-server"
+CHE_IMAGE_REPO=${CHE_IMAGE_REPO:-${DEFAULT_CHE_IMAGE_REPO}}
+DEFAULT_CHE_IMAGE_TAG="nightly-fabric8"
+CHE_IMAGE_TAG=${CHE_IMAGE_TAG:-${DEFAULT_CHE_IMAGE_TAG}}
+DEFAULT_CHE_LOG_LEVEL="DEBUG"
```
davidfestal: why put `DEBUG` as the default log level ? We had switched back to `INFO` some weeks ago because `DEBUG` used to add too many logs by default in the wsmaster pod logs afair.

README.md
l0rd: Well I'm not sure. I think there will be more people that just want to run Che (on minishift) than people interested in building the source code. And for now the `README.md` in the github repository is the only source of documentation for people that want to deploy Che on OpenShift.

dev-scripts/openshift_deploy.sh
l0rd: Yes you are right. Will fix it

dev-scripts/openshift_deploy.sh
l0rd: Yes thanks will fix that too :-)

l0rd:
Thanks @davidfestal for your review. I've updated the PR.

README.md
davidfestal: Don't you think we could mention both use-cases ?

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,234 @@
...
+subjects:
+- kind: ServiceAccount
+  name: che" | oc apply -f -
+  oc login "${OPENSHIFT_ENDPOINT}" --token="${OPENSHIFT_TOKEN}"  &> /dev/null
+fi
+
+# ----------------------------------------------
+# Get latest version of fabric8-online templates
+# ----------------------------------------------
+echo -n "[CHE] Retrieving latest version of fabric8 online Che template..."
```
benoitf: question: how do we manage when users want to do changes on the fabric8 teemplates?
here it is downloading it from remote but how do we do when developer want to test changes of the template locally ?

dev-scripts/openshift_deploy.sh
```diff
@@ -0,0 +1,234 @@
...
+# --------------------------------------------------------
+
+DEFAULT_CHE_IMAGE_REPO="docker.io/rhchestage/che-server"
+CHE_IMAGE_REPO=${CHE_IMAGE_REPO:-${DEFAULT_CHE_IMAGE_REPO}}
+DEFAULT_CHE_IMAGE_TAG="nightly-fabric8"
+CHE_IMAGE_TAG=${CHE_IMAGE_TAG:-${DEFAULT_CHE_IMAGE_TAG}}
+DEFAULT_CHE_LOG_LEVEL="INFO"
+CHE_LOG_LEVEL=${CHE_LOG_LEVEL:-${DEFAULT_CHE_LOG_LEVEL}}
+DEFAULT_CHE_DEBUGGING_ENABLED="true"
+CHE_DEBUGGING_ENABLED==${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
```
benoitf: is it == ? for assigning CHE_DEBUGGING_ENABLED ?

README.md
```diff
@@ -145,73 +145,64 @@ in the Minishift docker environment, as detailed in next section.
...
-
-However, to explicitly enable the Keycloak integration, you can override the `CHE_KEYCLOAK_DISABLED` variable:
+#### openshift.io only options
 
-```bash
-bash -c "export CHE_KEYCLOAK_DISABLED=false ; dev-scripts/minishift_deploy.sh"
-```
+* `OPENSHIFT_ID`: Red Hat Developers username with some exceptions:
+  * If the username is an e-mail address the OPENSHIFT_ID will the part before the @ symbol (e.g. for cbrown@redhat.com `OPENSHIFT_ID` will be cbrown)
+  * Dots should be replaced by dashes ((e.g. for c.brown `OPENSHIFT_ID` will be c-brown)
```
benoitf: could it be done by the script ?

README.md
amisevsk: Hm, yeah, I'm not sure either. This is probably fine, and should cover 90%+ of cases. We could maybe include language about logging into OpenShift to find it, but that might be even less clear. 

This works fine, we can take a second look if it becomes an issue later.

README.md
amisevsk: Not reliably, these are just guidelines. I'd rather have nothing than something that doesn't work 100% of the time. 

The username generation has to be done to avoid conflicts, and since we're using email addresses, it'll be unpredictable when e.g. two users sign up with `email@domain1.com` and `email@domain2.com`.

benoitf: there is no way to get username from email address ?

l0rd: Since we just use it to get the right project name in osio I will just try:
```bash
DEFAULT_CHE_OPENSHIFT_PROJECT=$(oc get projects -o=custom-columns=NAME:.metadata.name --no-headers | grep "\-che$")
```

dev-scripts/openshift_deploy.sh
l0rd: We usually don't do that. But yes that maybe usufull if you need to test locally the yaml file. Let me add this option.

dev-scripts/openshift_deploy.sh
l0rd: Will fix it thanks

dev-scripts/openshift_deploy.sh
benoitf: it's because I had to it twice that I was thinking about my usecase (BTW I agree that it's not usual, but it would be nice to have alternative to grab templates)

README.md
l0rd: I've update the PR to reflect that

dev-scripts/openshift_deploy.sh
l0rd: Fixed

dev-scripts/openshift_deploy.sh
l0rd: I had a look at there are too many changes to do to make it in current PR and that may lead to regressions so I would prefer to postpone it to an distinct PR. I've justed added a todo in the script.

l0rd:
@benoitf I've addressed your comments and there should be no more shellcheck warnings.

dev-scripts/openshift_deploy.sh
benoitf: sure

PR #237, opened 2017-08-01
Inject keycloak user token to bayesian LS

sunix:
- Keeping the last token used (from Keycloak)AuthenticationFilter) and make it available to the bayesian LS at start time.
- update `KeycloakAuthenticationFilter` so it works well with new upstream websocket URIs

### DIFF
--- a/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
+++ b/builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
@@ -69,6 +69,7 @@ public class KeycloakAuthenticationFilter extends org.keycloak.adapters.servlet.
 
         HttpServletRequest request = (HttpServletRequest) req;
         String authHeader = request.getHeader("Authorization");
+        String upgradeHeader = request.getHeader("Upgrade");
         String requestURI = request.getRequestURI();
         String requestScheme = req.getScheme();
 
@@ -76,9 +77,10 @@ public class KeycloakAuthenticationFilter extends org.keycloak.adapters.servlet.
             LOG.debug("No 'Authorization' header for {}", requestURI);
         }
 
-        if (isSystemStateRequest(requestURI) || isWebsocketRequest(requestURI, requestScheme)
+        if (isSystemStateRequest(requestURI) || isWebsocketRequest(requestURI, requestScheme, upgradeHeader)
                 || isKeycloakSettingsRequest(requestURI) || isWorkspaceAgentRequest(authHeader)
-                || isRequestFromGwtFrame(requestURI) || isStackIconRequest(requestURI)) {
+                || isRequestFromGwtFrame(requestURI) || isStackIconRequest(requestURI)
+                || isWsAgentUserTokenRequest(requestURI)) {
             LOG.debug("Skipping {}", requestURI);
             chain.doFilter(req, res);
         } else if (userChecker.matchesUsername(authHeader)) {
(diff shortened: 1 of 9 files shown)


builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -41,7 +42,10 @@
     private boolean keycloakDisabled;
 
     @Inject
-    private KeycloakUserChecker userChecker;
+    private KeycloakTokenStore         kcTokenStore;
```
tsmaeder: Field injection requires the use of some injection framework when writing tests. Personally, I'm not a fan. Why not use constructor injection? Is there an opinion on the Che team?

sunix: Personally, I find it more readable than useless constructor and parameters in constructor when it's actually doing nothing else then just setting the value to the field. Using injection or mock framework is not a problem to me.

sunix:
I've added https://github.com/redhat-developer/rh-che/pull/237/commits/45acdcfb5668c9b217121974ca9e905b3baeb0d9 so it works well with the latest master

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -128,10 +135,14 @@ private boolean isWorkspaceAgentRequest(String authHeader) {
...
-                || requestScheme.equals("wss") || requestURI.contains("/websocket/")
-                || requestURI.endsWith("/token/user");
+    private boolean isWsAgentUserTokenRequest(String requestURI) {
+        return requestURI.endsWith("/token/user");
+    }
+
+    private boolean isWebsocketRequest(String requestURI, String requestScheme, String upgradeHeader) {
+        return "websocket".equals(upgradeHeader) && //
+               (requestURI.endsWith("/ws") || requestURI.endsWith("/eventbus") || requestScheme.equals("ws") || requestScheme.equals("wss")
+                || requestURI.contains("/websocket") || requestURI.contains("/wsagent"));
```
davidfestal: it seems that, if you explicitly add the `websocket` upgrade header (in a curl request for example), you can access to many  wsagent rest endpoints without any security.

sunix: there is a `&&`

davidfestal: sure, but the condition:
```
"websocket".equals(upgradeHeader) && requestURI.contains("/wsagent")
```
seem to allow many wsagent requests, due to the use of the `contains()` method.

sunix: ok so the problem is not about websocket upgrade header :)

sunix: fixed

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
```diff
@@ -128,10 +135,14 @@ private boolean isWorkspaceAgentRequest(String authHeader) {
...
-    private boolean isWebsocketRequest(String requestURI, String requestScheme) {
-        return requestURI.endsWith("/ws") || requestURI.endsWith("/eventbus") || requestScheme.equals("ws")
-                || requestScheme.equals("wss") || requestURI.contains("/websocket/")
-                || requestURI.endsWith("/token/user");
+    private boolean isWsAgentUserTokenRequest(String requestURI) {
+        return requestURI.endsWith("/token/user");
+    }
+
+    private boolean isWebsocketRequest(String requestURI, String requestScheme, String upgradeHeader) {
+        return "websocket".equals(upgradeHeader) && //
```
ibuziuk: @sunix are you sure that all websocket requests **must** contain "Upgrade" header ?

ibuziuk: also, it is not clear for me why are you using // ?

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
davidfestal: thanks !

builds/fabric8-che/plugins/keycloak-plugin-server/src/main/java/com/redhat/che/keycloak/server/KeycloakAuthenticationFilter.java
sunix: yes websocket is an upgrade of http
`//` is a trick to keep carriage return even if the formatter wouldn't do so
so it's to make the code more readable and avoid other people breaking the formatting chosen

Issue #241, opened 2017-08-02
dashboard build fail because of dist folder is not copied.


Jooho:
Hi, 

When I build redhat che using dev-scripts/build_fabric8.sh, dashboard build always fail because it 
looks for index.html to replace. However, the index.html under dist folder is not on the list to copy to ```/rh-che/target/builds/fabric8/fabric8-che/assembly/dashboard/target/patched/```

In order to fix this issue, dist folder should be added to patch list in ```/rh-che//builds/fabric8-che/assembly/dashboard/pom.xml```

```
121  <include name="e2e/**" />
122  <include name="dist/**" />    <=== ADD
```
Thanks,
Jooho Lee

amisevsk:
Hi @Jooho, I've just checked out a clean copy of rh-che and run `/rhche/dev-scripts/build_fabric8.sh`, and the build completed successfully. Could you clarify the issue you are facing?

l0rd:
Thanks @Jooho! Please provide us the steps to reproduce the issue because we have never been able to see that before.

Jooho:
Sure. 

This is full steps that I did.  I create RHEL 7U3 vm and then did following steps:

~~~
export tmp_dir=${HOME}/new/test
mkdir ${tmp_dir}
cd ${tmp_dir}

#in order to use internet from container
sysctl -w net.ipv4.ip_forward=1

# install necessary packages

yum install -y docker git wget net-tools  bridge-utils bind-utils bzip2 patch gcc bash-completion    java-1.8.0-openjdk java-1.8.0-openjdk-debug java-1.8.0-openjdk-devel java-1.8.0-openjdk-headless java-1.8.0-openjdk-headless-debug  

# oc command
yum install atomic-openshift-clients

#yum install fontconfig freetype freetype-devel fontconfig-devel libstdc++ 

# maven install 
wget http://apache.mirror.rafal.ca/maven/maven-3/3.5.0/binaries/apache-maven-3.5.0-bin.tar.gz
tar xvf apache-maven-3.5.0-bin.tar.gz
mv apache-maven-3.5.0 maven


# Install nodejs(https://nodejs.org/en/download/)
wget https://nodejs.org/dist/v4.4.6/node-v4.4.6-linux-x64.tar.xz
tar xvf node-v4.4.6-linux-x64.tar.xz
mv node-v4.4.6-linux-x64 nodejs
sudo chown -R $(whoami) $(npm config get prefix)/{lib/node_modules,bin,share}


[...]

PR #242, opened 2017-08-02
Dashboard build fail because of no dist folder

Jooho:
This PR is for [#241 Issue](https://github.com/redhat-developer/rh-che/issues/241)
Added ```<include name="dist/**" /> ``` to copy list.


l0rd:
Thanks @Jooho for this PR. However we are going to close it since the issue has already been solved with another PR https://github.com/redhat-developer/rh-che/pull/268.

PR #243, opened 2017-08-02
Integrate cleanup and rollupdate commands in openshift_deploy script

l0rd:
This PR add one option to `openshift_deploy.sh`: 

```bash
[-c|--command] <cmd> # where <cmd> can be `deploy`, `cleanup` or `rollupdate`
```

With the command option we should replace `minishift_clean.sh` and `minishift_rollupdate.sh`.

Moreover the new cleanup procedure (`openshift_deploy.sh -c cleanup`) doesn't delete PVs anymore. That was something that we were doing with `minishift_clean.sh` but is not necessary anymore on minishift > v1.0.0.


benoitf:
it may not be related directly to this change but:
- could we move all scripts which are at the root folder to an inner folder (because they're not useful for end users)
- put this script in root folder of rh-che
- rename it to rh-che.sh or che.sh (because it's odd for me to use "openshift_deploy.sh" script it may let me think that I'm deploying "openshift" while I want to deploy che.
- as the command may be deploy, cleanup, etc it's also odd to do "./openshift_deploy cleanup" (see my previous remark)
- so I could do : 
      - `rhche.sh deploy` (with openshift or minishift flavor)
      - `rhche.sh cleanup` (with openshift or minishift flavor)
      - `rhche.sh rollupdate` (with openshift or minishift flavor)
      - (and maybe commands to build locally the stuff before deploying it)

benoitf:
I'm approving the current PR as it does what it is saying :-) but I added some global feedback

l0rd:
@benoitf agree with you. This PR is a small step towards that goal needed to fix cleanup that is currently broken.

Issue #131, opened 2017-06-27
Add option to start workspaces on a different namespace


ibuziuk:
see - https://issues.jboss.org/browse/CHE-100

l0rd:
One issue is that wsmaster-wsagent authentication as it is implemented now will fail. To solve this issue we need to figure out how authentication will work. Maybe with Sergii team.

PR #244, opened 2017-08-03
rh-che 131 - Changes necessary to enable running workspace on different namespace

amisevsk:
The only change of substance is that instead of hard-coding `http://che-host:8080/` as the Che server API, we instead get the value from the CHE_API environment variable, which is added by Che server when starting the workspace. 

With this change it is possible to run Che and start workspaces in another namespace on minishift, with a few caveats:
1. Manual setup is required: Necessary steps are
    - Create a second namespace
    - Ensure PVC `claim-che-workspace` is bound to the second namespace instead of `eclipse-che`
    - Update che deployment config so that `CHE_OPENSHIFT_PROJECT` is `<second-namespace>`
    - Add admin rights to che service account on `<second-namespace>`: `oc adm policy add-role-to-group admin system:serviceaccounts -n <second-namespace>`
2. When running on minishift with the main project name `eclipse-che`, there are no issues, however keycloak authentication for messages between wsmaster and wsagent does not currently work when enabled. I expect this will be easier to implement once Che server is multitenant, as currently there's no good way of validating that incoming messages are what we expect.

Addresses #131 
Depends on https://github.com/eclipse/che/pull/5884

### DIFF
--- a/builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
+++ b/builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
@@ -46,14 +46,29 @@ import com.redhat.che.keycloak.shared.KeycloakSettings;
  * @author amisevsk
  */
 public class UserAuthValve extends KeycloakAuthenticatorValve {
-    
+
     private static final Log LOG = LogFactory.getLog(UserAuthValve.class);
-    private static final String API_ENDPOINT = "http://che-host:8080/api";
-    private static final String USER_VALIDATOR_ENDPOINT = API_ENDPOINT + "/token/user";
+
     private static final String AUTHORIZATION_HEADER = "Authorization";
-    
+
+    /** Name of env var that holds address of Che Host on the cluster */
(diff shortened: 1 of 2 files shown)


builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
```diff
@@ -46,14 +46,27 @@
  * @author amisevsk
  */
 public class UserAuthValve extends KeycloakAuthenticatorValve {
-    
+
     private static final Log LOG = LogFactory.getLog(UserAuthValve.class);
-    private static final String API_ENDPOINT = "http://che-host:8080/api";
-    private static final String USER_VALIDATOR_ENDPOINT = API_ENDPOINT + "/token/user";
+
+    /** Name of env var that holds address of Che Host on the cluster */
```
davidfestal: shouldn't this comment be the line below ?

amisevsk: It should, I messed up while rearranging the static fields. Thanks -- fixed.

PR #263, opened 2017-08-11
use another pattern than / character for sed as / is used by the value

benoitf:
For example properties(ENDPOINT URLs) contains / character so it fails. Using | character is fine

error I get is "bad flag in substitute command /"

I've tested on my computer but you should validate on different Operating System than mine :-)

Change-Id: I681db6831acb6aacf9e609021574839ecba4803f


benoitf:
@ibuziuk yes you're right ! I've updated

PR #270, opened 2017-08-16
Use localhost instead of che-host service name for checking KC

benoitf:
Make UserAuthValve and KeycloakUserChecker use localhost instead
of che-host for checkout keycloak tokens when CHE_API env var
is not set

Signed-off-by: Angel Misevski <amisevsk@redhat.com>

### DIFF
--- a/builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
+++ b/builds/fabric8-che/assembly/user-auth-valve/src/main/java/com/redhat/che/valve/UserAuthValve.java
@@ -62,7 +62,7 @@ public class UserAuthValve extends KeycloakAuthenticatorValve {
         if (cheApiEndpoint != null) {
             CHE_API_ENDPOINT = cheApiEndpoint.replaceAll("/wsmaster/api", "/api");
         } else {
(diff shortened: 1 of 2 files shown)


amisevsk:
This is a work around for the issue of che-host not being able to communicate with its own service on minishift 1.4.1 + OS 3.6. As an alternative, this [comment](https://github.com/minishift/minishift/issues/1269#issuecomment-322781706) worked for me.

sunix:
I don't know what's the problem and if this is related but ... I still cannot access to che-server (the server is starting) with minishift 1.4.1 with this patch
I've got this at start up of the script openshift-deploy.sh

```
[CHE] Creating an OS route to debug Che wsmaster...error: [the provided version "apps.openshift.io/v1" has no relevant versions: group apps.openshift.io has not been registered, no matches for apps.openshift.io/, Kind=DeploymentConfig]
See 'oc expose -h' for help and examples.
```

when accessing to the che-server url from the browser, i've got
```
504 Gateway Time-out

The server didn't respond in time.
```

Though these error are gone with  `ifconfig docker0 promisc` I have other issue when using the workspace ( got .error 404 on PUT on activity.js ) 

Maybe all my issue are related to keycloak + minishift 1.4.1

Issue #249, opened 2017-08-04
Simplify rh-che build: remove upstream build and align to Che upstream release


sunix:

sunix:
Creating 3 PR for adding sources as maven artifact:
 - dashboard :[che#5971](https://github.com/eclipse/che/pull/5971)
 - assembly-ide-war : Adding missing source folders to assembly-ide-war sources Maven artifact [che#5966](https://github.com/eclipse/che/pull/5966) and [che#5979](https://github.com/eclipse/che/pull/5979)

Sources will be downloaded through maven, unpacked and used for performing patches

PR #275, opened 2017-08-17
Simplifying build (#249)

sunix:
Fix https://github.com/redhat-developer/rh-che/issues/249

- This PR is removing the clone and build of upstream eclipse/che
- Maven sources artifacts are used when the build process need to perform patches (instead of patching upstream project)
- Maven classifier is used to dissociate with and without dasboard assemblies
- Maven invoker is not used anymore
- Docker build is not using upstream build, based on top of eclipse/che-server:nightly-centos and including /run-pmcd.sh script patched to the entrypoint
- README file has been updated


dockerfiles/build.include
```diff
@@ -0,0 +1,151 @@
+#!/bin/sh
```
benoitf: do you know why we had to fork the build.include script of che ?

sunix: We don't have to fork :) but it was simpler at the stage where I am. Maybe we can download/curl it ? or do you have any other suggestion ?

benoitf: if it's because you need the files, we may have a pom.xml in dockerfiles/ as well to embed the sources

sunix: Yes, i will work on this option

sunix: finally, I prefer to keep this file in this PR ( and prioritize another issue for this ) as this script is not production code and would not change a lot and is quite tiny

builds/fabric8-che/assembly/assembly-main/src/assembly/base-component.xml
```diff
@@ -0,0 +1,58 @@
...
+        <fileSet>
+            <directory>${project.build.directory}/keycloak/</directory>
+            <outputDirectory>tomcat/lib</outputDirectory>
+        </fileSet>
+        <fileSet>
+            <directory>${project.build.directory}/stacks/</directory>
+            <outputDirectory>stacks/</outputDirectory>
+        </fileSet>
+    </fileSets>
+</component>
```
riuvshin: new line missing

dockerfiles/build.include
```diff
@@ -0,0 +1,151 @@
...
+  OUTPUT_PATH=${OUTPUT_PATH//\"}
+  echo ${OUTPUT_PATH}
+}
+
+get_mount_path() {
+  FULL_PATH=$(get_full_path "${1}")
+  POSIX_PATH=$(convert_windows_to_posix "${FULL_PATH}")
+  CLEAN_PATH=$(get_clean_path "${POSIX_PATH}")
+  echo $CLEAN_PATH
+}
```
riuvshin: missing new line

dockerfiles/che-fabric8/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY pcp.repo /etc/yum.repos.d/pcp.repo
+RUN yum install -y pcp patch && yum clean all
+COPY ./run-pmcd.sh /run-pmcd.sh
+RUN chmod a+x     /run-pmcd.sh
+RUN mkdir -p      /etc/pcp /var/run/pcp /var/lib/pcp /var/log/pcp
+RUN chmod -R 0777 /etc/pcp /var/run/pcp /var/lib/pcp /var/log/pcp
+EXPOSE 44321
+
+COPY entrypoint.patch /entrypoint.patch
+RUN patch -p0 < entrypoint.patch
```
riuvshin: missing new line

dockerfiles/che-fabric8/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [ -f "${LOCAL_ASSEMBLY_ZIP}" ]; then
+  rm "${LOCAL_ASSEMBLY_ZIP}"
+fi
+
+echo "Linking assembly ${BUILD_ASSEMBLY_ZIP} --> ${LOCAL_ASSEMBLY_ZIP}"
+ln "${BUILD_ASSEMBLY_ZIP}" "${LOCAL_ASSEMBLY_ZIP}"
+
+init --name:server "$@" 
+build
```
riuvshin: missing new line

builds/fabric8-che/pom.xml
```diff
@@ -18,7 +18,6 @@
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
         <version>5.18.0-SNAPSHOT</version>
```
riuvshin: Afaik we wanted to use tag version of upstream not snapshot or this is separate thing ?

config
```diff
@@ -25,3 +19,5 @@ export RH_CHE_DOCKER_IMAGE_VERSION=${RH_CHE_DOCKER_IMAGE_VERSION:-${DEFAULT_RH_C
 export DOCKER_HUB_USER=${DOCKER_HUB_USER:-${RHCHEBOT_DOCKER_HUB_USER}}
 export DOCKER_HUB_PASSWORD=${DOCKER_HUB_PASSWORD:-${RHCHEBOT_DOCKER_HUB_PASSWORD}}
 export DOCKER_HUB_NAMESPACE=${DOCKER_HUB_NAMESPACE:-${RH_CHE_DOCKER_HUB_NAMESPACE}}
+
+export CHE_DOCKER_BASE_IMAGE=eclipse/che-server:5.17.0-centos
```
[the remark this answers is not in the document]
sunix: the problem is that current `5.18.0-snapshot` has the docker tag `nightly-centos` but after the next release maven version will change to `5.19.0-snapshot` but docker tag will remain unchanged `nightly-centos`. So at the moment we can't really sync this. But tagged version are using the version so it would work if we had a 5.18.0-SNAPSHOT-centos upstream but we don't have at the moment. I think it is simple enough with that so we don't over-engineering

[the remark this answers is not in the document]
davidfestal: OK

builds/fabric8-che/assembly/assembly-wsagent-war/pom.xml
```diff
@@ -11,13 +11,12 @@
...
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <artifactId>fabric8-ide-assembly-parent</artifactId>
         <groupId>com.redhat.che</groupId>
-        <version>@redhat.che.version@</version>
+        <version>1.0.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
sunix: We generate the docker image whatever the maven version by just picking up the right assembly in the right folder and with the right name.
Versions in maven are important if we are performing tag/release. So if we do, I'm happy to keep the exact same version in rh-che and che.
After docker generates several tags and the one that previously had che upstream version seems not to used by `openshift_deploy.sh`. So let me know if the build is actually working another way but I doubt it has to detect the che upstream version.
So I'm not sure that even if we keep that version, we can see it. So to me it's useless to keep something that we can not use. Let me know if i'm wrong.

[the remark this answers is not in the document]
davidfestal: @l0rd what's your thought on this ?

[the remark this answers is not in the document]
sunix: @l0rd @davidfestal https://github.com/redhat-developer/rh-che/pull/275/commits/4cafbb37d07db0d18ff84fee3a0e341b70b84f51

Adding back upstream revision using BuildInfo.properties generated by upstream che build

PR #277, opened 2017-08-21
Update commands for spring boot

benoitf:
- Update build command to remove the clean part
- Add a new 'clean build' command which do a clean before the build

It fixes https://github.com/openshiftio/openshift.io/issues/621

Change-Id: I18dfb7df8537048ecb71948022a7604ced12921c


l0rd:
@benoitf don't we need to update (upstream stacks.json)[https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json] too?

ibuziuk:
@l0rd upstream is also update - https://github.com/eclipse/che/pull/6050

l0rd:
Thanks @ibuziuk LGTM

PR #288, opened 2017-08-23
Updating `che.version`  property

ibuziuk:


benoitf:
maybe we need a script to update properties in all expected file

PR #255, opened 2017-08-08
Add openshift container platform flavor

Jooho:
Hi,

openshift-deploy.sh contains minishift and openshift.io only but many clients are interested in deploying it to openshift container platform. Therefore, I add openshift container platform flavor.

Thanks,
Jooho Lee.

Issue #284, opened 2017-08-21
Move openshift deployment script and documentation to upstream


l0rd:
After latest community call we have concluded that the deployment script and doc to deploy Che on OpenShift should be moved upstream. 

The tasks to achieve that are the following:

- [x] Extending `openshift_deploy.sh` to deploy Che on OCP (c.f. the already existing PR #255)
- [x] Moving `openshift_deploy.sh` to upstream Che
- [x] Moving the documentation to deploy Che on OpenShift to [eclipse.org/che/docs](http://www.eclipse.org/che/docs/)
- [x] Adding in the documentation the instructions to add a custom stack when running Che on OpenShift (the Stack should be based on an existing Docker image)

l0rd:
cc @bmicklea

PR #290, opened 2017-08-23
Add OpenShift Container Platform support in openshift_deploy.sh

l0rd:
This PR is a rebase of @Jooho #255 with some fixes


l0rd:
This PR is first step of  #284

PR #306, opened 2017-08-29
Update regex for OS user to name converter

amisevsk:
OSIO usernames are derived from Red Hat Developers usernames. Some older users have usernames that are emails, or contain characters that are forbidden in OpenShift namespaces. This means that keycloak authentication will currently fail for certain users (e.g. those with an underscore character in their username).

This PR puts the conversion between username and project name in sync with the fabrici8-tenant.


l0rd:
Thanks @amisevsk. Can you please add a test for the underscore?

amisevsk:
@l0rd Good point! Done.

aslakknutsen:
Related https://github.com/openshiftio/openshift.io/issues/619

PR #309, opened 2017-08-30
new docker push endpoints at registry.devshift.net

jfchevrette:
registry.devshift.net is going pull-only

the new endpoint to push images from trusted sources is push.registry.devshift.net


l0rd:
@jfchevrette not really related to this PR but asking now before I forgot: is there the possibility to define any kind of webhook in `devshift.net`?

PR #313, opened 2017-09-04
Login before pushing to devshfit docker registry

l0rd:
Signed-off-by: Mario Loriedo <mario.loriedo@gmail.com>


vpavlin:
Please add `DEVSHIFT` to grep in https://github.com/redhat-developer/rh-che/blob/master/cico_build.sh#L13 and I think this will be ready to go

l0rd:
@vpavlin I've updated the PR

PR #317, opened 2017-09-05
Clean when building for the second time

l0rd:
Signed-off-by: Mario Loriedo <mario.loriedo@gmail.com>


benoitf:
BTW it would be nice if we could avoid the clean step

Issue #315, opened 2017-09-04
Update scripts to use new maven central link


l0rd:
Artifacts for https://github.com/fabric8-services/fabric8-tenant-che are now at http://central.maven.org/maven2/io/fabric8/tenant/tenant-che/.

Old github repo https://github.com/fabric8io/fabric8-online and maven central link http://central.maven.org/maven2/io/fabric8/online/apps/che/ are deprecated and should not be used anymore.

PR #320, opened 2017-09-05
Update maven central URL for che yaml

l0rd:
This fixes https://github.com/redhat-developer/rh-che/issues/315


l0rd:
Thanks @ibuziuk I've fixed that too

Issue #295, opened 2017-08-25
Use image from devshift.net Docker registry


l0rd:
- For che-server image deployment yml files
- For che stacks images in stacks.json

l0rd:
I've tested successfully that che-server images from devshift.net works well on minishift and osio. The problem is that after commit cd3b12cc58ccbfface40c8388838566d1961b7ef we cannot retrieve devshift images anymore. 

#313 and https://github.com/almighty/almighty-jobs/pull/284 have not solved the problem: script cico_do_docker_build_tag_push.sh hangs at line https://github.com/redhat-developer/rh-che/blob/master/cico_do_docker_build_tag_push.sh#L68.

I am currently working with @vpavlin to solve this.

l0rd:
UPDATE: I've currently disabled pushing images to devshift to investigate the problem and it looks like `DEVSHIFT_USERNAME` and `DEVSHIFT_PASSWORD` are not set

> DEVSHIFT_USERNAME is unset.
> DEVSHIFT_PASSWORD is unset.

l0rd:
This was fixed by #327 but we needed to revert it because of an issue with the lenght of chars in image tag (https://github.com/openshiftio/openshift.io/issues/798). As soon as eclipse/che#6245 get merged and a new version of Che is released we should apply c0bcc62 back again.

Issue #80, opened 2017-05-29
Push to registry.devshift.net only if verification passes


mlabuda:
We are, for now, pushing to devshift registry before execution of verification job. That means we are providing "prod" images there, because images to devshift are not retagged after verification and no additional push is performed.

l0rd:
Fixed by #324

PR #324, opened 2017-09-06
Fix and re-enable pushing images to devshift

l0rd:
Related to #295. Variables `DEVSHIFT_USERNAME` and `DEVSHIFT_PASSWORD` were not set when running script `cico_do_docker_build_tag_push.sh` because they were not exported in `cico_build.sh`. Now explicitly export them.

### DIFF
--- a/cico_build.sh
+++ b/cico_build.sh
@@ -12,9 +12,9 @@ then
   set +x
   cat jenkins-env | grep -e PASS -e DEVSHIFT > inherit-env
   . inherit-env
-  if [ -z "${DEVSHIFT_USERNAME+x}" ]; then echo "WARNING: failed to get DEVSHIFT_USERNAME from jenkins-env file in centos-ci job."; fi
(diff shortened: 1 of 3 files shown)


vpavlin:
Maybe you want to fail if you cannot push to registry.devshift.net? Otherwise you'll try to deploy something that is not in the registry..

l0rd:
@vpavlin I've done what you have suggested. I have also moved pushing to devshift only after validation of the build (in script `cico_deploy.sh`). That was something we needed to do anyway (c.f. #80). 

With these changes we should get closer to switching to devshfit as our main (only?) docker registry. 

What is missing to get rid of the Docker Hub are:

1. Set a webhook for che/che that points to 
2. Create a new stage repository (e.g. `che-stage/che`) to push images that will go through functional tests before beeing promoted.

And as a bonus (not really necessary but nice to have):

3. Rename `che/che` to `che/che-server`. This is a more appropriate name (upstream images `eclipe/che` and `eclipse/che-server` both exist and `eclipse/che` is a CLI utility).

Issue #291, opened 2017-08-24
Vertx stack debug command should include the compile goal


sunix:
Feedback from osio workshop:
vertx stack debug command should include the compile goal
See screenshot:
![selection_056](https://user-images.githubusercontent.com/650571/29672802-8d29e3c4-88ed-11e7-839e-508dc3af7440.png)

otherwise it won't take the latest changes in account when running debug command

burrsutter:
We fixed this once before, did it go away?

kameshsampath:
compile and package goals are must from the mvn command perspective as part of vertx-maven-plugin we don't tie to any maven lifecycle phase

PR #330, opened 2017-09-08
Compile goal is now included in vertx stack debug command

JPinkney:
Adds compile goal to vertx stack debug command.

New command is: "scl enable rh-maven33 'mvn compile vertx:debug -f ${current.project.path}'"


l0rd:
@JPinkney thank you and congratulations for your first PR! 
One [stacks.json file is also maintained upstream](https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L2375). You should consider creating a PR on that github repo too.

sunix:
Fix https://github.com/redhat-developer/rh-che/issues/291

Issue #329, opened 2017-09-08
cleanup copyright headers


benoitf:
There are codenvy copyright headers or broken copyright headers in the current rh-che files

Upstream che has a plugin to format/check the files so we could enable this plugin to ensure we don't have "odd" copyrights

l0rd:
Who take this should also probably work on #192 too

JPinkney:
Fixed as part of #332

Issue #192, opened 2017-07-13
No license for rh-che project


sunix:
There is no LICENSE file nor license header in source code files (java, xml ...)
Which license are we suppose to use ?

l0rd:
As for the license type I would say [EPL](http://www.eclipse.org/legal/epl-v10.html). @gorkem ? 

For the copyright headers I think most of the java source files already have one and we should verify it and enforce a check at build time (for .xml, .sh etc...too of course).

gorkem:
Yes, EPL is the natural choice.

JPinkney:
Fixed as part of #332

PR #332, opened 2017-09-12
Fixed license header files and enforced check at build time

JPinkney:
Fixed license header files and enforced check and build time.

Relevant issues: #329 and #192 

Signed-off-by: jpinkney <josh.pinkney@mail.utoronto.ca>


l0rd:
Great @JPinkney! That was something that needed to be fixed since a long time. Can you please add copyright header to `*.sh` and `Dockerfile` as well?

builds/fabric8-che/pom.xml
```diff
@@ -265,4 +265,22 @@
             <url>https://maven.codenvycorp.com/content/repositories/codenvy-public-snapshots/</url>
         </pluginRepository>
     </pluginRepositories>
+    <build>
```
sunix: Why do we need that ? isn't the license-maven-plugin inherited from upstream project ?

Also you should remove this part: https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/plugins/che-plugin-bayesian-lang-server/pom.xml#L84-L92

JPinkney: You were right, Its inherited from the upstream project. I also removed that part from the Bayesian language server pom.xml, I'm updating now

sunix: great :+1:

PR #336, opened 2017-09-13
Removing non-working scripts

ibuziuk:


ibuziuk:
@davidfestal those are not working because `update_openshift_connector.sh` was removed in one of the previous commits during build simplification task

davidfestal:
@ibuziuk Ah, OK. Thanks for the explanation !

PR #340, opened 2017-09-18
allow to disable SSL for ocp flavor

riuvshin:
allow to disable SSL for ocp flavor


riuvshin:
@l0rd can you please review this

dev-scripts/openshift_deploy.sh
```diff
@@ -141,6 +141,17 @@ elif [ "${OPENSHIFT_FLAVOR}" == "ocp" ]; then
   DEFAULT_CHE_DEBUGGING_ENABLED="false"
   CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
 
+elif [ "${OPENSHIFT_FLAVOR}" == "ocpdev" ]; then
```
l0rd: Instead of adding a new flavour you may want to add a new env variable (e.g. `ENABLE_SSL`) and based on it's value you decide what value will have `che-openshift-secure-routes` and `che-secure-external-urls`. This is already done for env variable [CHE_KEYCLOAK_DISABLED](https://github.com/redhat-developer/rh-che/blob/master/dev-scripts/openshift_deploy.sh#L263) for example.

riuvshin: Sure I could do that, but my problem that I have no idea which exactly options should be changed  or enabled / disabled.  Can you please help me figure what should I wrap in `ENABLE_SSL` condition ?

l0rd: @riuvshin you should:

Set variable `${ENABLE_SSL}` in the first part of the script (variable `${DEFAULT_ENABLE_SSL}` should be true for osio and false otherwise):

```bash
ENABLE_SSL=${ENABLE_SSL:-${DEFAULT_ENABLE_SSL}}
```

And then set `che-secure-external-urls` and `che-openshift-secure-routes` based on the value of ${ENABLE_SSL} (the following lines should be added after lines [286](https://github.com/redhat-developer/rh-che/blob/master/dev-scripts/openshift_deploy.sh#L286) and [296](https://github.com/redhat-developer/rh-che/blob/master/dev-scripts/openshift_deploy.sh#L296), and replace lines [269 and 270](https://github.com/redhat-developer/rh-che/blob/master/dev-scripts/openshift_deploy.sh#L269):

```bash
if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    che-openshift-secure-routes: \"true\"/    che-openshift-secure-routes: \"false\"/" ; else cat -; fi | \
if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    che-secure-external-urls: \"true\"/    che-secure-external-urls: \"false\"/" ; else cat -; fi | \
```

With this changes If you want to run Che on ocp **without** SSL you just need to:

```bash
export ENABLE_SSL=false
export OPENSHIFT_FLAVOR=ocp
./openshift_deploy.sh
```

l0rd: Oh actually the lines to disable SSL should be 3:

```bash
if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    che-openshift-secure-routes: \"true\"/    che-openshift-secure-routes: \"false\"/" ; else cat -; fi | \
if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    che-secure-external-urls: \"true\"/    che-secure-external-urls: \"false\"/" ; else cat -; fi | \
if [ "${ENABLE_SSL}" == "false" ]; then grep -v -e "tls:" -e "insecureEdgeTerminationPolicy: Redirect" -e "termination: edge" ; else cat -; fi | \
```

And you can probably merge the 3 `if` conditions in 1 unique `if`.

riuvshin: thanks @l0rd for the detailed explanation, I did what you have proposed but when I run a workspace pop up showed up and says that unable to connect to ide. I figured out that it uses `https` protocol to reach ws-agent and fails.
so I added this
```
if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    che.docker.server_evaluation_strategy.custom.external.protocol: https/    che.docker.server_evaluation_strategy.custom.external.protocol: http/" ; else cat -; fi | \
```

after that IDE is able to load but import project is failed with:

```
Project: web-java-spring Sources cannot be cloned
The system is not configured to trust the security certificate provided by the Git server
```

It seems we have to tune something more but I don't know what exactly

riuvshin:
@l0rd pr updated, I've applied it only to `ocp` because on minishift it always be like same and `osio` will always be secure

riuvshin:
@l0rd updated

l0rd:
@riuvshin LGTM but you need to rebase your PR sorry 😞

benoitf:
@l0rd using "squash and merge" would avoid to rebase

Issue #311, opened 2017-09-01
Use Logback JsonLayout for Che server logs on OSIO


l0rd:
Using @aslakknutsen words:

> We're currently looking at getting the CheHost logs etc pushed to the OSD Elastic as well. It would have been nice to get CheHost to log json so we don't have to deal with multi line exceptions etc

sunix:
https://github.com/redhat-developer/rh-che/pull/341

sunix:
upstream issue:
https://github.com/eclipse/che/issues/6537

ibuziuk:
need to test with rh-che multi-user branch

PR #341, opened 2017-09-19
Logging in json format using logstash-logback-encoder (#311)

sunix:
This is logging into `/data/logs/catalina.log.json` and then archiving in `/data/logs/archive/2017/09/19/catalina.log.json`

For now, by default, it looks like:
```
{"@timestamp":"2017-09-19T06:48:11.877+00:00","@version":1,"message":"Database: jdbc:h2:che (H2 1.4)","logger_name":"org.flywaydb.core.internal.dbsupport.DbSupportFactory","thread_name":"localhost-startStop-1","level":"INFO","level_va
lue":20000}                                                                                                                                                                                                                               
{"@timestamp":"2017-09-19T06:48:12.957+00:00","@version":1,"message":"Flyway 4.0.3 by Boxfuse","logger_name":"org.flywaydb.core.internal.util.VersionPrinter","thread_name":"localhost-startStop-1","level":"INFO","level_value":20000}   
{"@timestamp":"2017-09-19T06:48:12.965+00:00","@version":1,"message":"Database: jdbc:h2:che (H2 1.4)","logger_name":"org.flywaydb.core.internal.dbsupport.DbSupportFactory","thread_name":"localhost-startStop-1","level":"INFO","level_va
lue":20000}                                                                                                                                                                                                                               
{"@timestamp":"2017-09-19T06:48:13.058+00:00","@version":1,"message":"Searching for sql scripts in locations [classpath:che-schema]","logger_name":"org.eclipse.che.core.db.schema.impl.flyway.CustomSqlMigrationResolver","thread_name":"
localhost-startStop-1","level":"INFO","level_value":20000}                                                                                                                                                                                
{"@timestamp":"2017-09-19T06:48:13.166+00:00","@version":1,"message":"Successfully validated 11 migrations (execution time 00:00.110s)","logger_name":"org.flywaydb.core.internal.command.DbValidate","thread_name":"localhost-startStop-1
","level":"INFO","level_value":20000}                                                                                                                                                                                                     
{"@timestamp":"2017-09-19T06:48:13.262+00:00","@version":1,"message":"Current version of schema \"PUBLIC\": 5.11.0.1","logger_name":"org.flywaydb.core.internal.command.DbMigrate","thread_name":"localhost-startStop-1","level":"INFO","l
evel_value":20000}                                                                                                                                                                                                                        
{"@timestamp":"2017-09-19T06:48:13.273+00:00","@version":1,"message":"Schema \"PUBLIC\" is up to date. No migration necessary.","logger_name":"org.flywaydb.core.internal.command.DbMigrate","thread_name":"localhost-startStop-1","level"
:"INFO","level_value":20000}                                                                                                                                                                                                              
{"@timestamp":"2017-09-19T06:48:18.300+00:00","@version":1,"message":"Idling che server scheduled [timeout=3600] seconds]","logger_name":"org.eclipse.che.api.workspace.server.idle.ServerIdleDetector","thread_name":"localhost-startStop
-1","level":"INFO","level_value":20000}                                                                                                                                                                                                   
{"@timestamp":"2017-09-19T06:48:18.232+00:00","@version":1,"message":"List containers registered in the api: []","logger_name":"org.eclipse.che.plugin.docker.machine.cleaner.DockerAbandonedResourcesCleaner","thread_name":"Annotated-sc
heduler-1","level":"INFO","level_value":20000}                                                                                                                                                                                            
{"@timestamp":"2017-09-19T06:48:18.387+00:00","@version":1,"message":"Keycloak is disabled","logger_name":"com.redhat.che.keycloak.server.KeycloakAuthenticationFilter","thread_name":"localhost-startStop-1","level":"INFO","level_value"
:20000}
```

In this PR, the major problem was that we have to use jackson and logback dependencies in the tomcat root classloader (not the webapp one) and it causes classloading issues. So some jars had to be moved to the tomcat/lib folder.

This PR uses logstash-logback-encoder. We could have use https://github.com/qos-ch/logback-contrib as well. Pros for logstash one is that it suits well with ELK (then it's a matter of naming). Pros for  Logback JSON extension is that it's hosted by qos-ch github repo. But classloading hell would be the same.

### DIFF
--- a/builds/fabric8-che/assembly/assembly-main/pom.xml
+++ b/builds/fabric8-che/assembly/assembly-main/pom.xml
@@ -24,6 +24,22 @@
     <name>Fabric8 IDE :: Assemblies Tomcat</name>
     <dependencies>
         <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.redhat.che</groupId>
             <artifactId>fabric8-ide-assembly-wsagent-server</artifactId>
             <type>tar.gz</type>
@@ -41,6 +57,14 @@
         <dependency>
             <groupId>com.redhat.che</groupId>
             <artifactId>user-auth-valve</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>net.logstash.logback</groupId>
+            <artifactId>logstash-logback-encoder</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
         </dependency>
     </dependencies>
     <build>
--- a/builds/fabric8-che/assembly/assembly-main/src/assembly/base-component.xml
+++ b/builds/fabric8-che/assembly/assembly-main/src/assembly/base-component.xml
@@ -31,6 +31,19 @@
                 <include>com.redhat.che:fabric8-ide-assembly-wsagent-server</include>
             </includes>
         </dependencySet>
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <unpack>false</unpack>
+            <outputDirectory>tomcat/lib</outputDirectory>
+            <includes>
+                <include>net.logstash.logback:logstash-logback-encoder</include>
+                <include>ch.qos.logback:logback-classic</include>
+                <include>org.slf4j:slf4j-api</include>
+                <include>com.fasterxml.jackson.core:jackson-annotations</include>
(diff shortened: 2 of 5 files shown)


benoitf:
@sunix 
about

> In this PR, the major problem was that we have to use jackson and logback dependencies in the tomcat root classloader (not the webapp one) and it causes classloading issues. So some jars had to be moved to the tomcat/lib folder.

do you have tried to set
<Loader delegate="true"/>  in context.xml of the webapp ?

sunix:
@benoitf no I haven't try. Should we try this ?

benoitf:
@sunix it may reduce the number of "hacks" around excluding a lot of dependencies from WEB-INF/lib as I suppose it may be hard to maintain

sunix:
Ok I will give a try

l0rd:
@sunix would it be possible to configure the encoder in `che.properties` and move this PR upstream or is it a crazy idea? I ask that because the more we put in rh-che the more painful it will be to maintain it. And it makes sense to have a property like this upstream because Che users may be interested about json format logs.

sunix:
@l0rd I'm not sure how easy it will be (will need to enable/disable the appender in logback.xml with a property value), but in any case we would need to open a IP review request for the new dependencies. I can give a try anyway.

skabashnyuk:
I would say that without logstash or ELK , logs in json format is not that interesting in comparison of class loading issues.  However, logstash or ELK or kibana can be potentially interesting for Eclipse Che multiuser aka onprem in future.

l0rd:
@sunix thanks. We are not in a hurry with this issue so it would be cool if you could investigate on how we could port it to upstream.

l0rd:
@sunix can we close this PR now that we have #311 ?

sunix:
yes closing it

PR #342, opened 2017-09-19
openshift_deploy.sh improvements: skip tls on minishift and MINISHIFT_IP env variable

sunix:
Few enhancement to openshift_deploy.sh script: 
  1. definined a ${MINISHIFT_IP} to be use when using openshift_deploy against a remote minishift
  2. enabling insecure skip tls with minishift but not with the others

### DIFF
--- a/dev-scripts/openshift_deploy.sh
+++ b/dev-scripts/openshift_deploy.sh
@@ -115,6 +115,8 @@ if [ "${OPENSHIFT_FLAVOR}" == "minishift" ]; then
   CHE_KEYCLOAK_DISABLED=${CHE_KEYCLOAK_DISABLED:-${DEFAULT_CHE_KEYCLOAK_DISABLED}}
   DEFAULT_CHE_DEBUGGING_ENABLED="true"
   CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
+  DEFAULT_OC_SKIP_TLS="true"
+  OC_SKIP_TLS=${OC_SKIP_TLS:-${DEFAULT_OC_SKIP_TLS}}
   DEFAULT_CHE_APPLY_RESOURCE_QUOTAS="false"
   CHE_APPLY_RESOURCE_QUOTAS=${CHE_APPLY_RESOURCE_QUOTAS:-${DEFAULT_CHE_APPLY_RESOURCE_QUOTAS}}
 
@@ -134,6 +136,8 @@ elif [ "${OPENSHIFT_FLAVOR}" == "osio" ]; then
   CHE_KEYCLOAK_DISABLED=${CHE_KEYCLOAK_DISABLED:-${DEFAULT_CHE_KEYCLOAK_DISABLED}}
   DEFAULT_CHE_DEBUGGING_ENABLED="false"
   CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
+  DEFAULT_OC_SKIP_TLS="false"
+  OC_SKIP_TLS=${OC_SKIP_TLS:-${DEFAULT_OC_SKIP_TLS}}
 
 elif [ "${OPENSHIFT_FLAVOR}" == "ocp" ]; then
   # ----------------------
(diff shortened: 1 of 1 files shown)


benoitf:
@sunix please review the changes done on the script at https://www.shellcheck.net/

sunix:
@benoitf done

dev-scripts/openshift_deploy.sh
```diff
@@ -91,27 +91,32 @@ OPENSHIFT_FLAVOR=${OPENSHIFT_FLAVOR:-${DEFAULT_OPENSHIFT_FLAVOR}}
...
-  # ---------------------------
-  # Set minishift configuration
-  # ---------------------------
-  echo -n "[CHE] Checking if minishift is running..."
-  minishift status | grep -q "Running" ||(echo "Minishift is not running. Aborting"; exit 1)
-  echo "done!"  
-
-  DEFAULT_OPENSHIFT_ENDPOINT="https://$(minishift ip):8443/"
+  if [ -z "${MINISHIFT_IP}" ]; then
+    # ---------------------------
```
l0rd: I would prefer if we could avoid nested `if`. And it looks like if env variable `${MINISHIFT_IP}` is set we are not going to verify if minishift is running anymore. 

Why don't you just check if minishift is running and then:
```
MINISHIFT_IP=${MINISHIFT_IP:-$(minishift ip)}
```

dev-scripts/openshift_deploy.sh
```diff
@@ -131,6 +136,8 @@ elif [ "${OPENSHIFT_FLAVOR}" == "osio" ]; then
   CHE_KEYCLOAK_DISABLED=${CHE_KEYCLOAK_DISABLED:-${DEFAULT_CHE_KEYCLOAK_DISABLED}}
   DEFAULT_CHE_DEBUGGING_ENABLED="false"
   CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
+  DEFAULT_OPENSHIFT_SKIP_TLS="false"
+  OPENSHIFT_SKIP_TLS=${OPENSHIFT_SKIP_TLS:-${DEFAULT_OPENSHIFT_SKIP_TLS}}
```
l0rd: That's more about OC than OPENSHIFT here. So I would rather call the variable `OC_SKIP_TLS`.

dev-scripts/openshift_deploy.sh
sunix: Yes, if you use it 'remotely', you may not have `minishift` command available. So the status check as it is written won't work.

dev-scripts/openshift_deploy.sh
sunix: @l0rd  Fixed

dev-scripts/openshift_deploy.sh
l0rd: Ok I'm trying to understand here because you usually don't use minishift remotely. You want to use minishift from within a container (e.g. from Che) right? In this case why don't you just use ocp flavor? I'm saying that because it's confusing: if the flavor minishift is select I expect the command minishift to be available. That may be possible if you use the changes proposed [here](https://github.com/redhat-developer/rh-che/pull/340#pullrequestreview-63538831)

sunix: The reason is that I would like to reuse the exact same configuration ( 
  OPENSHIFT_ENDPOINT, OPENSHIFT_USERNAME, OPENSHIFT_PASSWORD,  OPENSHIFT_NAMESPACE_URL, CHE_KEYCLOAK_DISABLED, etc ...) wheras with `ocp` flavor, I would have to reconfigure all of it manually as far as I understood.

riuvshin:
@sunix is that skip TLS applicable for ocp flavor?

l0rd:
@riuvshin actually SKIP_TLS is not for Che but for the tool `oc`. I thought the same thing when I first saw the PR.

ibuziuk:
@sunix are you going to open PR for upstream ?

sunix:
@ibuziuk not sure, @l0rd should I ?

l0rd:
@sunix yes there will be a copy of this file upstream until we get it in the Che CLI so @sunix or @riuvshin (because he made a commit too) can one of you sync latest version of this script upstream too?

davidfestal:
Would it be possible to wait for this PR https://github.com/eclipse/che/pull/6390 to be merged first ?

Issue #343, opened 2017-09-19
Move `builds/fabric8-che` maven project to root folder of rh-che


sunix:
Now that the invoker is not used anymore, it makes sense to move builds/fabric8-che to the root folder
This will envolves:
- updating all the scripts
- update CI (and test results location?)
- notify everybody working on other "big" task: can cause merge conflicts

benoitf:
it would fix https://github.com/redhat-developer/rh-che/issues/328 as well

benoitf:
fixed

PR #352, opened 2017-09-22
Move builds/fabric8-che/* to root directory

benoitf:
It moves the folder of builds/fabric8-che/* to the root directory
 - merge of existing pom.xml and builds/fabric8-che/pom.xml
 - update scripts path

it is link to https://github.com/redhat-developer/rh-che/issues/343

Once it is there we could iterate but I wanted to do the simpler PR to ease the merge.

Change-Id: I4c9804d3c23792fceb5eed5bdc8c16d712b16b56


benoitf:
@sunix @l0rd is it ok for you ? I could merge it on Monday morning

PR #356, opened 2017-09-26
allow to configure che server image pull policy

riuvshin:
for CI / QA use cases I need to be able to set image pull policy to "Always"


riuvshin:
@l0rd @benoitf please review this :)

Issue #379, opened 2017-10-18
Need to migrate to a new github token url


sbose78:
related to https://github.com/fabric8-services/fabric8-auth/issues/149

Related PRs:

- https://github.com/fabric8-services/fabric8-tenant-che/pull/46
- https://github.com/redhat-developer/rh-che/pull/381
- https://github.com/eclipse/che/pull/6809

Related init tenant issue - https://github.com/fabric8-services/fabric8-tenant/issues/414

ibuziuk:
Migration is done - closing

PR #381, opened 2017-10-18
rh-che #379: Migrating to a new GitHub token endpoint

ibuziuk:

### What does this PR do?
 Migrating to a new GitHub token endpoint

https://github.com/redhat-developer/rh-che/issues/379

### DIFF
--- a/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/service/KeycloakTokenProvider.java
+++ b/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/service/KeycloakTokenProvider.java
@@ -10,6 +10,7 @@
  */
 package com.redhat.che.keycloak.token.provider.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.redhat.che.keycloak.token.provider.util.UrlHelper;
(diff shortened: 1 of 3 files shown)


plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/service/KeycloakTokenProvider.java
```diff
@@ -86,18 +82,20 @@ public String obtainGitHubToken(String keycloakToken)
...
     ObjectMapper mapper = new ObjectMapper();
     JsonNode json = mapper.readTree(responseBody);
-    JsonNode token = json.get(ACCESS_TOKEN);
-    if (token != null) {
-      return token.asText();
-    } else {
-      return null;
-    }
+    JsonNode accessToken = json.get(ACCESS_TOKEN);
+    return accessToken.asText();
```
benoitf: It can't have npe if accesstoken is not retrieved ?

ibuziuk: this should never happen I believe

alexeykazakov: If you got 200 response with a JSON then correct. access_token should be always there. If there is no token available (user has not linked GitHub account) then there will be 401 response with a JSON error message and ""WWW-Authenticate" header with details what client should do to solve the problem (it should initiate account linking by calling our GitHub linking API).

ibuziuk: 401 should fail in `getResponseBody` and exception will be thrown

PR #387, opened 2017-10-25
Add centOS dotnet stack

l0rd:
### What does this PR do?
- Ordered stacks as in upstream `stacks.json` to make it easier to compare
- Added a new centos dotnet stack


l0rd:
Upstream related PR https://github.com/eclipse/che/pull/6978

PR #393, opened 2017-10-26
Updated nodejs stack to use nodejs6 in default run command

JPinkney:
### What does this PR do?
Changed the node stack to use rh-nodejs6 instead of rh-nodejs4 in the default run command.

In https://github.com/eclipse/che-dockerfiles/commit/f2133bd0fed61ed47d81b9c66c08d83dcfa79aba#diff-6e0645ff7ade28082a8ffd0a66e62e22 node was changed from rh-nodejs4 to rh-nodejs6 but that wasn't reflected in the stacks.

Running the stack in rh-che in minishift

### DIFF
--- a/assembly/fabric8-stacks/src/main/resources/stacks.json
+++ b/assembly/fabric8-stacks/src/main/resources/stacks.json
@@ -673,7 +673,7 @@
 },
 {
   "name": "CentOS nodejs",
-  "id": "nodejs4",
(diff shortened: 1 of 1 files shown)


JPinkney:
Closing as we have removed the scl enable dependency. New PR for updated stacks.json: https://github.com/redhat-developer/rh-che/pull/411

PR #405, opened 2017-10-31
Revert "openshift.io #1238: Adding keycloak token validation to the '/user' endpoint"

ibuziuk:
This reverts commit 6fdd55a7d4209ef7077a66e2bb32f1c972bf92fe.

should fix regression of https://github.com/openshiftio/openshift.io/issues/1255

### DIFF
--- a/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/contoller/TokenController.java
+++ b/plugins/keycloak-plugin-token-provider/src/main/java/com/redhat/che/keycloak/token/provider/contoller/TokenController.java
@@ -104,7 +104,7 @@ public class TokenController {
     try {
       validator.validate(keycloakToken);
       token = tokenProvider.obtainOsoToken(keycloakToken);
-    } catch (KeycloakException e) {
+    } catch (Exception e) {
       return Response.status(Response.Status.BAD_REQUEST).entity(e.getMessage()).build();
     }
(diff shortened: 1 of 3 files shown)


ibuziuk:
Exception was thrown from `UserAuthValve.java` which expected true / false responses:

```
  /**
   * Verify that a logged in user has access to the current project by making a request against the
   * /api/token/user endpoing on Che server.
   *
   * <p>Base URL for the request is set to the default service name for Eclipse Che ({@code
   * che-host})
   *
   * @param keycloakToken value of the auth header on the request (i.e. the keycloak token).
   * @return true if the user matches the current project owner, false otherwise
   */
  private boolean userMatches(String keycloakToken) {
    URL url;
    HttpURLConnection conn;
    try {
      url = new URL(USER_VALIDATOR_ENDPOINT);
      conn = (HttpURLConnection) url.openConnection();
      conn.setRequestMethod("GET");
      conn.addRequestProperty(AUTHORIZATION_HEADER, keycloakToken);
      try (InputStream is = conn.getInputStream()) {
        BufferedReader in = new BufferedReader(new InputStreamReader(is));
        StringBuilder response = new StringBuilder();
        String line;
        while ((line = in.readLine()) != null) {
          response.append(line);
        }
        return "true".equals(response.toString());
      }
    } catch (IOException e) {
      LOG.fatal("Exception while validating user", e);
    }
    return false;
  }
```

PR #411, opened 2017-11-06
Removed scl enable rh-nodejs6 dependency in node stack

JPinkney:
### What does this PR do?
Removes scl enable rh-nodejs6 dependency in node js stack

scl enable rh-nodejs is no longer needed as of eclipse/che-dockerfiles@8e70296cb5a9bd79add294b3ce1bd55daf414798

Tested manually by using the run command in the stack on a nodejs project

### DIFF
--- a/assembly/fabric8-stacks/src/main/resources/stacks.json
+++ b/assembly/fabric8-stacks/src/main/resources/stacks.json
@@ -673,7 +673,7 @@
 },
 {
   "name": "CentOS nodejs",
-  "id": "nodejs4",
+  "id": "nodejs6",
   "creator": "Dharmit Shah",
   "scope": "general",
     "source": {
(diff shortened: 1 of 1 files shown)


assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -673,7 +673,7 @@
 },
 {
   "name": "CentOS nodejs",
-  "id": "nodejs4",
+  "id": "nodejs6",
```
l0rd: I think that it would be better to rename it "nodejs-centos"

l0rd:
@JPinkney you should open a PR for upstream Che too https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L2243

JPinkney:
@l0rd I have a PR here: https://github.com/eclipse/che/pull/7209/files

PR #415, opened 2017-11-07
Update rh-che with version 5.20

l0rd:


pom.xml
```diff
@@ -199,7 +199,7 @@
             <dependency>
                 <groupId>org.eclipse.che.dashboard</groupId>
                 <artifactId>che-dashboard-war</artifactId>
-                <version>${che.version}</version>
+                <version>5.18.0</version>
```
benoitf: I dont remember from daily call why we should use 5.18.0 UD ?

l0rd: That's because current dashboard v5.19 and higher has it's own keycloak client implementation and it doesn't make sense to patch it for rh-che assembly. That's why I've kept the old version. 

Another solution would have been to simply use upstream UD (without rebuilding it for rh-che anymore). But I am not sure that the upstream UD keycloak component would have worked on OSIO out of the box. @davidfestal what do you think?

davidfestal: I'm not sure at all that this would work out-of-the-box.

l0rd:
I have tested removing dashboard (https://github.com/l0rd/rh-che/commit/d69f79ac94c737a8912829f00cc625b4328d4893) and it doesn't work out of the box: for some reason OpenShift liveness probe fails to check that Che is up and running.

PR #421, opened 2017-11-09
Enable functional tests on pull requests.

rhopp:
Do not merge yet.

PR #423, opened 2017-11-09
Add check to cico_build.sh if current build is from pull request.

rhopp:
In order to have PR check job, we need to check in cico_build.sh for env varialbes telling us that current build is from PR check job. If it is, it should checkout right branch.


rhopp:
Please review ;-)

rhopp:
This PR is required for me to be able to test #421 in ci.centos.

rhopp:
I've added `set +x`. I think it's better to see what's happening instead of doing just `echo`.

PR #425, opened 2017-11-10
Temporary removing PVC due to infrastructure problem

ibuziuk:
Error from server (Forbidden): persistentvolumeclaims "che-data-volume" is forbidden: exceeded quota: persistent-volume, requested: requests.storage=1Gi, used: requests.storage=102376Mi, limited: requests.storage=100Gi

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -26,18 +26,6 @@ objects:
     selector:
       app: che
 - apiVersion: v1
-  kind: PersistentVolumeClaim
-  metadata:
-    labels:
-      app: che
-    name: che-data-volume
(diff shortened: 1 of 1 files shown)


davidfestal:
what about lines https://github.com/redhat-developer/rh-che/blob/32b65d4427cfbe7b0e9470db97b8d69b45c8f7e8/openshift/rh-che.app.yaml#L268 and https://github.com/redhat-developer/rh-che/blob/32b65d4427cfbe7b0e9470db97b8d69b45c8f7e8/openshift/rh-che.app.yaml#L271

?

ibuziuk:
@davidfestal fixed

PR #458, opened 2017-11-23
Install 32bit OpenJDK over 64bit.

jerboaa:
This PR prepares the che-server image to have a 32 bit OpenJDK installed.
In addition to https://github.com/fabric8-services/fabric8-tenant-che/pull/69/ it will
reduce the VmRSS size of the che-server by about 25%

My testing showed that the VmRSS of the JVM levels off at about 240-250MB from
about 335MB of 64bit OpenJDK.

### DIFF
--- a/dockerfiles/che-fabric8/Dockerfile
+++ b/dockerfiles/che-fabric8/Dockerfile
@@ -6,6 +6,11 @@
 
 FROM eclipse/che-server:local
 
+# Install 32bit OpenJDK, remove 64bit version.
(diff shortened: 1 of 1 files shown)


jerboaa:
Here is a chart of the VmRSS sizes for 32 bit vs 64 bit OpenJDK. Since the container limit is well below the max supported memory size of a 32 bit JVM this should be fine.

osio_che_server_32bit_vs_64bit_openjdk

I've noted no regressions when using this for my OSIO account.

ibuziuk:
[test]

PR #460, opened 2017-11-23
Bump to version 5.21.0

l0rd:
### What does this PR do?
Do the necessary modifications to rh-che to make it work with upstream v5.20.1

#453 

Started Che and a vertx workspace on minishift, created a vertx booster from the codebase on osio, and still on osio tested java debugging and bayesian LS with a bad dependency in the pom.xml.


l0rd:
[test]

PR #463, opened 2017-11-30
Added bayesian LS to centos dotnet stack

JPinkney:
### What does this PR do?
Added bayesian LS to centos dotnet stack

None, was just an observation while looking through the stacks

Tested the stack on minishift


msrb:
Bayesian already supports .NET, but only partially, and we don't really advertise it anywhere yet. The main focus is still on Java and NPM.

cc @samuzzal-choudhury @miteshvp

PR #461, opened 2017-11-29
Che6 on OSIO

sleshchenko:
### What does this PR do?
This PR contains changes that are required to be able to run RH Che on OSIO.
It's much easier to review PR commit by commit.
List of changes that are made:
- Added importing dependency managment from che-parent;
- Updated assembly-wsmaster-war;
- Updated assembly-wsagent-war module;
- Removed outdated workspace-logging-override module (che6 doesn't have an API for storing/retrieving machines and agents logs). Here is an issue to provide an ability to override agents logs dir https://github.com/eclipse/che/issues/7807. When this issue will be fixed, I think overriding of logs dirs will be performed by configuration properties;
- Moved WorkspaceSubjectsRegistry from eclipse/che#master for BayesianTokenProvider;
- Updated fabric8-multi-tenant-manager module. Added OpenShiftClientFactory and OpenShiftProjectFactory which will provide different workspace environment configuration according to a current user.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/7155

Depends on ~~https://github.com/eclipse/che/pull/7604~~

### How have you tested this PR?
Yes. Rh Che Server was deployed on https://dev.rdu2c.fabric8.io:8443/, workspaces start successfully
<details>
<summary>Workspace start demo</summary>

rh-che6
</details>

<details>
<summary>Deploying rh-che with PR changes</summary>

It's deployed with [deploy_che.sh](https://github.com/eclipse/che/blob/che6/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh) script from Che repository. It is needed to specify the following configuration parameters for deploying rh-che 
```
# Deploy rh-che based on che6
# https://dev.rdu2c.fabric8.io:8443/
export OPENSHIFT_FLAVOR=ocp

export CHE_IMAGE_REPO="sleshchenko/che-server"
export CHE_IMAGE_TAG="nightly-fabric8"

export CHE_MULTIUSER=true
export CHE_DEDICATED_KEYCLOAK=false

export CHE_OPENSHIFT_PROJECT=che6-server

export OPENSHIFT_ROUTING_SUFFIX=dev.rdu2c.fabric8.io
export OPENSHIFT_TOKEN={TOKEN_HERE}
export OPENSHIFT_ENDPOINT=https://dev.rdu2c.fabric8.io:8443
export ENABLE_SSL=false

export CHE_INFRA_OPENSHIFT_PVC_STRATEGY="common"

export IMAGE_PULL_POLICY=Always
```
</details>

PR #465, opened 2017-12-05
Che6 on OSIO

sleshchenko:
### What does this PR do?
This PR contains changes that are required to be able to run RH Che on OSIO.
It's much easier to review PR commit by commit.
List of changes that are made:
- Changed che-parent version to 6.0.0-M3-SNAPSHOT
- ~~Removed outdated deploy script and yaml files (actual version of deploy script is in eclipse/che repository, yaml files are outdated it could be useful to have actual version here, but IMHO it's better to have only one way of deploying Che Server)~~ They will be updated in another PR
- Fixed Che server base image (che6 doesn't have che-server-multiuser image anymore, so che-server image is the right one)
- Renamed agents to installers in stacks.json (agent field in machine config is renamed to installers in che6 branch)
- Adapted bayesian agent to che6 installer approach (che6 changed new approach for binding agents/installers)
- Fixed usage of revision number while evaluation of image tag (BuildInfo.properties is not packaged into assembly-ide-war anymore, so it is needed to generate revision number inside of rh-che repository)

https://github.com/eclipse/che/issues/7155


sleshchenko:
Guys, my previous PR is a quite big. To make PR review easier I've moved part of changes that are separated to small commits here. Let's try to merge it part by part.
@ibuziuk @l0rd @davidfestal Please review

.ci/cico_do_docker_build_tag_push.sh
```diff
@@ -12,7 +12,7 @@ ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 RH_CHE_TAG=$(git rev-parse --short HEAD)
 
-UPSTREAM_TAG=$(sed -n 's/^revision = \(.\{7\}\).*/\1/p' ${ABSOLUTE_PATH}/../assembly/assembly-build-info/target/dependency/WEB-INF/classes/org/eclipse/che/ide/ext/help/client/BuildInfo.properties)
```
davidfestal: Afaik the information in the build-info was expected to be the one of *upstream*.
It doesn't seem to be the case anymore.

The idea was that the `rh-che` docker image tag would allow knowing the upstream Che commit on which it is based.

sleshchenko: I thought that the main goal is to have different tags for different builds to be able to deploy any of them or roll back in case of an unsuccessful update.

So BuildInfo is not packaged into assembly-ide-war. If it is required to know the upstream Che commit I can create an issue in eclipse/che.

davidfestal: @mario wdyt ?

l0rd:
@sleshchenko the template and configmap yaml files are used to deploy on osio. We cannot remove them. And because of that are also up to date. I agree with you, we need to avoid duplication of code, but we cannot simply remove it from here.

.ci/cico_do_docker_build_tag_push.sh
l0rd: Having the upstream commit is kinda of useful. When running a che server image we can easily associate it to the right upstream and downstream commit. And we can deduce if an upstream fix is included in the image or not. That said that is not a blocker but a nice to have. So yes please @sleshchenko open and issue on upstream Che to discuss that.

sleshchenko: Ok, I'll create an issue for that

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/rh-che.properties
```diff
@@ -1,24 +0,0 @@
...
-#
-
-# This is a routing suffix that can be used when building workspace agents external address
-# with a custom server evaluation strategy. This is similar to the che.docker.ip.external
-# property, but specific to the workspaces. So this is mainly useful when workspaces are not
-# hosted on the same machine as the main Che server and workspace master.
-# This is used as the default value if no other value is provided by the
-# `com.redhat.che.multitenant.Fabric8WorkspaceEnvironmentProvider.getUserCheTenantData()`
-# method. 
-che.fabric8.workspaces.routing_suffix=NULL
```
l0rd: Why this property has disappeared? How are we going to set [the agents routing suffix](https://github.com/redhat-developer/rh-che/blob/multi-tenant-rh-che/openshift/rh-che.config.yaml#L36)?

sleshchenko: It is not actual property for OpenShift SPI 6. There will fetch actual host from created routes. Is it OK or have I missed something?

.ci/cico_do_docker_build_tag_push.sh
sleshchenko: FYI I've discovered that each jar contains `META-INF/MANIFEST.MF` that has the line with revision number like `SCM-Revision: b956dfbacdf17b56c0b5d6b111d6858a9f2dba02`. 
Anyway, I'll create an issue, describe it there and then it will be decided which way is the best to do that.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/rh-che.properties
l0rd: @davidfestal ?

davidfestal: It's used in the server evaluation strategy used on OSIO (AlwaysExternal). And initially, it was also used to build the URLs of the workspace routes. But right, now it isn't anymore. Openshift decides what the workspace route hostname should be. So do you mean that, as a replacement for server evaluation strategy,  che6 retrieves this value directly from the routes ?
This I agree this wouldn't be necessary anymore.

sleshchenko: > Do do you mean that, as a replacement for server evaluation strategy, che6 retrieves this value directly from the routes?
@davidfestal Yes. It's implemented in che6 branch in this way

.ci/cico_build.sh
```diff
@@ -25,7 +25,7 @@ then
   if [ -z "${RHCHEBOT_DOCKER_HUB_PASSWORD+x}" ]; then echo "WARNING: failed to get RHCHEBOT_DOCKER_HUB_PASSWORD from jenkins-env file in centos-ci job."; else export RHCHEBOT_DOCKER_HUB_PASSWORD; fi
   set -x
   yum -y update
-  yum -y install centos-release-scl java-1.8.0-openjdk-devel git patch bzip2 golang docker subversion
```
ibuziuk: @sleshchenko why subversion is now required for the build ?

sleshchenko: because there is not subversion plugin anymore

sleshchenko: Or it is required by some another component?

ibuziuk: @sleshchenko oh, sorry overlook - I thought it was added not removed ;-)

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/rh-che.properties
davidfestal: OK

l0rd: 👍

.ci/cico_do_docker_build_tag_push.sh
davidfestal: OK, thanks

sleshchenko: I've create an issue for that https://github.com/eclipse/che/issues/7854

sleshchenko: @davidfestal @l0rd I've fixed usage of eclipse revision number in the following commit https://github.com/redhat-developer/rh-che/pull/461/commits/ff1385c2ca9bf8e42bfe17ee1608b7cd5e2c1c47

Issue #440, opened 2017-11-16
Port to multi-tenant branch missing commits from rh-che master branch


l0rd:
An example was #433 but there must be other stuff that hasn't been synch with multi-tenant branch.

ibuziuk:
Commit diff between `multi-tenant-rh-che` and `master`

> * d8655d6 - (origin/master, origin/HEAD, master) Bump to version 5.21.0 (11 days ago)
> * 071cdda - Bump to version 5.20.1 (3 weeks ago)
> * 8beb12e - Update to the docker build script (3 weeks ago)
> * 1c334f4 - Sync rh-che with upstream che v5.20 (3 weeks ago)
> * 961798b - Enable functional tests on pull requests. (3 weeks ago)
> * e470151 - Install git and do checkout after that. (4 weeks ago)
> * 4f0549f - Add check to cico_build.sh if current build is from pull request. (4 weeks ago)
> * 398d911 - Changed nodejs6 to nodejs-centos (4 weeks ago)
> * cbcf46b - Removed scl enable rh-nodejs6 dependency in node stack (4 weeks ago)
> * 7bab41c - Update to che v5.18.1 (5 weeks ago)
> * 4e7de79 - (ibuziuk/master, 5.18.1) Revert "openshift.io #1238: Adding keycloak token validation to the '/user' endpoint" (5 weeks ago)
> * 6fdd55a - openshift.io #1238: Adding keycloak token validation to the '/user' endpoint (5 weeks ago)
> * e882a92 - openshift.io #1215: Fixing typo in stacks.json (6 weeks ago)
> * 4c9cf65 - Fix missing comma in stacks.json (6 weeks ago)
> * 9008bbb - Add CentOS .NET 2.0 Stack (6 weeks ago)
> * 0c5380e - Sync stacks.json with upstream version (6 weeks ago)
> * 04230c8 - Fixing local_docker_create_images_and_tag.sh script with right location of cico scripts (6 weeks ago)
> * bfad140 - rh-che #379: Adding missing 'com.fasterxml.jackson.core' dependency (7 weeks ago)
> * 4a87a36 - rh-che #379:  Migrating to a new GitHub token endpoint, but also supporting an old one

ibuziuk:
4 Commits related to updates in `keycloak-plugin-token-provider` are not relevant since this plugin does not exist in `multi-tenant-rh-che` branch:
[...]

PR #469, opened 2017-12-05
Fixing local_docker_create_images_and_tag.sh script with right location of cico scripts

ibuziuk:
### What does this PR do?

Porting [1] from master to multi-tenant branch 

[1] https://github.com/redhat-developer/rh-che/commit/04230c85c9d7020a8671a94005ac46f819d9a157

### Issue:

https://github.com/redhat-developer/rh-che/issues/440#issuecomment-349394313


ibuziuk:
@sunix so you are saying that your PR should not be applied to `multi-tenant-rh-che` ? is it correct ?

sunix:
correct :)

PR #471, opened 2017-12-06
Fix GWT IDE war for rh-che based on che6 branch

sleshchenko:
### What does this PR do?
Fixes GWT IDE war for rh-che based on che6 branch.

Note that ide-gwt-app compiling module has a lot of copied/pasted code from che repository but it will be removed when https://github.com/eclipse/che/issues/7519 will be resolved.
https://github.com/eclipse/che/issues/7155

Yes. Che server was deployed to [OpenShift cluster](https://dev.rdu2c.fabric8.io:8443)


assembly/fabric8-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/IDE.html
```diff
@@ -0,0 +1,313 @@
+<!--
```
benoitf: why do we need the IDE.html from upstream che ?

sleshchenko: yes for now. See PR description

benoitf: I mean, I don't see why we need duplicates. IDE.html is in the upstream artifact so it can be copied as a dependency instead of adding it to source repository

sleshchenko: I tried to do it with maven-war-plugin
```
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-war-plugin</artifactId>
                <configuration>
                    <overlays>
                        <overlay>
                            <artifactId>che-ide-gwt-app</artifactId>
                            <type>war</type>
                            <includes>
                                <include>_app/IDE.html</include>
                            </includes>
                        </overlay>
                        <overlay />
                    </overlays>
                </configuration>
            </plugin>
```
But it doesn't work in this way. I guess it's because of gwt-app packaging, so GWT maven plugin assemblies war but not maven war plugin. So I'd fix it later where there will be a clear way to extend upstream packaging.
Do you know how to easily include IDE.html from upstream?

benoitf: https://github.com/redhat-developer/rh-che/blob/master/assembly/assembly-ide-war/pom.xml#L177-L197 (here it was using ide jsp file)

sleshchenko: It has [war](https://github.com/redhat-developer/rh-che/blob/master/assembly/assembly-ide-war/pom.xml#L23) packaging, I think It'd work even with configuring maven war plugin.

So, I tried the suggested way and it doesn't work for me, maybe I do something wrong.
https://gist.github.com/sleshchenko/d7a5377d6090ea55f47f4aec3b159a92#file-pom-xml-L333-L336
https://gist.github.com/sleshchenko/d7a5377d6090ea55f47f4aec3b159a92#file-pom-xml-L349-L370

Are you OK to merge it with copied/pasted IDE.html and fix it later?

assembly/fabric8-ide-gwt-app/pom.xml
```diff
@@ -0,0 +1,387 @@
...
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+    <artifactId>fabric8-ide-gwt-app</artifactId>
+    <packaging>gwt-app</packaging>
+    <name>Che IDE :: GWT App</name>
+    <description>Compiles Che IDE GWT application and packages it as a WAR-like archive which contains pure output of the GWT compiler</description>
+    <properties>
+        <generated.sources.directory>${project.build.directory}/generated-sources/gen</generated.sources.directory>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
azatsarynnyy: @sleshchenko FYI explicit setting `sourceEncoding` property can be omitted there since the `gwt:enforce-encoding` goal does it during `initialize` build phase within the `gwt-app` maven lifecycle

sleshchenko: thanks, I'll fix it

assembly/fabric8-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/IDE.html
sleshchenko: @benoitf https://github.com/eclipse/che/issues/7519 is merged. So I'll rework it during this PR

sleshchenko:
https://github.com/eclipse/che/issues/7519 is resolved :tada: 
Dear @reviewers, I will notify you when this PR is reworked according to new ide assembly extending approach.

assembly/fabric8-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/IDE.html
benoitf: ok it's better then

sleshchenko: @benoitf Now IDE.html is copied from assembly-ide-war

sleshchenko:
@davidfestal @ibuziuk @l0rd I've rework my PR please review

sleshchenko:
@l0rd I've tested it on https://dev.rdu2c.fabric8.io:8443. More information about deploying configuration and workspace start demo you can find here https://github.com/redhat-developer/rh-che/pull/461
As far as I understand there is no ability to check it on local `OCP` or `minishift` while workspaces started on `console.starter-us-east-2.openshift.com` requires access to workspace master that has no public accessible URL.

PR #476, opened 2017-12-07
Bump to version 5.21.1

l0rd:
Related to eclipse/che#7766 and openshiftio/openshift.io#1304


l0rd:
We should wait the release of v5.21.1 before merging this.

PR #478, opened 2017-12-11
Updated the ping threshold to 25

l0rd:
### What does this PR do?
Update the config to make Che wait for 50 seconds after the first successful ping to the wsagent before continuing with the workspace bootstrap.

https://github.com/openshiftio/openshift.io/issues/1599#issuecomment-350675000
https://github.com/fabric8-services/fabric8-tenant-che/pull/92

I have tried different value (10, 15, 20 and 25) and only with 25 I had 10 successful workspace creations in a raw on osio.


ibuziuk:
Approved and also reopened https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/946
@kbsingh looks like we need to do yet another cm update

l0rd:
ConfigMap has been updated

PR #511, opened 2018-01-15
Added dummy jenkinsfile.

rhopp:
This PR adds dummy Jenkinsfile. 
Once this is merged, CD team can add rh-che repo into https://jenkins.cd.test.fabric8.io/ and job will be created.

Below is transcript of chat with Rupali from CD team:
> rhopp 15:31 Is there a need to have the jenkinsfile already merged into master, or is it posiible to try it from my fork/branch? I would like to be sure everything is right beofre merging into master 

> rupalib 15:41 so we can add an empty Jenkins file to master and then we can make changes to it in your branch and then make Pull request to test your changes

> rupalib 15:52 
#!/usr/bin/groovy 
node {
   echo 'Hello World'
}

> rupalib15:53 add something like the above and you can commit the Jenkinsfile and let us know I will add the org where your repository and the Job will be created.

> rupalib 15:54 https://github.com/fabric8io/fabric8-pipeline-library/wiki/FAQ

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,4 @@
+#!/usr/bin/groovy
+node {
+   echo 'Hello World'
+}


rhopp:
Thinking about this, it should be doable in ci.centos.org, but it would require some scripting, adding credentials (for accessing the dev cluster/badger automatically from duffy machine). 
I really liked the simplicity of jenkinsfile. Just use io.fabric8 libraries and call deployOpenShiftNode: https://github.com/fabric8-ui/fabric8-ui/blob/master/Jenkinsfile#L58

rhopp:
And as a response to your question:
> what is the goal here? To move the build of Che from centos CI to fabric8 CI?

After reading mail thread on devtools-saas ("Re: Help us unblock Planner tests") and quick chat with jaseemabid I'm not sure this is the right path for us (or anybody in osio).

image

PR #512, opened 2018-01-16
Updated year to 2018 in files license headers

sleshchenko:
### What does this PR do?
Update year to 2018 in files license headers.

It also contains revert of commenting of docker pull command.

-

no needed


sleshchenko:
@l0rd I want but I'd say that I can't. Because license year is defined in che-parent, `multi-tenant-rhche` branch depends on che-parent#5.22.0 tag and [2017](https://github.com/eclipse/che-parent/blob/5.22.0/pom.xml#L73) year is defined there.
Even 5.22.1-SNAPSHOT has [2017](https://github.com/eclipse/che-parent/blob/5.22.x/pom.xml#L73) year in license headers.
@skabashnyuk Is there plan to update license headers in 5.x Che?

BTW I think it will not produce a big problem while merging.

l0rd:
@sleshchenko makes sense

Issue #532, opened 2018-02-02
Change the token used to access Che tenant namespace in OSIO


l0rd:
We currently use users' tokens to create Che workspaces in Che tenant namespace. We need to change that flow to:

At startup:
- [x] Use an openshift secret to login to auth service - PR is sent https://github.com/redhat-developer/rh-che/pull/538
- [x] Get che token from auth service - PR is sent https://github.com/redhat-developer/rh-che/pull/538

When creating/stopping/idling a workspace
- [x] Get current user ID from user jwt token - PR is sent https://github.com/redhat-developer/rh-che/pull/533
- [x] Use che token + user ID (as a query param) for proxy requests (auth will return a Che ServiceAccount token)

And this should be deployed using a canary rollout using unleash. So we will also need to:

- [x] Add unleash support in rh-che - PR is sent https://github.com/redhat-developer/rh-che/pull/537

sbose78:
Support for Che to 'login' and get an OSiO service account token tracked in https://github.com/fabric8-services/fabric8-auth/issues/316

ibuziuk:
Some updates related to:

```
At startup:
- [ ] Use an openshift secret to login to auth service
- [ ] Get che token from auth service
```
Housekeeping issue[1] is closed and now `service.account.id` & `service.account.secret` are added to che secret in prod + preview. What is required from che side is:
- add `service.account.id` / `service.account.secret` as env vars to template[2]
- obtain Che SA token via `POST /api/token` [3] where values provided in env vars are used in request parameters

[1] https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1222#note_264499 
[2] https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.app.yaml#L246
[3] https://fabric8-services.github.io/fabric8-auth/reference.html#_service_account_authentication

sbose78:
Auth service is prod-preview and prod now supports service account login from Che.

alexeykazakov:
Not in prod yet.
We need to redeploy prod to pick up the new configuration.

ibuziuk:
PR for adding  che SA to 'fabric8-tenant-che-mt' - https://github.com/fabric8-services/fabric8-tenant-che/pull/97
PR for adding fabric8-auth endpoint to template -https://github.com/redhat-developer/rh-che/pull/535

@davidfestal am I correct that the migration job does not remove service accounts [1]  ?

[...]

PR #533, opened 2018-02-03
rh-che-532: Adding JWT token parser for obtaining identity id / sub

ibuziuk:

### What does this PR do?
Adding JWT token parser for obtaining identity id / sub

JUnit tests


ibuziuk:
> Afaik, at least for che5, the Keycloak-related filters in upstream Che already extract the subject and userId from the JWT claims and store it as the current userId / subject in a TLS.

good point so, it is not required to do low-level jwt parsing for extracting user-id, but for toggle functionality we would also need `session_state` (@aslakknutsen am I correct here - toggle context would require user-id & session state) and AFAIK it is not part of the subject

ibuziuk:
closing since session-state is optional parameter for toggle

PR #536, opened 2018-02-07
rh-che #532: Adding service.account.id / service.account.secret as env vars to openshift template

ibuziuk:
### What does this PR do?
Adding service.account.id / service.account.secret as env vars to openshift template

https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1222#note_264499
#532

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -263,6 +263,16 @@ objects:
               secretKeyRef:
                   key: che.jdbc.username
                   name: che
+          - name: CHE_OPENSHIFT_SERVICE__ACCOUNT_SECRET
+            valueFrom:
+              secretKeyRef:
+                  key: service.account.secret
+                  name: che
+          - name: CHE_OPENSHIFT_SERVICE__ACCOUNT_ID
+            valueFrom:
+              secretKeyRef:
+                  key: service.account.id
+                  name: che
           - name: CHE_WORKSPACE_AGENT_DEV_PING__SUCCESS__THRESHOLD
             valueFrom:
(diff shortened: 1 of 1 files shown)


openshift/rh-che.app.yaml
```diff
@@ -263,6 +263,16 @@ objects:
               secretKeyRef:
                   key: che.jdbc.username
                   name: che
+          - name: CHE_AUTH_SERVICE_ACCOUNT_SECRET
```
garagatyi: I believe that these names don't comply with Che env vars naming policy. Consider 
CHE_OPENSHIFT_SERVICE__ACCOUNT__ID
CHE_OPENSHIFT_SERVICE__ACCOUNT__SECRET

garagatyi: Or 
CHE_OPENSHIFT_SERVICE__ACCOUNT_ID
CHE_OPENSHIFT_SERVICE__ACCOUNT_SECRET

ibuziuk:
@garagatyi thanks for the pointer, PR is updated - could you point me to the che env var naming convention ?

garagatyi:
@ibuziuk I don't think that Che has docs about that, at least I'm not aware of it. But it is empirical knowledge based on tendencies in Che codebase:
- dots in properties separate components (in case of env vars single underscores)
- underscores in properties separate words in a single component term (in case of env vars double underscores)
E.g. 
`che.component1.sub_component_2_mb` -> `CHE_COMPONENT1_SUB__COMPONENT__2__MB`
where MB is unit of measurement in case when property is a number

PR #537, opened 2018-02-07
rh-che #532: Adding toggle for Che SA token switch

ibuziuk:
### What does this PR do?
Adding toggle for Che SA token switch


garagatyi:
I removed myself from reviewers since I can't validate correctness or decisions in this PR. I'm not familiar with components used here yet.

PR #538, opened 2018-02-07
[WIP] RH-CHE-532: add retrieving of service account token

garagatyi:
### What does this PR do?
Retrieves service account token from Auth service to allow communication between Che and OS API using new auth system.
@ibuziuk @davidfestal I'm not sure that I got what has to be done right, so, please, validate this solution. 
BTW I haven't tested this PR because I'm not sure how to do that in dev environment. And I think that because of recent addition of env vars that depends on presence of secrets I'm not sure rhche on minishift works

#532


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/rh-che.properties
```diff
@@ -22,3 +22,8 @@ che.fabric8.workspaces.routing_suffix=NULL
 che.fabric8.multitenant=false
 che.fabric8.user_service.endpoint=https://api.openshift.io/api/user/services
 
+# Enable or disable authentication using service account token retrieved using secrets
+che.openshift.service_account.auth_enable=false
```
ibuziuk: @garagatyi maybe it would be more handy not having `che.openshift.service_account.auth_enable` property (will need to open a new housekeeping issue for adding it to prod / prod-preview cm and update rh-che openshift templete) but rather have a fail-safe impl. of obtaining the token when it would be blank if the exception occurs ?

garagatyi: Sorry, I didn't get last part of the suggested idea. On which values such an implementation would rely on?

ibuziuk: Well, I just do not understand the need of extra property, could we just do return [1] if other properties e.g. `che.openshift.service_account.id` etc are blank ?

[1] https://github.com/redhat-developer/rh-che/pull/538/files#diff-f2821ac62d8158e3bafa443f0876cddeR104

garagatyi: Oh, I see. Yeah, we can do that. But I still need to check it on minishift to figure out what is the behavior when secret is missing in the namespace

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8WorkspaceEnvironmentProvider.java
```diff
@@ -85,6 +93,40 @@ public Fabric8WorkspaceEnvironmentProvider(
             .build(CacheLoader.from(this::loadUserCheTenantData));
   }
 
+  @Inject
+  private void setServiceAccountToken(
+      @Named("che.openshift.service_account.auth_enable") boolean enableServiceAccountTokenAuth,
+      @Nullable @Named("che.openshift.service_account.secret") String serviceAccSecret,
+      @Nullable @Named("che.openshift.service_account.id") String serviceAccId,
+      @Nullable @Named("che.fabric8.auth.endpoint") String authEndpoint) {
```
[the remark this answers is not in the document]
garagatyi: tnx

garagatyi:
@ibuziuk not yet, unfortunately

PR #553, opened 2018-02-28
Update yaml files for che6

davidfestal:
Update yaml files for running OSIO on Che 6


openshift/rh-che.app.yaml
```diff
@@ -74,40 +74,15 @@ objects:
...
-            valueFrom:
-              configMapKeyRef:
-                key: workspace-storage-create-folders
-                name: che
           - name: CHE_LOCAL_CONF_DIR
             valueFrom:
               configMapKeyRef:
                 key: local-conf-dir
                 name: che
-          - name: CHE_OPENSHIFT_PROJECT
```
garagatyi: I think we can inline value instead of setting it in config map since we won't use any other value except `openshift`

garagatyi: sorry, wrong line. I was thinking about CHE_INFRASTRUCTURE_ACTIVE

davidfestal: yes, I thought about that also. @l0rd @ibuziuk wdyt ? should we still use a config map value for this ?

ibuziuk: +1 to inline CHE_INFRASTRUCTURE_ACTIVE - less config we have the better

davidfestal: done in commit https://github.com/redhat-developer/rh-che/pull/553/commits/cd863556f3729f1c0f150130a7cf23407f114aa8

PR #554, opened 2018-03-02
Changing resources names / labels to rhche for che6 deployment on prod / prod-preview osd

ibuziuk:

### What does this PR do?
Changing resources names / labels to rhche for che6 deployment on prod / prod-preview osd

https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1351

Not tested

NOTE: `service.account.secret / service.account.id` were not updated to `rhche` deliberately since they would be missing from the begging. New housekeeping issue should be created once `rhche` secret would be available for porting those values from `che` secret

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -2,21 +2,21 @@ kind: Template
 apiVersion: v1
 metadata:
   labels:
-    app: che
-  name: che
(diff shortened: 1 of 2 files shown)


garagatyi:
@davidfestal I'm not sure we need this volume at all.

PR #555, opened 2018-03-02
rh-che #541: Login to user project using oc CLI in workspace containers

### LINKED ISSUE #541: Inject user OSO token in runtime machine
In an OSIO Che workspace, if a user want to have access to the OSO OpenShift API, he needs to `oc login`. That should be automated: we should inject the OSO token of current user as soon as the workspace container is created. And we should set the current project the user namespace.

For instance, with the vertx quickstart, if a user run `mvn clean fabric8:deploy -Popenshift` from the Che terminal it should build successfully and deploy the quickstart application in the user namespace (not in the *-che namespace).

Note that the token is stored in stored in  `~/.kube/config`

garagatyi:
### What does this PR do?
Injects environment variable to log in to a project on OSO where a user has edit rights. 

Fixes #541


openshift/rh-che.config.yaml
```diff
@@ -20,6 +20,7 @@ data:
   workspaces-memory-limit: "2400"
   workspaces-memory-limit-max: "2400mb"
   enable-workspaces-autostart: "false"
+  keycloak-oso-endpoint: "https://sso.prod-preview.openshift.io/auth/realms/fabric8/broker/openshift-v3/token"
```
l0rd: Can't you just reuse `che-keycloak-auth-server-url` instead of having a new config map?

davidfestal: 2 points here:
- the `che-keycloak-auth-server-url` will disappear with the switch to `fabric8_auth`
- It seems that `fabric8_auth` doesn't provide an endpoint to get the OSO token, as it does for GitHub

So the current way implemented in this PR works *for now*, but we will probably have to ask the Auth team about the best way to get the OSO token for such a need (which from their POV should not exist anymore because of the OSO proxy)

garagatyi: @l0rd I can replace it, but I used the approach OSIO-che team implemented previously. I thought you evaluated it as a better approach. Do you this it is better to change the approach?

@davidfestal there is an endpoint for that I believe. Here are swagger of it http://swagger.goa.design/?url=github.com%2Ffabric8-services%2Ffabric8-auth%2Fdesign#!/token/token_Retrieve
The idea is to call something like `GET https://auth.prod-preview.openshift.io/api/token?for=https://api.starter-us-east-2.openshift.com` where `https://api.starter-us-east-2.openshift.com` should be retrieved from cluster field of UserCheTenantData.
One thing to mention is that Aslak said to me that we should not call this API. I don't know the reason but will write an email to figure it out.

alexeykazakov: You can’t use sso.openshift.io to get OSO tokens anymore.
It’s depricated. It doesn’t support multi cluster and it doesn’t have tokens for users provisioned since January.
You have to use auth.openshift.io to get the tokens.

alexeykazakov: To be clear. We still support login via sso (but working on moving it to Auth) but token management has already moved to auth completely.

garagatyi: @davidfestal @l0rd @alexeykazakov I reworked PR to get OSO token from Auth service.

garagatyi:
I reworked PR to get OSO token from Auth service. Please, review my changes.

garagatyi:
I updated PR to fetch the token from Auth service. I checked it - it works.

garagatyi:
@l0rd @ibuziuk whom can I ping to get my user to the whitelist of CI?

rhopp:
@garagatyi Jus send PR to this repo adding your gh username to admin list: https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L12

garagatyi:
@rhopp thank you, Radim!

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/OpenshiftUserTokenProvider.java
```diff
@@ -94,18 +95,30 @@ private String getUserDescription(Subject subject) {
     return subject.getUserName() + "(" + subject.getUserId() + ")";
   }
 
-  @Nullable
   private String getOsToken(final String keycloakToken) throws RuntimeException {
+    Request request =
+        new Request.Builder().url(tokenEndpoint).get()
+            .header("Authorization", "Bearer " + keycloakToken).build();
     try {
-      return httpJsonRequestFactory
```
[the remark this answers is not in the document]
garagatyi: Because response contains custom content type header which is not supported by this class

PR #559, opened 2018-03-05
Updating scripts for rh-che6 deployment on osd

ibuziuk:
### What does this PR do?
- sync cico scripts with master
- changing image from `che-server-multiuser` to`rh-che-server`

https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1351


ibuziuk:
Looks like approach with copping ci files from master was wrong since build / images were changed in che6

Issue #572, opened 2018-03-06
Simplify deployment of che-server on osio/minishift


ScrewTSW:
**Issue problem:** 
We need a way to make deployment of che-server easier on the user, simplifying/automating the steps.

**Red Hat Che version:** 

version: 5.22.x & 6.1.x

**Runtime:**

runtime used:
   - [X] minishift (include output of `minishift version`)
   - [X] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

l0rd:
@ScrewTSW  closing because it looks like a duplicate of #489 #540. Feel free to re-open

PR #573, opened 2018-03-06
Created scripts for deploying che5 and che6 to openshift

ScrewTSW:
### What does this PR do?
Adds simple, single-command interface for deploying che-servers onto openshift [or compatible]

### Issues
~~'Error: secrets "che" not found' when 'oc apply -f -' on rh-che.app.yaml~~
Workspaces refuse to start: https://pastebin.com/MXefeiya
~~Location of `build-config.yaml` and `image-stream.yaml` moved.~~

Related #572
Closes redhat-developer/che-functional-tests/issues/164

Manual testing against Openshift.io and local minishift instance

### DIFF
--- a/dev-scripts/deploy_custom_rh-che.sh
+++ b/dev-scripts/deploy_custom_rh-che.sh
@@ -1,0 +1,338 @@
+#!/bin/bash
+# Copyright (c) 2018 Red Hat, Inc.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+
+usage="\033[93;1m$(basename "$0") \033[0;1m{-5} [-u <username>] [-p <passwd>] [-o <token>] \033[90m{-t <tag>} {-r <registry>} {-n} {-s} {-h} \033[0m-- Login to dev cluster and deploy che with specific build tag
+
+\033[32;1mwhere:\033[0m
+    \033[1m-5\033[0m  \033[93muse che-server V5 \033[31;1m!!MUST COME AS FIRST FLAG!!\033[0m
+    \033[1m-u\033[0m  \033[93musername for openshift account\033[0m
+    \033[1m-p\033[0m  \033[93mpassword for openshift account\033[0m
+    \033[1m-o\033[0m  \033[93mopenshift token - \033[31;1meither token or username and password must be provided\033[0m
+    \033[1m-h\033[0m  show this help text
+    \033[1m-n\033[0m  do not delete files after script finishes
+    \033[1m-s\033[0m  wipe sql database (postgres)
+    \033[1m-t\033[0m  [\033[1mdefault=latest\033[0m] tag for specific build (first 7 characters of commit hash)
+    \033[1m-r\033[0m  docker image registry from where to pull
+
+\033[32;1mrequirements\033[0m:
+    \033[1moc\033[0m  openshift origin CLI admin (dnf install origin-clients)
+    \033[1mjq\033[0m  json CLI processor (dnf install jq)
+    \033[1myq\033[0m  yaml CLI processor based on jq (pip install yq)"
+
+export RH_CHE_DEPLOY_SCRIPT_CLEANUP="true";
+export RH_CHE_WIPE_SQL="false";
+export RH_CHE_IS_V_FIVE="false";
+export RH_CHE_OPENSHIFT_USE_TOKEN="false";
+export RH_CHE_OPENSHIFT_URL=https://dev.rdu2c.fabric8.io:8443;
+export RH_CHE_JDBC_USERNAME=pgche;
+export RH_CHE_JDBC_PASSWORD=pgchepassword;
+export RH_CHE_JDBC_URL=jdbc:postgresql://postgres:5432/dbche;
+
+function setVars() {
+  if [ "$RH_CHE_IS_V_FIVE" == "true" ]; then
+    export RH_CHE_PROJECT_ID="c82f44b-fabric8-9cc154c";
+    export RH_CHE_DOCKER_IMAGE_REGISTRY="rhche/che-server-multiuser";
+    export RH_CHE_PROJECT_NAMESPACE=che5-automated;
+    export RH_CHE_GITHUB_BRANCH=master;
+  else
+#    export RH_CHE_PROJECT_ID="nightly-fabric8";
+#    export RH_CHE_DOCKER_IMAGE_REGISTRY="dfestal/che-server";
+    export RH_CHE_PROJECT_ID="67da9ee-fabric8-ce16ce1";
+    export RH_CHE_DOCKER_IMAGE_REGISTRY="rhche/rh-che-server";
+    export RH_CHE_PROJECT_NAMESPACE=che6-automated;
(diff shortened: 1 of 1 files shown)


dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,306 @@
...
+oc annotate --overwrite=true route/che kubernetes.io/tls-acme=true
+echo -e "\e[92;1mSUCCESS: Che deployed on \e[34mhttps://che-$RH_CHE_PROJECT_NAMESPACE.dev.rdu2c.fabric8.io/\e[0m"
+
+# CLEANUP
+if [ "$RH_CHE_DEPLOY_SCRIPT_CLEANUP" = true ]; then
+  echo -e "\e[1mCleaning up.\e[0m"
+  clearEnv
+fi
+unsetVars
+echo -e "\e[92;1mDone.\e[0m"
```
ibuziuk: please, add new line in the end

ScrewTSW: done.

l0rd:
@ScrewTSW please have a look at [this gist](https://gist.github.com/l0rd/01b741a92325e398ad3596584e583fcc) where I pasted the output of [shellcheck](https://github.com/koalaman/shellcheck). I may want to address a few problems.

ScrewTSW:
`/bin/bash: .ci/cico_run_functional_tests.sh: No such file or directory` lol... what is this jenkins job?

ScrewTSW:
@l0rd Will update the script :)

ibuziuk:
this is PR check job, which is broken and planned to be fixed with your solution of deployment che on dev cluster on each PR ;-)

ScrewTSW:
@ibuziuk oh boy... but no pressure, right? :D

ibuziuk:
not at all ;-)

ScrewTSW:
@l0rd ok, issues reported by `shellcheck` were fixed except for the declaration ones, tested the script, works as intended

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,304 @@
+#!/bin/bash
+# Copyright (c) 2018 Red Hat, Inc.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+
+usage="\e[93;1m$(basename "$0") \e[0;1m{-5} [-u <username>] [-p <passwd>] \e[90m{-t <tag>} {-r <registry>} {-n} {-s} {-h} \e[0m-- Login to dev cluster and deploy che with specific build tag
```
l0rd: You should replace `\e` with `\033` c.f. [this stackoverflow comment](https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux#comment24883926_5947802)

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,304 @@
...
+    export RH_CHE_PROJECT_ID="67da9ee-fabric8-ce16ce1";
+    export RH_CHE_DOCKER_IMAGE_REGISTRY="rhche/rh-che-server";
+    export RH_CHE_PROJECT_NAMESPACE=che6-automated;
+    export RH_CHE_GITHUB_BRANCH=rh-che6;
+    export ECLIPSE_CHE_GITHUB_BRANCH=6.2.x;
+  fi
+}
+
+function unsetVars() {
+  unset RH_CHE_SERVER_USERNAME;
```
l0rd: I think the name of this variable is misleading. If I understand correctly this is **not** a Che server username but rather the OpenShift cluster username. And anyway you should provide the option to only provide a token instead of username/password.

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,304 @@
...
+  checkCheStatus
+  echo -e "\e[0;38;5;60mChe-server status: Available:\e[0;1m$RH_CHE_STATUS_AVAILABLE \e[0;38;5;60mProgressing:\e[0;1m$RH_CHE_STATUS_PROGRESS \e[0;38;5;60mTimeout:\e[0;1m$CHE_STARTUP_TIMEOUT\e[0m"
+  CHE_STARTUP_TIMEOUT=$((CHE_STARTUP_TIMEOUT-1))
+done
+if [ $CHE_STARTUP_TIMEOUT == 0 ]; then
+  echo -e "\e[91;1mFailed to start che-server: timed out\e[0m"
+  exit 1
+fi
+
+oc annotate --overwrite=true route/che kubernetes.io/tls-acme=true
```
l0rd: What do you need this annotation for?

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,304 @@
...
+
+#echo "$CHE_CONFIG_YAML"
+if ! (echo "$CHE_CONFIG_YAML" | oc apply -f - 2>1 > /dev/null); then
+  echo -e "\e[91;1mFailed to apply configmap.\e[0m"
+  exit 1
+fi
+echo -e "\e[92;1mChe config deployed on \e[34m$RH_CHE_PROJECT_ID\e[0m"
+
+# PROCESS CHE APP CONFIG
+sed -i 's/rhche/che/g' ./rh-che.app.yaml
```
l0rd: Why do you need to do this replacement? Isn't this massive replacement of `rhche` error prone?

dev-scripts/deploy_custom_rh-che.sh
ibuziuk: this one is required for enabling valid acme certificate for che server, otherwise ws-agent -> wsmaster communication would fail due to untrusted cert

dev-scripts/deploy_custom_rh-che.sh
ScrewTSW: renamed.
Can you use origin-clients `oc` with a token?

dev-scripts/deploy_custom_rh-che.sh
ScrewTSW: It is only used as a `name:` property in the config files. I can redo it to leave it as rhche, but that would require modifications in other places including the keycloak redirects provided by Ilja because the server url automatically changes from `https://che.whatever...` to `https://rhche.whatever...` and I didn't find a different way to change that.

Maybe there only needs to be a slight modification for that to work? however I'm not sure what I need to change as the origin deployment starts throwing out issues like `app "rhche" not found` if I don't modify the entire config.

I think you are correct on this call, will have to look into it.

dev-scripts/deploy_custom_rh-che.sh
ScrewTSW: done.

ScrewTSW:
@l0rd Fixed some issues, still have some questions for the two issues mentioned.

dev-scripts/deploy_custom_rh-che.sh
ScrewTSW: done and fixed to use default route from config

ScrewTSW:
@l0rd @ibuziuk @davidfestal Fixed all the issues brought up in discussions, please consider giving it a review :)

ScrewTSW:
Welcome :) I'm happy I was able to help and in the end keep the support for `che5`

PR #580, opened 2018-03-07
rh-che #552: set workspace idle timeout to 30 min

### LINKED ISSUE #552: Reduce workspaces idle timeout to 30 min on OSIO
Che workspaces are currently idled after 1h of inactivity of the user. We should reduce it to 30 minutes.

garagatyi:
### What does this PR do?
Set workspace idle timeout to 30 minutes.
Also, allow to configure it from the config map.
BTW I believe that idle timeout doesn't work at all. We need to upgrade to upstream Che 6.2.x where this issue is fixed AFAIK.
Also, removes duplicated declaration of env var in rh-che deployment.

Fixes #552 
Related to https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1458

Yes, and it didn't work because of a bug in upstream Che. Looks like it is fixed in 6.2.0 but it is not released yet. Actually, the release process is in progress. I'll check the PR after 6.2.0 is released.


ibuziuk:
@garagatyi we need to create housekeeping issue with request of adding new entry to cm on prod / prod-preview. Otherwise if PR would be merged deployment on osd would fail

garagatyi:
@ibuziuk thank you for the information!

ibuziuk:
most welcome

garagatyi:
Just figured out that workspace idling behaves in a way I didn't expect:
1. It works, but since bugfix from Che is not in place it shouldn't )
1. It Throws exceptions
1. It doesn't clean workspace's routes and services
1. It stops WS after 1h of inactivity even though I configured 15 minutes in my deployment

garagatyi:
After upgrade to upstream Che 6.2.0 it works in the meaning that workspaces stop after the delay that is set with the config map. 
I'll check whether idling throws exceptions with permissions on dev cluster but I think it is safe to merge this PR now since permissions issue is a separate topic.
@ibuziuk @davidfestal Please, review the PR

PR #601, opened 2018-03-22
rh-che #597: fix Bayesian LS matching files by name instead of path

### LINKED ISSUE #597: Bayesian lsp not working on che6 on osio
**Issue problem:** 
Bayesian ls is not working on che6 on osio.
Bayesian LSP server is enabled in workspace config, but even though I don't get any bayesian warnings in pom.xml.
Doing `ps aux |grep baye` in terminal does not show any bayesian process running.

**Red Hat Che version:** 

version: 6.2.0

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**
* Create new workspace (I used vert.x)
* Import project (I used vertx-http-booster)
* Open pom.xml
* Insert this dependency in dependencies section:
```
<dependency>
    <groupId>ch.qos.logback</groupId>
    <artifactId>logback-core</artifactId>
    <version>1.1.10</version>
</dependency>
```
* Error marker should appear, but it does not.

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [x] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

garagatyi:
### What does this PR do?
Fix matching files in the ws-agent related code for Bayesian LS. It should match path, not name.

Fixes #597


garagatyi:
@benoitf @davidfestal Can you review my PR?

PR #605, opened 2018-03-26
Fix small shell error in CI script

garagatyi:
### What does this PR do?
Fix small shell error in CI script


garagatyi:
@davidfestal can you review the PR?

garagatyi:
Incorrect shell syntax is used. And Shell shows an error because of that. But it is ignored by the script.

.ci/cico_do_build_che.sh
```diff
@@ -16,7 +16,7 @@ mvnche() {
   which scl 2>/dev/null
   if [ $? -eq 0 ]
   then
-    if [ `scl -l 2> /dev/null | grep rh-maven33` != "" ]
+    if [ `scl -l 2> /dev/null | grep -q rh-maven33` ]
```
davidfestal: why not simply fix it with:

```
if [ "$(scl -l 2> /dev/null | grep rh-maven33)" != "" ]
```
?

I'm not a shell expert, but I'm a bit worried about what happens when the content of `scl -l 2> /dev/null | grep -q rh-maven33` would be an empty string (=> when the `scl` result doesn't contain `rh-maven33`) ?

garagatyi: Your suggestion contains incorrect syntax too. So this doesn't solve the problem.
Since test command `[ ... ]` verifies return code, not string, it should work.

garagatyi:
@davidfestal do you have other concerns regarding this PR?

davidfestal:
> Your suggestion contains incorrect syntax too. So this doesn't solve the problem.

Just for my information, what is incorrect syntax in my last proposal ?

> Since test command [ ... ] verifies return code, not string, it should work.

I believe you :-) Let's merge then.

.ci/cico_do_build_che.sh
garagatyi: Sorry, I misread your suggestion. Wrapping result of the execution with quotes would help too. I'll update the PR since this syntax seems more widely understood that grep parameter `-q`

garagatyi:
@davidfestal I updated PR in accordance with your suggestion

Issue #615, opened 2018-03-29
Unable to push to github repo from prod-preview


rhopp:
**Reproduction Steps:**
1) Create workspace with codebase from users github repo (github repo connected to this user in openshift.io) (In my case it was https://github.com/rhopp/vertx-http-booster.git)
2) Make some changes
3) Commit those changes
4) Push those changes - ERROR:
![image](https://user-images.githubusercontent.com/1215011/38089492-dc6ead4e-335f-11e8-80b4-0e8ab75582ad.png)

rhche logs:
```
{"@timestamp":"2018-03-29T12:39:00.871+00:00","@version":1,"message":"Internal Server Error occurred, error time: 2018-03-29 12:39:00","logger_name":"org.eclipse.che.api.core.rest.RuntimeExceptionMapper","thread_name":"http-nio-8080-exec-7","level":"ERROR","level_value":40000,"stack_trace":"java.lang.IllegalArgumentException: Null URI template isn't allowed
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
	at org.everrest.core.impl.uri.UriBuilderImpl.uri(UriBuilderImpl.java:746)
	at javax.ws.rs.core.UriBuilder.fromUri(UriBuilder.java:119)
	at org.eclipse.che.multiuser.keycloak.server.KeycloakServiceClient.getIdentityProviderToken(KeycloakServiceClient.java:126)
	at org.eclipse.che.multiuser.keycloak.server.oauth2.KeycloakOAuthAuthenticationService.token(KeycloakOAuthAuthenticationService.java:89)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.everrest.core.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:140)
	at org.everrest.core.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:60)
	at org.everrest.core.impl.RequestDispatcher.doInvokeResource(RequestDispatcher.java:306)
	at org.everrest.core.impl.RequestDispatcher.invokeSubResourceMethod(RequestDispatcher.java:297)
[...]

PR #618, opened 2018-03-30
Restore the GitHub token injection for Che 6 on OSIO

davidfestal:
### What does this PR do?

Restore the GitHub token injection ability for Che 6 on OSIO, which was disabled and replaced by Keycloak specific linked account management in upstream Che 6 multi-user.

https://github.com/redhat-developer/rh-che/issues/615

Yes, the involved rest endpoints were tested manually on dev-cluster.

### DIFF
--- a/assembly/assembly-wsmaster-war/pom.xml
+++ b/assembly/assembly-wsmaster-war/pom.xml
@@ -76,6 +76,28 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                  <execution>
+                    <id>repack</id>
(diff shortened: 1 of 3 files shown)


assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -101,7 +123,7 @@
                             <type>war</type>
                         </overlay>
                     </overlays>
-                    <packagingExcludes>WEB-INF/lib/che-core-ide-stacks-${che.version}.jar,
```
garagatyi: Why don't we need exclusion of stacks anymore?

davidfestal: Well, it seems it was mentioned twice in the `packagingExcludes` element. I just removed the duplicates.

PR #619, opened 2018-04-03
Feature tag and push cico script

ScrewTSW:
### What does this PR do?
Adds automated cico build and deployment script for PR-checks

Closes redhat-developer/che-functional-tests/issues/165

manual testing against local and remote `docker`, deployed the build on `dev.rdu2c.fabric8.io` , could not test interaction with jenkins yet


.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,94 @@
...
+  docker rmi eclipse/che-server:local
+  docker rmi eclipse/che-server:nightly-centos
+  docker tag eclipse/che-server:nightly rhcheautomation/${RH_CHE_AUTOMATION_BUILD_TAG}
+  docker rmi eclipse/che-server:nightly
+  docker login -u ${RH_CHE_AUTOMATION_DOCKERHUB_USERNAME} -p ${RH_CHE_AUTOMATION_DOCKERHUB_PASSWORD}
+  docker push rhcheautomation/che-server:${RH_CHE_AUTOMATION_BUILD_TAG}
+}
+
+# Getting dependencies ready
+sudo yum install dnf
```
garagatyi: Is it OK to install anything on CI or is this would run in a container? BTW do we have sudo there?

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,94 @@
...
+sudo dnf install --skip-broken --assumeyes \
+            docker \
+            git \
+            make \
+            jq \
+            origin-clients \
+            python3
+systemctl start docker
```
garagatyi: Am I mistaken that we will run docker daemon as a part of a CI job? Who will shut it down after script execution ends?

.ci/cico_build_and_deploy_custom_rh-che.sh
rhopp: cico scripts are run on nodes, which are freshly provisioned for each build in ci.centos.org (the nodes are clean centos7 machines with root access I think). If you are interested in knowing more, here should be all the info: https://wiki.centos.org/QaWiki/CI/GettingStarted#head-6ed39ce58614481f837ba710d97dfa01ef4ba818

garagatyi: I see no usage of sudo here:
https://github.com/kbsingh/centos-ci-scripts/blob/master/create_looped_disk.sh#L15
https://github.com/kbsingh/centos-ci-scripts/blob/master/build_python_script.py#L29
Can you re-check whether it is available or not?

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: I think the nodes are cleanly wiped every time the task is started on them.
Is there some documentation available with details about what is available on the nodes and which services are running?

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: The nodes should be running under root naturally.

.ci/cico_build_and_deploy_custom_rh-che.sh
garagatyi: I don't know whether we have such docs

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,94 @@
...
+
+cat jenkins-env | \
+    grep -E "(KEYCLOAK|BUILD_NUMBER|JOB_NAME|RH_CHE)" | \
+    sed 's/^/export /g' | \
+    sed 's/= /=/g' | \
+    > export_env_variables
+
+source export_env_variables
+
+retrieveActiveToken
```
ibuziuk: @ScrewTSW got a bit lost with `retrieveActiveToken` function.
why do we need OSIO user token aka `ACTIVE_TOKEN` for pushing image to dockerhub ?

ibuziuk:
@ScrewTSW could you please squash two commits into one and with the message from PR description

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,94 @@
...
+set -e
+set +o nounset
+
+/usr/sbin/setenforce 0
+
+function retrieveActiveToken() {
+  CURL_OUTPUT=$(curl -H "Content-Type: application/json" -X POST -d '{"refresh_token":"'$KEYCLOAK_TOKEN'"}' https://auth.prod-preview.openshift.io:8443/api/token/refresh)
+  ACTIVE_TOKEN=$(echo $CURL_OUTPUT | jq --raw-output ".token | .access_token")
+  if [[ -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_USERNAME}" ]]; then
+    empty_credentials="OSIO username is empty, "
```
ibuziuk: `empty_credentials` should be renamed to `error_message` IMO. also consider reworking var not set case as done in other ci script (but it is personal preference) - https://github.com/redhat-developer/rh-che/blob/master/.ci/cico_deploy.sh#L54

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: it is to check that the account is still provisioned. and the token can be used in the deploy script arguments

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: done.

ScrewTSW:
@ibuziuk squashed and reworded

.ci/cico_build_and_deploy_custom_rh-che.sh
ibuziuk: > it is to check that the account is still provisioned
Not sure we need to make this check, but even if we need it why not to do it when executing functional tests ? From one hand I understand that you want to fail-fast, but not provisioned ci account is an edge case (once provisioned ci account is not going to be revoked, so do not think we should check it )

>  and the token can be used in the deploy script arguments
Sorry, I do not understand `why` do we need this token for deployment on dev cluster ? AFAIK, it is only needed for performing tests so, it would be cleaner to obtain it when we need it, and not just in case. Here is how it is done on che-starter side [1] token is obtained and set as env var only when tests are going to be executed. So, for me it sounds like `cico_build_and_deploy_custom_rh-che.sh` is not the right place for obtaining osio token, since it is not needed for build & deploy on dev cluster

[1] https://github.com/redhat-developer/che-starter/blob/master/cico_run_tests.sh#L32

ScrewTSW:
[test]

ScrewTSW:
[test]

rhopp:
[test]

rhopp:
[test]

ScrewTSW:
I removed them after discussion with @rhopp as it's already being done in our tests suite.

Part three of this automation process is running tests against the build which will be done with a very small and simple script, hence the check was removed for now.

ScrewTSW:
Also please don't close yet, I'm fine-tuning the script and resolving issues.

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,133 @@
...
+                                                                -password-stdin 2>1 | {
+    echo "Docker login failed"
+	exit 1
+  }
+fi
+if [ -z ${RH_CHE_AUTOMATION_RDU2C_USERNAME} ] || [ -z ${RH_CHE_AUTOMATION_RDU2C_PASSWORD} ]; then
+  echo "RDU2C credentials not set"
+  CREDS_NOT_SET="true"
+else
+  oc login https://dev.rdu2c.fabric8.io:8443 --insecure-skip-tls-verify=true \
```
ibuziuk: Consider extracting dev cluster url to env var

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
@l0rd @ibuziuk @garagatyi @davidfestal @rhopp could I ask for a final review?

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: done.

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# Retrieve and test credentials
+
+if [ ! -f "./jenkins-env" ]; then
+  echo "CRITICAL ERROR: Jenkins env was not provided by jenkins job"
+  exit 1
+fi
+
+set +x
+cat ./jenkins-env | grep -E "(KEYCLOAK|BUILD_NUMBER|JOB_NAME|RH_CHE)" | sed 's/^/export /g' | sed 's/= /=/g' > ./export_env_variables
```
l0rd: Instead of `cat $file | grep -E $string` you should `grep -E $string $file`

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,135 @@
...
+  export RH_CHE_AUTOMATION_BUILD_TAG=rh-che-automated-build-$(git rev-parse --short HEAD)
+  export CHE_DOCKER_BASE_IMAGE=eclipse/che-server:nightly-centos
+  echo "che-docker-base-image:${CHE_DOCKER_BASE_IMAGE}"
+  docker pull ${CHE_DOCKER_BASE_IMAGE}
+  docker tag ${CHE_DOCKER_BASE_IMAGE} eclipse/che-server:local
+  rm -rf ./dockerfiles/che-fabric8/eclipse-che
+  mkdir ./dockerfiles/che-fabric8/eclipse-che
+  cp -r ./assembly/assembly-main/target/eclipse-che-fabric8-1.0.0-SNAPSHOT/eclipse-che-fabric8-1.0.0-SNAPSHOT/* ./dockerfiles/che-fabric8/eclipse-che/
+  bash ./dockerfiles/che-fabric8/build.sh || exit 1
+  docker rmi eclipse/che-server:local
```
l0rd: Removing the images free some space but may make your next build slower (because you need to download the layers once again). Do we really need to aggressively free space?

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,135 @@
...
+  docker rmi eclipse/che-server:nightly-centos
```
l0rd: Should you use the env variable here (`docker rmi ${CHE_DOCKER_BASE_IMAGE}`)?

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,135 @@
...
+  docker tag eclipse/che-server:nightly rhcheautomation/che-server:${RH_CHE_AUTOMATION_BUILD_TAG}
+  docker rmi eclipse/che-server:nightly
+  docker images
```
l0rd: Is this necessary?

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: `grep -E $string < ./file` ?

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: It's not shared anyways. the duffy nodes get wiped at the end of each jenkins job, and as far as I know there is no central docker for `ci.centos.org`. We have just been discussing this issue.

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: will replace.

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: that was for debugging purposes. I can remove the listing.

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -0,0 +1,135 @@
...
+
+echo "Running ${JOB_NAME} build number #${BUILD_NUMBER}, testing creds:"
+
+CREDS_NOT_SET="false"
+#curl ${ORIGIN_CLIENTS_URL} > origin-clients.rpm
+#yum install --assumeyes ./origin-clients.rpm
+curl "https://mirror.openshift.com/pub/openshift-v3/clients/${OC_VERSION}/linux/oc.tar.gz" | tar xvz -C /usr/local/bin
+if [ -z ${RH_CHE_AUTOMATION_DOCKERHUB_USERNAME} ] || [ -z ${RH_CHE_AUTOMATION_DOCKERHUB_PASSWORD} ]; then
+  echo "Dockerhub credentials not set"
+  CREDS_NOT_SET="true"
```
l0rd: Why don't you just exit here instead of setting `CREDS_NOT_SET=true` and continuing? The same applies for the next `if` blocks. It's better to exit as soon as possible and avoiding the usage of the `CREDS_NOT_SET` env variable.

ScrewTSW: Because I want to know which of the three secure store credentials fail.

ScrewTSW: and echo an error message for each one.

ScrewTSW:
@l0rd thank you, I've already provided feedback to three of the issues raised. I will fix the rest and update the PR.

ScrewTSW:
[test]

.ci/cico_build_and_deploy_custom_rh-che.sh
ScrewTSW: done.

PR #639, opened 2018-04-17
Feature run ee tests custom deployment

ScrewTSW:
### What does this PR do?
Adds functionality to run EE tests against the custom built and deployed rhche-server

redhat-developer/che-functional-tests/issues/213

not tested yet. PR created for testing

### DIFF
--- a/.ci/cico_build_and_deploy_custom_rh-che.sh
+++ b/.ci/cico_build_and_deploy_custom_rh-che.sh
@@ -5,38 +5,33 @@
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
 
+# ERROR CODES:
+# 1 - missing file
+# 2 - missing credentials
+# 3 - OpenShift login failed
+# 4 - command execution failed
+
 set -x
 set -e
 set +o nounset
 
 /usr/sbin/setenforce 0
 
-export CHE_DOCKER_BASE_IMAGE=eclipse/che-server:nightly-centos
-export ORIGIN_CLIENTS_URL=http://mirror.centos.org/centos/7/paas/x86_64/openshift-origin/origin-clients-3.7.1-2.el7.x86_64.rpm
-export OPENSHIFT_RDU2C_URL=https://dev.rdu2c.fabric8.io:8443/
-export OC_VERSION=3.9.19
+export RH_CHE_AUTOMATION_SERVER_DEPLOYMENT_URL=https://rhche-che6-automated.dev.rdu2c.fabric8.io/
+export BASEDIR=$(pwd)
+export DEV_CLUSTER_URL=https://dev.rdu2c.fabric8.io:8443/
+export OC_VERSION=3.7.48
+export TARGET="rh-integration-test"
 
-function tagAndPushDocker() {
-  export RH_CHE_AUTOMATION_BUILD_TAG=rh-che-automated-build-$(git rev-parse --short HEAD)
-  echo "che-docker-base-image:${CHE_DOCKER_BASE_IMAGE}"
-  docker pull ${CHE_DOCKER_BASE_IMAGE}
-  docker tag ${CHE_DOCKER_BASE_IMAGE} eclipse/che-server:local
-  rm -rf ./dockerfiles/che-fabric8/eclipse-che
-  mkdir ./dockerfiles/che-fabric8/eclipse-che
-  cp -r ./assembly/assembly-main/target/eclipse-che-fabric8-1.0.0-SNAPSHOT/eclipse-che-fabric8-1.0.0-SNAPSHOT/* ./dockerfiles/che-fabric8/eclipse-che/
-  bash ./dockerfiles/che-fabric8/build.sh || exit 1
-  docker rmi eclipse/che-server:local
-  docker rmi ${CHE_DOCKER_BASE_IMAGE}
-  docker tag eclipse/che-server:nightly rhcheautomation/che-server:${RH_CHE_AUTOMATION_BUILD_TAG}
-  docker rmi eclipse/che-server:nightly
-  set +x
-  docker login -u ${RH_CHE_AUTOMATION_DOCKERHUB_USERNAME} -p ${RH_CHE_AUTOMATION_DOCKERHUB_PASSWORD}
-  set -x
-  docker push rhcheautomation/che-server:${RH_CHE_AUTOMATION_BUILD_TAG}
+function BuildTagAndPushDocker() {
+  echo "Docker status:"
+  docker images
+  DeveloperBuild="true" .ci/cico_build.sh
+  echo "After build:"
+  docker images
 }
 
 # Retrieve and test credentials
-
 if [ ! -f "./jenkins-env" ]; then
   echo "CRITICAL ERROR: Jenkins env was not provided by jenkins job"
   exit 1
@@ -47,7 +42,7 @@ echo "***IMPORT jenkins-env NAMES DEBUG***"
 cat ./jenkins-env | sed 's;=.*;;' | sort
 echo "***==============================***"
 
-grep -E "(KEYCLOAK|BUILD_NUMBER|JOB_NAME|RH_CHE)" ./jenkins-env | sed 's/^/export /g' | sed 's/= /=/g' > ./export_env_variables
+grep -E "(DEVSHIFT|KEYCLOAK|BUILD_NUMBER|JOB_NAME|RH_CHE)" ./jenkins-env | sed 's/^/export /g' | sed 's/= /=/g' > ./export_env_variables
 if [ ! -f "./export_env_variables" ]; then
   echo "CRITICAL ERROR: sed edit of ./jeninks_env failed"
   exit 1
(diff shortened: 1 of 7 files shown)


l0rd:
@ScrewTSW is the title of the message (`WIP: DO NOT MERGE!`) up to date or is the PR ready to be reviewed?

ScrewTSW:
@l0rd It's up to date, I have just opened this merge request and am testing the behavior.

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
still the same error.

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

Katka92:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[tewst]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
@l0rd @ibuziuk @alexeykazakov @davidfestal @rhopp @Katka92 The PR check seems to be finally functional :)
Ready for review and prepared for all the changes :D

ibuziuk:
@ScrewTSW well done & great job! Just for the record, am I correct that the only test that is currently running is MavenTestCase [1] (execution of a build command in workspace created against vertx stack) ? What I find really cool that each PR job contains screenshots & surefire report [2] about the test execution (would be awesome if eventually CI would post this info back to PR).

[1] https://github.com/redhat-developer/che-functional-tests/blob/master/tests/src/test/java/redhat/che/functional/tests/suites/SmokeTestSuite.java#L24
[2] http://artifacts.ci.centos.org/devtools/che-functional-tests/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/195/

ScrewTSW:
@ibuziuk Yes, currently there is only MavenTestCase running, because GitTestCase will be unstable until some PRs happen at our side.

There are plans to report back into the PR :) I am still trying to figure out how to get that to happen. There are some variables set by webhooks that could be useful and some teams are doing similar stuff to what we want, so it's very likely that it can be done :)

config
```diff
@@ -2,17 +2,21 @@
 
 TARGET=${TARGET:-"centos"}
 
-export NAMESPACE=${DOCKER_NAMESPACE:-"rhchestage"}
+export NAMESPACE="rhchestage"
 
-if [ $TARGET == "rhel" ]; then
+if [ "$TARGET" == "rhel" ] || [ "$TARGET" == "rh-integration-test" ]; then
```
l0rd: Actually I think that the original intent of @riuvshin was to use variable `TARGET` to indicate the Dockerfile to use for the docker build (`rhel` or `centos`). Probably the variable should have been called differently. Anyway here you are using this variable for something different. This is difficult to understand for someone that read this code without other explanations and can easily lead to errors. As I commented above you could use a `PRCheckBuild` variable and verify if it's set to `true` here (and below).

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -5,38 +5,33 @@
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
 
+# ERROR CODES:
+# 1 - missing file
+# 2 - missing credentials
+# 3 - OpenShift login failed
+# 4 - command execution failed
+
```
l0rd: This file is only ment to be used by prcheck job right? Why don't you rename the script so that it is straightforward (something like `cico_rhche_prcheck.sh`)?

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -5,38 +5,33 @@
...
+export RH_CHE_AUTOMATION_SERVER_DEPLOYMENT_URL=https://rhche-che6-automated.dev.rdu2c.fabric8.io/
+export BASEDIR=$(pwd)
+export DEV_CLUSTER_URL=https://dev.rdu2c.fabric8.io:8443/
+export OC_VERSION=3.7.48
+export TARGET="rh-integration-test"
+
+function BuildTagAndPushDocker() {
+  echo "Docker status:"
+  docker images
+  DeveloperBuild="true" .ci/cico_build.sh
```
l0rd: `DeveloperBuild` was ment to set to `true` when running locally on a dev laptop. But here, for PR check purposes you are setting it to true for cento-ci. And for build of the Docker images you set it to false. That is extremely confusing and really easy to break. Can't you just use a new variable `PRCheckBuild` and set it to true here and then checking it in `cico_build.sh`?

.ci/cico_build_and_deploy_custom_rh-che.sh
```diff
@@ -62,33 +57,34 @@ echo "***======================================***"
...
 fi
-if [ -z ${RH_CHE_AUTOMATION_RDU2C_USERNAME} ] || [ -z ${RH_CHE_AUTOMATION_RDU2C_PASSWORD} ]; then
+if [ -z "${RH_CHE_AUTOMATION_RDU2C_USERNAME}" ] || [ -z "${RH_CHE_AUTOMATION_RDU2C_PASSWORD}" ]; then
   echo "RDU2C credentials not set"
   CREDS_NOT_SET="true"
 else
-  oc login ${OPENSHIFT_RDU2C_URL} --insecure-skip-tls-verify \
-                                  -u ${RH_CHE_AUTOMATION_RDU2C_USERNAME} \
-                                  -p ${RH_CHE_AUTOMATION_RDU2C_PASSWORD} && echo "Credentials test OK" || {
+  if oc login ${DEV_CLUSTER_URL} --insecure-skip-tls-verify \
```
l0rd: We should avoid nested `if` as much as possible. And we should fail-fast too. So I would suggest instead of:
```bash
if [ somecondition ]; then
  echo "some error output"
  CREDS_NOT_SET="true"
else
   if [ anothercondition ]; then
       .....
   fi
fi

if [ "${CREDS_NOT_SET}" == "true" ]; then
   exit 2
fi
```

to 

```bash
if [ somecondition ]; then
  echo "some error output"; exit 2
fi

if [ anothercondition ]; then
   .....
fi
```

config
```diff
@@ -2,17 +2,21 @@
...
   export DOCKERFILE="Dockerfile.rhel"
   export REGISTRY=${DOCKER_REGISTRY:-"push.registry.devshift.net/osio-prod"}
 else
   export DOCKERFILE="Dockerfile"
   export REGISTRY=${DOCKER_REGISTRY:-"push.registry.devshift.net"}
 fi
 
-export DOCKER_IMAGE="rh-che-server"
+if [ "$TARGET" == "rh-integration-test" ]; then
+  export DOCKER_IMAGE="rh-che-automation"
```
l0rd: This is not critical but ideally I would love to have something like `rh-che-server:<upstream-tag>PR#<pr-number>` as image name and tag:
- `DOCKER_IMAGE="rh-che-server"` (the result of the build is always a che server)
- `RH_TAG_DIST_SUFFIX="PR#1234"` (when it's a PRCheckBuild)

config
ScrewTSW: That will require complex changes to the ci.centos.org yaml.
There is no logic in place that would count with multiple extra variables set in the templates.
Other than that, I'm using `&job_template_defaults` and not `&job_template_build_defaults` which contains the `$TARGET` logic. I don't know how much could that change of templates influence my pr-check.

ScrewTSW: For example, job_template_build_defaults doesn't contain our logic for artifact-ing the runs.
It doesn't save surefire reports.
I could theoretically circumvent that by adding `TARGET="rhel" PR_CHECK_BUILD="true"` into our `ci_cmd`
https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L2291

rhopp: I would also leave `TARGET=rhel` to indicate, that this is in fact still rhel build of rh-che. And adding another variable indicating, that this is PR check build is fine with me.

config
ScrewTSW: Will add this as an issue for further improvement, but don't you think it's better to keep the PR test builds separate from normal releases?
For example the registry could have a number of builds to hold set, so it auto-cleans. There will be ~~many~~ builds in the future.

ScrewTSW:
[test]

.ci/cico_rhche_prcheck.sh
```diff
@@ -65,18 +66,6 @@ fi
 if [ -z "${RH_CHE_AUTOMATION_RDU2C_USERNAME}" ] || [ -z "${RH_CHE_AUTOMATION_RDU2C_PASSWORD}" ]; then
   echo "RDU2C credentials not set"
   CREDS_NOT_SET="true"
```
l0rd: It's cool that you have moved all precondition checks higher in the script and that you exit before continuing with `oc login`. However why don't you `exit 2` here (as well as at lines 64 and 72) instead of continuing execution? Code would be shorter, you could get rid of a global variable (`CREDS_NOT_SET`) and we would fail even faster.

config
l0rd: Ok to do it in a separate PR. 

Anyway I still believe we should use `DOCKER_IMAGE="rh-che-server"`. 

I would understand if this images would be published to the docker hub where we access through the GUI and can list all the tags of a given image. But do anyone of us really look at the tags of a devshift image? Yes you are right: tomorrow instead of having 100 tags we will probably have 10K tags. But is it really a problem?

.ci/cico_rhche_prcheck.sh
ScrewTSW: Because I want to see which credentials are incorrect. every check has it's own `echo`. This thing has been already discussed in the PR I've opened against Ilja's repo.

ScrewTSW:
[test]

ScrewTSW:
@l0rd Well, as I was discussing it with servicedel, there is a way to automatically clear a docker repo from old tags, however it would require a plugin (it's not a build-in functionality)
I would just let it do it's thing for now and we will see if it becomes a problem later on.

.ci/cico_rhche_prcheck.sh
l0rd: So why don't you leave the specific `echo` for each error but than `exit 2` right away?

ScrewTSW: I'm confused
I am doing the checks in 3 separate if statements
and as soon as they are done with the check they do `exit 2`

Issue #632, opened 2018-04-13
RhChe doesn't work with the latest Eclipse Che 


garagatyi:
**Issue problem:** 
There are some changes in upstream Che that are not compatible with the code in RhChe.
What I found:
- class `OpenShiftServersConverter` was replaced by `org.eclipse.che.workspace.infrastructure.kubernetes.provision.server.ServersConverter`
- Method `getUserId` in `MachineTokenRegistry` was removed.

**Red Hat Che version:** 

version: (help/about menu)

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**

Describe how to reproduce the problem

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [ ] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

PR #646, opened 2018-04-20
Fix for #632 (RhChe doesn't work with the latest Eclipse Che)

davidfestal:
### What does this PR do?

This PR make `rh-che` compatible with the last upstream Che 

#632 

Tested on Dev cluster

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/RhCheInfraEnvironmentProvisioner.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/RhCheInfraEnvironmentProvisioner.java
@@ -33,9 +33,9 @@ import org.eclipse.che.workspace.infrastructure.kubernetes.provision.PodTerminat
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.env.EnvVarsConverter;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.limits.ram.RamLimitProvisioner;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.restartpolicy.RestartPolicyRewriter;
+import org.eclipse.che.workspace.infrastructure.kubernetes.provision.server.ServersConverter;
 import org.eclipse.che.workspace.infrastructure.openshift.OpenShiftEnvironmentProvisioner;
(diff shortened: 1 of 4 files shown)


davidfestal:
This PR should be merged only when we will use the 6.5.0 release

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/RhCheInfraEnvironmentProvisioner.java
```diff
@@ -64,7 +64,7 @@ public RhCheInfraEnvironmentProvisioner(
       @Named("che.infra.kubernetes.pvc.enabled") boolean pvcEnabled,
       OpenShiftUniqueNamesProvisioner uniqueNamesProvisioner,
       RouteTlsProvisioner routeTlsProvisioner,
-      OpenShiftServersConverter openShiftServersConverter,
+      ServersConverter openShiftServersConverter,
```
garagatyi: The usage of this injection should produce a warning. I think we might need to use type on injection of generic.

davidfestal: Fixed in commit https://github.com/redhat-developer/rh-che/pull/646/commits/c0990d663ff5f12400a772a786b209eb4f339a02

pom.xml
```diff
@@ -16,7 +16,7 @@
     <parent>
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
-        <version>6.3.0</version>
+        <version>6.5.0-SNAPSHOT</version>
```
l0rd: I know that I said the opposite a few days ago so please @davidfestal apologies...but thinking more about that what about setting 6.4.1 here and ask Roman to release a 6.4.1 that would include https://github.com/eclipse/che/commit/c66d34972816f2d8684dcca34e4b1592e1f2e95e?

davidfestal: No problem. I'll test it on the 6.4.0 branch, and will ask Roman to create the release when it works correctly.

l0rd: Thank you @davidfestal

davidfestal: done in commit https://github.com/redhat-developer/rh-che/pull/646/commits/3726f5b1ada13e5ad3e5c04e2205973982177879

PR #656, opened 2018-04-26
remove ability to override prod docker registry

riuvshin:
remove ability to override prod docker registry


riuvshin:
already discussed with @l0rd

PR #660, opened 2018-04-30
Fix broken PVC cleaning on workspace removal

davidfestal:
### What does this PR do?

This PR fixes the cleaning of workspace PVCs on workspace removal even, which was broken since the migration to Che 6. 

https://github.com/openshiftio/openshift.io/issues/1614#issuecomment-385423915

Yes, on Dev cluster

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
@@ -11,7 +11,9 @@
 package com.redhat.che.multitenant;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.commons.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


davidfestal:
@ibuziuk sure. It's an __asynchronous operation__, but it is started __right after_ the workspace is deleted, and executed with the interactive user, thanks to thread-local context propagation.

It doesn't take long but still requires starting a POD with a volume mount. And it doesn't seem there a way to be notified of the end of the operation.

One good thing through it that it logs any error during cleaning to the Che server logs.

davidfestal:
> It's an asynchronous operation,

That's the main difference with what cleaning endpoint would provide : the ability to know the result of the cleaning.

ibuziuk:
yeah, sure but from codebase UI perspective it is good to have it async

davidfestal:
perfect then !

ibuziuk:
@l0rd since we would need to redeploy rhche again it would be great to set auto-start to true together with this change and have one redeployment with 2 fixes (just need to get @rhopp +1 for doing this)

davidfestal:
@l0rd @ibuziuk I'd like to merge it as soon as possible to test this again on prod-preview. Are you OK ?

ibuziuk:
+1 ^

Issue #635, opened 2018-04-16
Fix the rh-che build on top of master CI job


l0rd:
- Merge rh-che6 branch to master
- Update master branch CI
- Delete rh-che6 branch CI
- Update the QE team CI (should not be needed but double check)

PR #676, opened 2018-05-14
Merge rh-che6 branch in master

l0rd:
### What does this PR do? 

```bash
git checkout rh-che6
git merge -s ours master
git checkout master
git merge rh-che6
```

#635 

Built locally, deployed to dev cluster, checked that creating a vert.x workspace could be created


l0rd:
>  I assume there were quite a few conflicts.

No there were no conflict (the first merge avoid conflicts).

> git checkout rhche-6
git branch -f master
git push origin master:master -f

If we do this we would loose some commit from master branch. Something we should avoid.

> We could also create rhche-5 branch

That's a good idea but since we don't rewrite git history we could simply create a tag in master. But since we use commit hashes to identify the versions we deploy in production is it really necessary?

l0rd:
Opened a PR to update cico-jobs: https://github.com/openshiftio/openshiftio-cico-jobs/pull/658

ScrewTSW:
Well, it would be good idea to at least have the tag, to be able to identify where the original master branch was.

PR #678, opened 2018-05-22
Tuninng OpenJDK JVM flags for che server

ibuziuk:
### What does this PR do?
Updating `che-server-java-opts` 

For some reason options set in https://github.com/redhat-developer/rh-che/pull/474 are not applied in master / prod & prod-preview osd

Related gitlab issue - https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/945

Tested on minishift

Original PR - https://github.com/fabric8-services/fabric8-tenant-che/commit/eccec4c622d174ceb5600b247021d08a0fb57a83

Note: that I have not updated memory limit for che server and it still uses `750Mi`, potentially it could be reduced to `512Mi` as it was done in the original single-tenant che PR.  Not sure if it worth reducing ram since no proper stress tests have not been performed with lower limits

### DIFF
--- a/openshift/rh-che.config.yaml
+++ b/openshift/rh-che.config.yaml
@@ -21,7 +21,7 @@ data:
   workspaces-memory-limit-max: "2400mb"
   enable-workspaces-autostart: "false"
   keycloak-github-endpoint: "https://auth.prod-preview.openshift.io/api/token?for=https://github.com"
-  che-server-java-opts: "-XX:+UseParallelGC -XX:MinHeapFreeRatio=25 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true -Xms50m -Xmx180m -Dfile.encoding=UTF8"
+  che-server-java-opts: "-XX:MaxRAMFraction=2 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true -Xms20m"
   che-workspaces-java-opts: "-XX:+UseG1GC -XX:+UseStringDeduplication -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:MaxRAM=1200m -Xms256m"
   che-openshift-secure-routes: "true"
   che-secure-external-urls: "true"


ibuziuk:
@garagatyi summary from the original PR:

[che-server] Tune OpenJDK JVM flags so as to use less memory.

> - Set MaxRAMFraction=2 so that MaxHeapSize (or -Xmx)
>   will get set to ~1/2 of the container limit. It's
>   ~1/4 by default.
> - Set -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
>   -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4
>   -XX:AdaptiveSizePolicyWeight=90 *and* -Xms20m (a
>   small value) so as to make the collector work
>   well for small JVMs.
> - Set -XX:+UnlockExperimentalVMOptions
>   -XX:+UseCGroupMemoryLimitForHeap for the JVM to know
>   to base the memory settings on the container limit.
>   It's the better alternative of the old MaxRAM setting.
> - Don't memory-map jar files: -Dsun.zip.disableMemoryMapping=true
>   as this also shaves off some bytes RSS.
> - Reduce the che-server container limit from 600MB to
>   512MB as it runs fine with the lower limit given the
>   above tunings.
>

garagatyi:
@ibuziuk thanks for the explanation

l0rd:
Hey @ibuziuk these were the settings for the 64 bit JDK. But we are currently using the 32 bit JDK and the original PR you mention to was reverted by https://github.com/fabric8-services/fabric8-tenant-che/commit/ef748b6eb39017764fe489de278eb512642c8420 (and in rh-che by 195007157fdacce07f8e7322c562a564edad037f#diff-89824f4a1e5e379f4f8378527ac0aab8)

ibuziuk:
@l0rd hmm... got a bit lost with java options for che-server. Options & memory limits in commits you are referring for tenant-che & rh-che still looks a bit different from what we currently have in master branch & on prod / prod-preview osd. Should we still update it ? Also it is not clear why for 32 bit JDK ram limit was increased - https://github.com/redhat-developer/rh-che/commit/195007157fdacce07f8e7322c562a564edad037f#diff-c98f3dce05b630b31da8dbdfee24da1fR295 (was more memory allocated 'just in case' or deliberately?

ibuziuk:
Just checked history and the difference  in options is about fixing charset encoding `-Dfile.encoding=UTF8` which is also already updated in cms on prod & prod-preview. So, as I understand no changes should be done here and this PR as well as https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/945 should be closed

l0rd:
> Should we still update it?

What problem are we are trying to solve with this PR? I think I am missing this info. From what I know updating wsmaster `java_opts` won't address any sprint backlog issue.

> Why for 32 bit JDK ram limit was increased

I think we increased the resource limit not because of the 32 bit but because we moved to dedicated cluster where we had less aggressive memory limits (note that resource request has remained the same).

ibuziuk:
> What problem are we are trying to solve with this PR? I think I am missing this info.
I got a notice about cm update in gitlab on Friday https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/945#note_332944 which results in this PR

ibuziuk:
Closing this one since PR is not relevant and the server options are correct in repo cm on prod & prod-preview

Issue #630, opened 2018-04-11
Add basic Woopra telemetry events into Che on OSIO


slemeur:
At this moment, we have no Woopra telemetry event reported from Che and have no visibility of when it is used. 

We should be adding a basic Woopra telemetry event so we can know when Che has been used.

l0rd:
It should be just about adding one js line. If it takes more than one day of dev let's skip it for now and tackle it during another sprint.

PR #685, opened 2018-05-26
Segment/Woopra -based telemetry in OSIO Che workspaces

davidfestal:
### What does this PR do?

It adds telemetry in the Che workspaces feeding the Segment and Woopra tools with events and precise durations.

Events types are:
- workspace start
- workspace use (any activity in the IDE or agents apart from editing a file)
- edition of file (when a file is saved): contains a property with the programming language
- workspace inactivity : when the user hasn't been using the workspace for more than 3 minutes but it was still not idled
- workspace stop: when the workspace is stopped by the Che server 

https://github.com/redhat-developer/rh-che/issues/630

Yes, on Dev cluster

The results of my last session pushed into the `test.openshift.io` can be seen [here](https://www.woopra.com/live/project/test.openshift.io/people/all/id/U2yqnIj1mSp7)

### DIFF
--- a/openshift/rh-che.config.yaml
+++ b/openshift/rh-che.config.yaml
@@ -46,3 +46,6 @@ data:
   che-logs-sentry-dsn: 'NULL'
   sentry-stacktrace-app-packages: 'com.redhat,org.eclipse.che'
   sentry-environment: 'dev'
+  analytics.segment_write_key: 'rIZ5NDpUfqkTNdfsBxOBcJoLDT3TxftM' 
+  analytics.woopra_domain: 'test.openshift.io'
+  
--- a/openshift/rh-che.config.yaml
+++ b/openshift/rh-che.config.yaml
@@ -46,6 +46,6 @@ data:
   che-logs-sentry-dsn: 'NULL'
   sentry-stacktrace-app-packages: 'com.redhat,org.eclipse.che'
   sentry-environment: 'dev'
-  analytics.segment_write_key: 'rIZ5NDpUfqkTNdfsBxOBcJoLDT3TxftM' 
-  analytics.woopra_domain: 'test.openshift.io'
+  analytics.segment_write_key: 'NULL' 
+  analytics.woopra_domain: 'NULL'
   
--- a/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
+++ b/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
@@ -24,20 +24,40 @@
     <name>Fabric8 IDE :: Plugins :: Segment Analytics Integration :: WsAgent part</name>
     <dependencies>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
         </dependency>
         <dependency>
             <groupId>com.google.inject.extensions</groupId>
-            <artifactId>guice-multibindings</artifactId>
+            <artifactId>guice-servlet</artifactId>
         </dependency>
         <dependency>
             <groupId>com.segment.analytics.java</groupId>
             <artifactId>analytics</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.segment.analytics.java</groupId>
+            <artifactId>analytics-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-languageserver</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-languageserver-shared</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace-shared</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
@@ -48,16 +68,27 @@
             <artifactId>che-core-workspace-activity-server</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.che.plugin</groupId>
-            <artifactId>che-plugin-json-server</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>javax.servlet</groupId>
             <artifactId>javax.servlet-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
             <scope>provided</scope>
         </dependency>
         <dependency>
@@ -76,20 +107,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>analyze</id>
-                        <configuration>
-                            <skip>true</skip>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
-        </plugins>
-    </build>
 </project>
--- a/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
+++ b/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
@@ -14,7 +14,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <artifactId>fabric8-ide-plugins-parent</artifactId>
+        <artifactId>che-plugin-analytics</artifactId>
         <groupId>com.redhat.che</groupId>
         <version>1.0.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
--- a/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
+++ b/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
@@ -1,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2016-2018 Red Hat, Inc.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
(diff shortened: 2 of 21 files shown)


davidfestal:
FTR I still have to run PR builds / check before merging, but this shouldn't prevent reviews

ScrewTSW:
[test]

openshift/rh-che.config.yaml
```diff
@@ -46,3 +46,6 @@ data:
   che-logs-sentry-dsn: 'NULL'
   sentry-stacktrace-app-packages: 'com.redhat,org.eclipse.che'
   sentry-environment: 'dev'
+  analytics.segment_write_key: 'rIZ5NDpUfqkTNdfsBxOBcJoLDT3TxftM'
```
l0rd: Even if this is not a valid key I think it's better if you set it to the empty string:
```
analytics.segment_write_key: '' 
```

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package com.redhat.che.plugin.analytics.wsagent;
+
+import static com.redhat.che.plugin.analytics.wsagent.EventProperties.*;
+
+public enum AnalyticsEvent {
+  WORKSPACE_STARTED("Start Workspace in Che", 3),
```
l0rd: Since these are events I would rather use the simple past in the event name:
```
s/Start/Started/
s/Use/Used/
s/Keep/Kept/
s/Stop/Stopped/
s/Edit/Edited/
```

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
```diff
@@ -0,0 +1,434 @@
...
+
+    void onActivity() {
+      lastActivityTime = System.currentTimeMillis();
+    }
+
+    void sendPingRequest() {
+      try {
+        URI uri =
+            new URI(
+                "http://www.woopra.com/track/ping?host="
```
l0rd: Please extract constant here and for the parameters

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/OverridenLastAccessTimeFilter.java
```diff
@@ -0,0 +1,67 @@
...
+import org.eclipse.che.workspace.activity.LastAccessTimeFilter;
+import org.eclipse.che.workspace.activity.WorkspaceActivityNotifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Overridden version of the {@link LastAccessTimeFilter} that ignores all accesses to the {@link
+ * AnalyticsActivityService}.
+ *
+ * <p>This important to avoid a REST messaging endless loop between the wsmaster and the wsagent
```
l0rd: This **is** important

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
davidfestal: In fact in Woopra by default the event label is built the following way: `Did <event name>`

This is why I thought that having verbs at the present would be better suited, because it finally renders like this in the UI:

`Did start workspace in che`

If I use the simple past, I get

`Did started workspace in che`

which is quite ugly IMO.

And to change the way event labels are shown in Woopra, you need to change the default schema of each event which requires edit rights I don't have.

In fact here it's more the name / ID of the immediate event than it's label.

plugins/che-plugin-analytics/che-plugin-analytics-wsmaster/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <configuration>
+                            <skip>true</skip>
```
garagatyi: Can you elaborate on why dependency check was disabled?

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
davidfestal: Fixed

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
davidfestal: fixed

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/OverridenLastAccessTimeFilter.java
davidfestal: fixed

openshift/rh-che.config.yaml
davidfestal: fixed

davidfestal:
@l0rd I fixed the various changes you requested, and tested the fixes successfully on dev-cluster with the deployed result of the last CI build. Is it OK for you now ?

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
```diff
@@ -0,0 +1,437 @@
...
+                  "Analytics Network Request Submitter");
+            }
+          });
+
+  private LoadingCache<String, EventDispatcher> dispatchers;
+
+  private String workspaceStartingUserId = null;
+
+  @Inject
+  public AnalyticsManager(
```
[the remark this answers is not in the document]
davidfestal: wdym by a user ? a plugin developer for example ?

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <configuration>
+                            <skip>true</skip>
```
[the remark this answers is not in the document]
davidfestal: Probably because I initially copied the `pomxml` from the bayesian language server `pom.xml` that had dependency analysis disabled ;-). But I can try to remove it and build again sure.

plugins/che-plugin-analytics/che-plugin-analytics-wsmaster/pom.xml
davidfestal: see answer here: https://github.com/redhat-developer/rh-che/pull/685#discussion_r191708384

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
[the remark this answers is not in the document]
garagatyi: No, I mean a user of Che on OSIO can send any activity to Woopra if I'm not mistaken.  Does it have any protection?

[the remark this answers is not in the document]
davidfestal: It seems to me it would be much easier for a user to send wrong data to Segment / Woopra from the  web application console if the integration was done in javascript on the client-side (just call some javascript method in the browser console). But here eveything is done inside Java code running inside the wsagent tomcat process, and using Segment / woopra settings that are finally stored in an OpenShift config map (or secret).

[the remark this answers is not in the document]
garagatyi: Yes, it should be even easier from the client side. So maybe it is not a problem. 
I was asking because I didn't see any injection of any key or so from config map. But after discussion of the JS-based solution, it doesn't seem a problem to me.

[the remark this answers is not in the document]
davidfestal: The keys are retrieved by the wsagent from a wsmaster service endpoint. And the wsmaster gets them finally from a config map.

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
```diff
@@ -0,0 +1,437 @@
...
+  private final long noActivityTimeout = 60000 * 3;
+  private final long pingTimeout = 30 * 1000;
+  private final String workspaceId;
+  private final String workspaceName;
+
+  private String segmentWriteKey;
+  private String woopraDomain;
+
+  private ScheduledExecutorService checkActivityExecutor =
+      Executors.newSingleThreadScheduledExecutor(
```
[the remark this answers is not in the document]
davidfestal: fixed

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/pom.xml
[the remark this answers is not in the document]
davidfestal: fixed

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsActivityService.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+import org.eclipse.che.api.core.rest.Service;
+
+@Singleton
+@Path(AnalyticsActivityService.PATH)
+public class AnalyticsActivityService extends Service {
```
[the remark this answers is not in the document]
davidfestal: Added javadoc

davidfestal:
> Looks like this PR lacks tests. Are you going to add some?

@garagatyi  You're right, but I'd prefer create a separate issue for adding tests after defining precisely what level of tests we're expecting, and plan the issue in a next sprint, without blocking the rollout of this feature in prod-preview / production

@l0rd wdyt ?

davidfestal:
[test]

ibuziuk:
[test]

l0rd:
I agree with @garagatyi that this PR is lacking tests. That's ok if you want to implement them in another PR (but it would be cool if you could do it in this PR :-))

davidfestal:
That would be great if this PR could be merged as soon as possible, so that people would start populating the workspace telemetry events in the test Segment / Woopra instances. So if anything can be improved after more user experience, it could be done before rolling the whole to production.

davidfestal:
I just created issue  https://github.com/redhat-developer/rh-che/issues/694 for the record

davidfestal:
@l0rd @garagatyi So are you OK if we merge this PR now so that PMs can start using / testing the Che telemetry now on prod-preview, and we plan adding the tests for a next sprint ?

l0rd:
@davidfestal I am ok

garagatyi:
I'm ok

PR #688, opened 2018-05-28
rh-che #532 Using 'identity_id' as query parameter in oso proxy url

ibuziuk:
### What does this PR do?
Using 'identity_id' as query parameter in oso proxy url in order to support che sa oath token processing

Some notes about testing oso-proxy with support of `identity_id` query parameter:
1.  Looks like oso proxy understand `identity_id` only via vanilla `curl`. `oc` commands does not seem to work with oso-proxy e.g. 

`oc login oso-proxy-url?identity_id=<identity_id> --token=<che_sa_token> -n che-namespace`

results in

> Error in configuration: context was not found for specified context

Not sure if this is crucial for che namespace lockdown, since all interactions with oso-proxy happens via kubernetes-client [1]

2. while testing oso-proxy with k8s client I figured out that query parameters are not correctly processed:

```
          ConfigBuilder configBuilder =
                  new ConfigBuilder()
                      .withMasterUrl("https://f8osoproxy-test-dsaas-preview.b6ff.rh-idev.openshiftapps.com?identity_id=" + identityId)
                      .withNamespace("namespace-che")
                      .withTrustCerts(true)
                      .withOauthToken(cheServiceAccountToken);
          
          
          OpenShiftClient openShiftClient = new DefaultOpenShiftClient(configBuilder.build());
          List<Pod> items = openShiftClient.pods().list().getItems();
          for (Pod pod : items) {
              System.out.println(pod.getMetadata().getName());
          } 
```

This sample fails with the following error message:
```
Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://f8osoproxy-test-dsaas-preview.b6ff.rh-idev.openshiftapps.com/?identity_id=<identity_id>/api/v1/namespaces/namespace-che/pods. Message: Unauthorized! Token may have expired! Please log-in again..
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:470)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:407)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:343)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:327)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:605)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:70)
	at com.redhat.che.multitenant.Test.main(Test.java:70)

```
Basically, k8s client does not seem to honour query parameters and simply concatenates `/api/` segments to it so that `/?identity_id=<identity_id>/` becomes part of url. It was discussed  with @aslakknutsen and @nurali-techie to temporary workaround this on oso-proxy side by parsing url segment with the query parameter and removing it once `identity_id` is obtained. The real solution would be using `requestConfig.setImpersonateUsername()` for setting `identity_id` which should pass it via `Impersonate-User` header. Unfortunately, this api is not supported by the current version of k8s client we are using `3.1.0` and update to the latest version `3.1.12` is required in upstream che project. 

[1] https://github.com/fabric8io/kubernetes-client

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8WorkspaceEnvironmentProvider.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8WorkspaceEnvironmentProvider.java
@@ -97,28 +97,21 @@ public class Fabric8WorkspaceEnvironmentProvider {
 
   public Config getWorkspacesOpenshiftConfig(Subject subject) throws InfrastructureException {
     checkSubject(subject);
-
     UserCheTenantData cheTenantData = getUserCheTenantData(subject);
-
     checkClusterCapacity(cheTenantData);
 
-    String osoProxyUrl = multiClusterOpenShiftProxy.getUrl();
-    LOG.debug("OSO proxy URL - {}", osoProxyUrl);
-    String userId = subject.getUserId();
-
     ConfigBuilder configBuilder =
-        new ConfigBuilder()
-            .withMasterUrl(osoProxyUrl)
-            .withNamespace(cheTenantData.getNamespace())
-            .withTrustCerts(true);
+        new ConfigBuilder().withNamespace(cheTenantData.getNamespace()).withTrustCerts(true);
+
+    String userId = subject.getUserId();
 
     if (cheServiceAccountTokenToggle.useCheServiceAccountToken(userId)) {
       LOG.debug("Using Che SA token for '{}'", userId);
-      // TODO provide Config to oso proxy which will use Che SA token obtained from fabric-auth and
-      // userId as username
-      configBuilder.withOauthToken(cheServiceAccountToken);
+      String osoProxyUrl = multiClusterOpenShiftProxy.getUrlWithIdentityIdQueryParameter(userId);
+      configBuilder.withMasterUrl(osoProxyUrl).withOauthToken(cheServiceAccountToken);
     } else {
-      configBuilder.withOauthToken(subject.getToken());
+      String osoProxyUrl = multiClusterOpenShiftProxy.getUrl();
+      configBuilder.withMasterUrl(osoProxyUrl).withOauthToken(subject.getToken());
     }
 
     return configBuilder.build();
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/multicluster/MultiClusterOpenShiftProxy.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/multicluster/MultiClusterOpenShiftProxy.java
@@ -33,4 +33,8 @@ public class MultiClusterOpenShiftProxy {
   public String getUrl() {
     return this.url;
   }
+
+  public String getUrlWithIdentityIdQueryParameter(String identityId) {
+    return this.url + "?identity_id=" + identityId;
+  }
 }


ibuziuk:
@aslakknutsen @nurali-techie have opened PR with k8s-client version update for proper passing of identity_id via `Impersonate-User` header - https://github.com/eclipse/che-dependencies/pull/105 
client update would require some time to finally land in rh-che, so in the meantime we could use workaround with segment url parsing on oso-proxy side to tackle `*-che` namespace lockdown

l0rd:
@ibuziuk can you please try to use k8s client `3.9.1` in rh-che (overriding the upstream che dependency) and use the `Impersonate-User` header? I believe that even if that's still a workaround the technical debt would be less in this case.

ibuziuk:
@l0rd ok, lemme try to override clients versions in our assembly, so that we could avoid bothering platform team later

ibuziuk:
@l0rd @davidfestal  @garagatyi @aslakknutsen @nurali have done some tests with setting `Impersonate-User` header against oso-proxy and figured out that currently it works only for `KubernetesClient` does not work for OpenShiftClient. Details are in the issue - https://github.com/fabric8io/kubernetes-client/issues/1085 
Basically, the upstream fix looks relatively simple but this would require API changes which most likely can not be done in service releases. Also, the ugly thing is that the che sa token change can be tested only on prod-preview (on dev cluster the che sa token will not be available), so the best thing that can be done is testing oso-proxy  locally with DefaulOpenShift client (without che). So even when new release of client will land in rh-che there will be no 100% guarantee that all would work smoothly. Any thoughts ?

l0rd:
@ibuziuk it looks like you have an answer from the kubernetes-client team that may solve the problem. If it's not the case I am afraid that we should go back to the solution using the query parameter and the old k8s client version.

ibuziuk:
It is quite frustrating that trivial task (passing `identity_id` to oso-proxy) is becoming a problem. Currently I believe there are 3 options how could we fix / workaround it:

- fix passing `identity_id` via `Impersonate-User` header. This would require updating client dependencies in upstream che, providing fix with adding `ImpersonatorInterceptor` in upstream che, making new release, updating version of che in rh-che (more details in my previous comment)

- fix k8s client to process query parameters correctly. Originally we planned to pass the `identity_id` via query parameter, but k8s client does not process query parameters correctly, which results i n the following errors (query parameter becomes a url segment):

> Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://f8osoproxy-test-dsaas-preview.b6ff.rh-idev.openshiftapps.com/?identity_id=<identity_id>/api/v1/namespaces/namespace-che/pods. Message: Unauthorized! Token may have expired! Please log-in again..

In order to fix it need to provide PR to `kubernetes-client`, request service release once PR got merged, update dependencies of `kubernetes-client` in rh-che &  support query parameters in oso-proxy.

- pass `identity_id` as a query parameter, add support of `identity_id` processing via query parameter + add a temp workaround for obtaining `identity_id` from url segment in oso-proxy. This does not require any updates on che / rh-che / kubernetes-client side and I think that workaround with url segment parsing should be pretty easy to implement on oso-proxy side. Once new version of `kubernetes-client` will be in rh-che it would be trivial task to drop query parameter & segment workaround in oso-proxy and pass the `identity_id` only via header from che. The good thing about this solution that it would be possible to start che namespace lockdown right away and not be dependent on the upstream projects releases and CQs.

@l0rd @aslakknutsen @nurali-techie WDYT ?

l0rd:
@ibuziuk I am ok with solution n.3, at this point it seams the more reasonable.

nurali-techie:
@ibuziuk @l0rd @aslakknutsen , Re: sol_no_3, we see that this is not ugly hack but it's supporting bug in k8s client (where it contact url + query_param wrongly).  oso-proxy is a middleware and we would like to have clean code there as most of call pass through it.  

We would like you to see if sol_no_2 is feasible for you as it will be clean and removes the bug in k8s client.

Currently, in oso-proxy we are supporting `Impersonate-User` (i.e. sol_no_1) which was requested by your team earlier.

aslakknutsen:
sol_no_1 is the only real solution here. It's part of the Openshift/k8 API already, so no need for something custom.

ibuziuk:
@aslakknutsen @nurali-techie No doubt, that using `Impersonate-User` header is supposed to be finally the only solution. However, in order to implement it on rh-che side k8s client dependencies must be updated in the upstream che project + requesting a new release of che which rh-che would depend on (details in my comments above). This would significantly slow down the process with *-che namespace lockdown. What I was proposing is having url parsing as a **temporary** short-term solution in oso-proxy until we will would drag new version of k8s client in rhche / che and make a new release of the upstream project. This would allow us to test the lockdown on prod-preview (the **only** environment namespace lockdown could be tested against) as soon as workaround with url segment parsing would be in place. I thought this would be a fair trade-off for us...
Could you please clarify why having url segment parsing is smth. not worth temporary considering ? is it because it has some side-effects, ugly from implementation pov,  difficult to implement or the platform team has no capacity to address  it ? Might be wrong but for me it sounds like it is one extra check if the header is missing in auth.go [1] + removing `identity-id` segment from request URL as it is currently done for the header [2] ?

[1] https://github.com/fabric8-services/fabric8-oso-proxy/blob/master/middlewares/osio/auth.go#L192
[2] https://github.com/fabric8-services/fabric8-oso-proxy/blob/master/middlewares/osio/auth.go#L211

nurali-techie:
@ibuziuk @l0rd @aslakknutsen it's not matter of effort but it's matter of clean code and oso-proxy is not meant for such sort of work-around.

We are fine with @ibuziuk point of "**temporary short-term solution in oso-proxy**".  While we are preparing this **temp_fix**, @ibuziuk can you please share your plan on removing this **temp_fix**.

nurali-techie:
@ibuziuk so we will support sol_no_1 and sol_no_3 simultaneously.

**sol_no_1: support `Impersonate-User`:**
Che will set UserID Request Header `Impersonate-User` to oso-proxy.
oso-proxy will remove this header `Impersonate-User` before forwarding to OpenShift server.

**sol_no_3: temp_fix:**
Che will send request to oso-proxy as per below url format.
url_format_to_oso_proxy:
`https://f8osoproxy.xxx.com/?identity_id=<identity_id>/api/v1/namespaces/jonh-che/pods`

Here, <identity_id> is actual UserID (same thing set as `Impersonate-User` in sol_no_1).
oso-proxy will extract <identity_id> and process request.  After that while forwarding to OpenShift server the request will be as below.

url_format_to_openshift_server:
`https://f8osoproxy.xxx.com/api/v1/namespaces/jonh-che/pods`

@ibuziuk Let me know in case of any correction.

ibuziuk:
@nurali-techie thank you for considering this temp workaround. Basically, while skimming the code I got an impression that there only 2 thing that should be done to cover all potential variations (please, correct me if I am wrong):

- add extra checks while obtaining `UserID` [1]. if `identity_id` is missing in the header, need to check it in query parameter, if query parameter is also missing - check `identity_id` in url segment
- before calling `next` [2] need to remove  `/?identity_id=...../` segment from request url if it is presented


[1] https://github.com/fabric8-services/fabric8-oso-proxy/blob/master/middlewares/osio/auth.go#L192
[2] https://github.com/fabric8-services/fabric8-oso-proxy/blob/master/middlewares/osio/auth.go#L217

ibuziuk:
[test]

ibuziuk:
PR with workaround is merged and seem to work just fine https://github.com/fabric8-services/fabric8-oso-proxy/pull/23/ 

Thanks @nurali-techie ! Going to merge this PR now and it would be possible to test che namespace lockdown on prod-preview \o/

ibuziuk:
[test]

Issue #653, opened 2018-04-26
Update default image in rh-che.app.yaml after RHEL work


riuvshin:
After merge of #651 we need to update default values of image here: https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.app.yaml#L338 and point it to centos image hosted on devshift.net registry

PR #695, opened 2018-05-31
Using devshift registry instead of dockerhub

l0rd:
### What does this PR do?

Change the docker image used in the rh-che openshift template

#653

Tested deploying on dev cluster using script `deploy_custom_rh-che.sh`


openshift/rh-che.app.yaml
```diff
@@ -359,6 +359,6 @@ objects:
       name: rhche-host
 parameters:
 - name: IMAGE
-  value: rhche/rhche-server
+  value: registry.devshift.net/che/rh-che-server
```
riuvshin: this is centOS based, right?

l0rd: @riuvshin good point. I have double checked and it's centos:

```bash
$ docker run -ti --entrypoint=bash registry.devshift.net/che/rh-che-server
[root@8be18b5606a9 /]# cat /etc/centos-release
CentOS Linux release 7.5.1804 (Core)
```

Issue #626, opened 2018-04-06
Change claim-che-workspace PVC from 10GB to 1GB


rhopp:
**Issue problem:** 

When `claim-che-workspace` PVC is not present in `<username>-che` namespace, che tries to create one, but 10G. Due to OSO resource restrictions, this fails -> workspace startup fails.

Possible workaround:
Go to openshift.io -> profile -> update profile -> reset environment, or update tenant.

**Red Hat Che version:** 

version: rhche.openshift.io

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**

Delete PVC from your `-che`  namespace and try to start workspace.

**Runtime:**

runtime used:
   - rhche.openShift.io

ibuziuk:
technically speaking this situation should never happen once user will not have edit rights [1] in *-che namespace. BTW, pvc claim-che-workspace is created during update / init tenant when template [1] is applied [2] (1Gi storage / ReadWriteOnce access mode)


[1] https://github.com/redhat-developer/rh-che/issues/532
[2] http://central.maven.org/maven2/io/fabric8/tenant/packages/fabric8-tenant-che-mt/

garagatyi:
But in case of on-prem OSIO/Che or Pro accounts on OS, we can't guarantee that, can we?

ibuziuk:
well, in case of on-prem / pro I believe che server could try to create pvc. I believe it already does it in che 6 but always trying to allocate default 10G. Posting @gorkem comments about this situation:

garagatyi:
Let's do this in 2 steps:
1. Set current defaults for RhChe to 1GB - easy and fast. Also, ensure that behavior will be as expected.
1. Create an issue to try to evaluate possible value as Gorkem suggested and prioritize it accordingly.

PR #696, opened 2018-05-31
Enforce workspace PVC size to 1GB

l0rd:
### What does this PR do?

Add the env variable `CHE_INFRA_KUBERNETES_PVC_QUANTITY` to override the default PVC size. Default was 10Gi, now it's 1Gi

#626

I have tested on dev cluster that the config map and the env variable are set correctly


l0rd:
I have created the corresponding [HK issue](https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1976)

Issue #540, opened 2018-02-08
Minishift development environment setup for running rh-che multi-user locally


ibuziuk:
Need to have a possibility to run multi-user rh-che locally on minishift. This would require to 

- [ ] mock fabric8-auth 
- [ ] mock fabric8-wit (api user service) 
- [ ] mock oso proxy
- [ ] create workspaces on minishift when running locally
- [ ] document rh-che multi-user local minishift setup

Once https://github.com/redhat-developer/rh-che/issues/532 is resolved it would not be possible to deploy multi-user rh-che to dev cluster and create workspaces against prod-preview oso cluster. Basically, with removing edit right we introduce hard dependency to fabric8-auth / oso proxy components. Need to come up with approach of testing rh-che locally or on dev cluster before merging PRs to master.

l0rd:
This issue is blocking #489

l0rd:
Some thoughts during sprint planning:
- For fabric8-auth mock: adding the code for the mock in fabric8-auth github repository
- For the proxy: maybe the proxy mock is not needed and we could use the openshift endpoint instead and using a dev flag for the token to use

davidfestal:
So PR https://github.com/redhat-developer/rh-che/pull/711 fixes this issue in a way that is different of the initial issue description. Finally there's no need for any mock of fabric8 services.

Here is a short description of the proposed solution:
1. RhChe uses the Keycloak directly (as does upstream Che) instead of the `fabric8_auth` OIDC provider (this is mainly a yaml config change)
2. We use a `standalone` property in RhChe, that will be set to `true` when running it on Minishift by the minishift addon
3. When the `standalone` property is set to `true`, the RhChe Java implementation provides a Kubernetes client that points to the current cluster / namespace with the `rhche` service account instead of pointing the the OSIO user cluster / namespace with the fabric8 proxy service account.
4. In order to keep the feature that logs in the `oc` client with user OpenShift token at workspace start, the minishift account is linked to the user Keycloak account using a Keycloak Openshift identity provider.
5. Additionally, in order to avoid the need for the user to explicitly link his minishift account to his main Che Keycloak account, we switch the Keycloak autentication to use the Openshift identity provider. This way, when the user first logs into the Che server, he logs with his minishift account and the minishift and keycloak account are linked automatically. Later, when he stars a workspace, and types the `oc project` command in a terminal, he's directly connected to the default `myproject` project available in minishift by default. This provides a behavior very similar with what exists under OSIO.
[...]

PR #711, opened 2018-06-18
 Allow running rh-che in Minishift without other f8 services

davidfestal:
### What does this PR do?

Allow running rh-che in Minishift without other f8 services 

https://github.com/redhat-developer/rh-che/issues/540

yes, on Minishift


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/rh-che.properties
```diff
@@ -21,3 +21,6 @@ che.openshift.service_account.secret=NULL
 # By default disable telemetry in rh-che if it's not deployed in production or prod-preview
 che.fabric8.analytics.segment_write_key=NULL
 che.fabric8.analytics.woopra_domain=NULL
+
+# Option to run rh-che on Minishift (without accessing sibling fabric8 services) 
+che.fabric8.minishift=false
```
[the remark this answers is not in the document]
garagatyi: :+1:

davidfestal:
> I doubt that supporting of ceylon code would be productive across the whole team. 

IMO productivity mainly depends on:
- whether the given code is easily readable / understandable 
- whether the language  learning curve is small / big coming from Java. In the case of Ceylon, and especially in this code, the code should read quite naturally for a Java developer (apart from the `new` keyword that is omitted when constructing objects)... But of course my feeling might be a bit biased ;-)
 
> And because of that, all the tickets with the changes to ceylon code would probably go to the author.

I hope it wouldn't ! Since coming on Che / OSIO, I've been working on several issues that involved languages I didn't know at all: Go, Typescript, for which the learning curve coming from Java is obviously greater that Ceylon. And I'm not sure how this has dropped productivity, but surely this has brought me some valuable additional experience.

> Can you elaborate why we can't keep java here to not make codebase more complicated?

I could translate this quite small piece of code (mainly 1 function in 1 file) into Java, but I'm not sure it would make the code less complicated: I would have to add a `pom.xml` to get the Keycloak dependencies and, code-wise, the code would certainly be longer, and then would probably gain in being split into several functions / components. Ceylon here allowed keeping the code concise (a single function for something that finally is nothing more that an initialization script) , though easily readable and fully typesafe. Found it a good Ceylon use case.

Now of course, if it's a team request, I can remove this Ceylon code, but I don't see how Java code will make things better here.

openshift/minishift-addons/rhche-prerequisites/rhche-prerequisites.addon
```diff
@@ -12,5 +12,8 @@ oc adm policy add-role-to-user admin developer -n #{NAMESPACE}
 echo [CHE] Deploying Postgres
 oc new-app -f templates/postgres-template.yaml -n #{NAMESPACE}
 
+echo [CHE] Adding Minishift certificate as a secret
+ssh docker exec origin /bin/bash -c 'oc -n rhche secret new-basicauth openshift-identity-provider --username='-' --password='-' --ca-cert=./openshift.local.config/master/ca.crt'
```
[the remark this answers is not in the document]
davidfestal: Well, not really since I need to have access to the Minishfit certificate. Doing it with `oc secret` would require me to have access to this certificate in some way this is not fully supported by the limited DSL of Minishift addons.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/rh-che.properties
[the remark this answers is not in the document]
davidfestal: done

openshift/minishift-addons/rhche/rhche.addon
```diff
@@ -9,12 +9,14 @@ echo [CHE] Create the Che server Template
 oc apply -f templates/rh-che.app.yaml -n openshift
 
 echo [CHE] Deploying Che (rh-che)
-oc apply -f templates/rh-che.secret.yaml -n #{NAMESPACE}
 oc apply -f templates/rh-che.config.yaml -n #{NAMESPACE}
-oc new-app -p IMAGE_CHE=#{CHE_DOCKER_IMAGE} -p CHE_VERSION=#{CHE_VERSION} -p ROUTING_SUFFIX=#{routing-suffix} rhche -n #{NAMESPACE} 
+oc new-app -p IMAGE_CHE=#{CHE_DOCKER_IMAGE} -p CHE_VERSION=#{CHE_VERSION} -p ROUTING_SUFFIX=#{routing-suffix} -p NAMESPACE=#{NAMESPACE} -p PROTOCOL=http  -p MINISHIFT_CONSOLE=https://#{ip}:8443 rhche -n #{NAMESPACE}
```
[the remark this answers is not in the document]
davidfestal: Adding those parameters for the main rhche container is not necessary. These 2 properties are required for the Keycloak initialization command run in the initContainer. That's why I didn't name them the same as the corresponding rhche properties that in fact are not required in our case.
But I can still rename `MINISHIFT_CONSOLE` to `OPENSHIFT_API_URL`, or even `OPENSHIFT_PROVIDER_BASE_URL`, which is what it is used for in the Keycloak initialization

openshift/minishift-addons/rhche/templates/rh-che.app.yaml
```diff
@@ -62,6 +73,12 @@ objects:
         labels:
           app: rhche
       spec:
+        initContainers:
+          - image: ' '
```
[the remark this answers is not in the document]
davidfestal: > Why is this field empty?

Because the image is defined later in the trigger definition: https://github.com/redhat-developer/rh-che/pull/711/files/6354c831ca5b3c9a4f131ce983eb6e9101e4bdbe#diff-f33365d04d532869c54d961801882a5eR294 

> Anyway why aren't you executing the keycloak configurator when the rh-che prerequisites are deployed (i.e. keycloak-template.yaml)?

Several reasons led me to do this:
- to ensure Keycloak is correctly configured *before* the RhChe server starts
- What would trigger the start of this command ? We might do this in a post hook in the Keycloak deployment config, but then we should add the corresponding code in the docker image of the Keycloak main container. This implies creating a dedicated Keycloak docker image for RhChe which would derive from the `eclipse/che-keycloak` image... Is it what we want ?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/rh-che.properties
```diff
@@ -21,3 +21,6 @@ che.openshift.service_account.secret=NULL
 # By default disable telemetry in rh-che if it's not deployed in production or prod-preview
 che.fabric8.analytics.segment_write_key=NULL
 che.fabric8.analytics.woopra_domain=NULL
+
+# Option to run rh-che in standalone mode, typically on on Minishift (without accessing sibling fabric8 services)
```
ibuziuk: typo "on on Minishift"

ibuziuk:
@davidfestal would it be possible to add some docs with a basic explanation of how `keycloak-configurator` works and how to test it on minishift ? This is an impressive and super useful change but frankly I failed to understand how it mocks wit / auth / oso-proxy etc

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/OpenshiftUserTokenProvider.java
```diff
@@ -43,8 +48,23 @@
 
   @Inject
   public OpenshiftUserTokenProvider(
```
ibuziuk: Will it be relevant to use `che.fabric8.standalone`. Currently it is a bit unclear in which case which config is used

openshift/minishift-addons/rhche/templates/rh-che.app.yaml
```diff
@@ -340,4 +328,4 @@ parameters:
   displayName: OpenShift Online Proxy Service
   description: URL of the OpenShift Online Proxy service
   value: '${PROTOCOL}://f8osoproxy-${NAMESPACE}.${ROUTING_SUFFIX}'
-
+
```
ibuziuk: New line in the end

.ci/cico_do_docker_build_tag_push.sh
```diff
@@ -15,6 +15,10 @@ RH_CHE_TAG=$(git rev-parse --short HEAD)
 UPSTREAM_TAG=$(sed -n 's/^SCM-Revision: \(.\{7\}\).*/\1/p' ${ABSOLUTE_PATH}/../assembly/assembly-wsmaster-war/target/war/work/org.eclipse.che/assembly-wsmaster-war/META-INF/MANIFEST.MF)
 
 # Now lets build the local docker images
+
+ADDONS=${ABSOLUTE_PATH}/../openshift/minishift-addons
+KEYCLOAK_CONFIGURATOR_IMAGE="rh-che-standalone-keycloak-configurator"
```
ibuziuk: maybe `KEYCLOAK_STANDALONE_CONFIGURATOR_IMAGE` is a better name ?

davidfestal:
> would it be possible to add some docs with a basic explanation of how keycloak-configurator works and how to test it on minishift ? This is an impressive and super useful change but frankly I failed to understand how it mocks wit / auth / oso-proxy etc

I added some explanations about this PR work [in the initial issue](https://github.com/redhat-developer/rh-che/issues/540#issuecomment-398702336)

.ci/cico_do_docker_build_tag_push.sh
davidfestal: yes, you're right

davidfestal: done

openshift/minishift-addons/rhche/templates/rh-che.app.yaml
davidfestal: fixed

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/OpenshiftUserTokenProvider.java
davidfestal: I fixed this in commit https://github.com/redhat-developer/rh-che/pull/711/commits/80846aa6d7d128530093ff874446af01162af0d9

davidfestal:
@l0rd @ibuziuk I made the required changes and tested the result on both Minishift and dev cluster
is it OK for you ?

ibuziuk:
@davidfestal approved :+1: @ScrewTSW do you happen to know if PR verification against dev cluster  is working currently ?

ScrewTSW:
@ibuziuk It should be working. let me try.

ScrewTSW:
[test]

davidfestal:
@l0rd @garagatyi is this PR OK for you now ?
@ibuziuk the PR check build fails, but it seems to be the same error as [this one](https://ci.centos.org/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/234/) which occurred on a 0-change [PR](https://github.com/redhat-developer/rh-che/pull/715). So the build failure might not be related to this PR changes.

ScrewTSW:
[test]

ibuziuk:
yeah, it looks like there a problem with PR job itself - https://ci.centos.org/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/ (AFAIK it was successful only once) @ScrewTSW is there are any plans to stabilize PR job in the next sprint ?

ScrewTSW:
@ibuziuk If we manage to finish what we have planned, there will be no need to stabilize these jobs :)

PR #721, opened 2018-06-26
226 feature che upstream selenium tests

ScrewTSW:
DO NOT MERGE until `upstream` has been back-ported!

### What does this PR do?
This issue adds new test framework based on Selenium, using eclipse/che upstream core and test modules.
Upstream che related PR : eclipse/che/pull/10191

### What issues does this PR fix or reference?
redhat-developer/che-functional-tests/issues/266

### How have you tested this PR?
Manual testing

PR #724, opened 2018-06-27
Adding build status for periodic ci jobs

ibuziuk:
### What does this PR do?
Adding build status for periodic ci jobs:
- https://ci.centos.org/view/Devtools/job/devtools-che-functional-tests-periodical-prod-preview.openshift.io-2a/
- https://ci.centos.org/view/Devtools/job/devtools-che-functional-tests-periodical-prod-preview.openshift.io-free-stg/

N/A

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 [![Build Status](https://ci.centos.org/buildStatus/icon?job=devtools-rh-che-build-che-credentials-master)](https://ci.centos.org/view/Devtools/job/devtools-rh-che-build-che-credentials-master/)
-
 [![Periodic tests against free-stg (prod-preview)](https://ci.centos.org/buildStatus/icon?job=devtools-che-functional-tests-periodical-prod-preview.openshift.io-free-stg)](https://ci.centos.org/view/Devtools/job/devtools-che-functional-tests-periodical-prod-preview.openshift.io-free-stg/)
-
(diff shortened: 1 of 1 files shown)


ibuziuk:
IMO, this would rise visibility of the periodic tests and we should focus on having all of those jobs stable

Katka92:
@ibuziuk I absolutely agree with rising visibility of test results, but I could not agree with having these particular tests as stable as possible. We tried to, but it is almost impossible to keep pace with upstream changes. Therefore were are changing our tests to use upstream artifacts [1]. Once it will be done, then we should focus on maximum stability of these tests as a priority before adding new test cases. 

[1] https://github.com/redhat-developer/rh-che/pull/721

README.md
```diff
@@ -1,4 +1,6 @@
 [![Build Status](https://ci.centos.org/buildStatus/icon?job=devtools-rh-che-build-che-credentials-master)](https://ci.centos.org/view/Devtools/job/devtools-rh-che-build-che-credentials-master/)
```
Katka92: What about having any labels (displayed by default or by hoovering)? It is not clear which job is failing seeing only:

screenshot from 2018-06-28 13-00-48

ibuziuk: MD format does not support hovering afaik (it is possible to add it by inlining html but this is not recommended )

ibuziuk:
@Katka92 going to close this PR since it would be better to add this to `functional-tests` README.md https://github.com/redhat-developer/rh-che/pull/721/files#r198805676 Will you be able to add it there as part of https://github.com/redhat-developer/che-functional-tests/issues/283 ?

PR #733, opened 2018-07-04
Create README for functional tests.

Katka92:
### What does this PR do?
Creates README for functional tests.


functional-tests/README.md
```diff
@@ -0,0 +1,71 @@
...
+* [When are tests running](#when-are-tests-running)
+
+## What are these tests meant for
+
+These functional tests serve to verify that Eclipse Che on Openshift is working well. Tests include integration specific cases and some upstream tests to
+verify Che itself.
+
+There are some jobs running periodically on clusters us-east-2, us-east-2a and free-stg as you can see in the begining of this readme.
+
+## How to run it
```
ibuziuk: @Katka92 `How to run` only says about env var. Could you add some info about what should be done once env vars are set ? I assume it is `mvn clean install`. Also would be great to have a clarification how to run individual tests

ibuziuk:
@Katka92 very goog README. Could you please address one comment and I believe we can merge

ScrewTSW:
[test] just testing pr-check (postgres fails to deploy and times out)

Katka92:
@ibuziuk I've added information that we run them via TestNG. In the future there should be script or something like that to simplify running them.

rhopp:
@Katka92 When you will be addressing the comments made by Ilya, could you please also add new parameted introudced in https://github.com/redhat-developer/rh-che/pull/767 - https://github.com/redhat-developer/rh-che/pull/767/files#diff-ddd2c886eb59ccb6894edbd26b9fcebeR38

Katka92:
@ibuziuk Now we do not have any tests and upstream tests can be run only as a part of suite. So I would like to merge this PR as an initial README and add additional info when we know exactly the right steps to run these tests etc. 

@rhopp I added this parametr and few words for explanation. Please review.

Issue #703, opened 2018-06-06
Need to pass `identity_id` via `Impersonate-User` header instead of query parameter


ibuziuk:
Currently `identity_id` is passed to oso-proxy via query parameter. Unfortunately, k8s client does not proceed query parameters correctly when set in `masterUrl` (details in the upstream issue https://github.com/fabric8io/kubernetes-client/issues/1089 ) and to make it work a workaround [2] was added to `oso-proxy` [2]. Assuming that oso-proxy currently support both approaches of `identity_id` processing this issue should be treated as internal implementation improvement which does not block *-che namespace lockdown.

[1] https://github.com/redhat-developer/rh-che/pull/688
[2] https://github.com/fabric8-services/fabric8-oso-proxy/pull/26

l0rd:
We should fix that as soon as we move to 6.7.0

ibuziuk:
PR is merged - https://github.com/redhat-developer/rh-che/pull/1089
Closing

PR #737, opened 2018-07-06
rh-che #703: Using 'Impersonate-User' header instead of query parameter for passing identity_id to oso-proxy

ibuziuk:
### What does this PR do?
Using 'Impersonate-User' header instead of query parameter for passing identity_id to oso-proxy

https://github.com/redhat-developer/rh-che/issues/703

Due to che sa token & oso-proxy dependency it is possible to fully test the PR only on prod-preview environment


plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8WorkspaceEnvironmentProvider.java
```diff
@@ -115,17 +116,19 @@ public Config getWorkspacesOpenshiftConfig(Subject subject) throws Infrastructur
...
+    String osoProxyUrl = multiClusterOpenShiftProxy.getUrl();
 
     if (cheServiceAccountTokenToggle.useCheServiceAccountToken(userId)) {
-      LOG.debug("Using Che SA token for '{}'", userId);
-      String osoProxyUrl = multiClusterOpenShiftProxy.getUrlWithIdentityIdQueryParameter(userId);
-      configBuilder.withMasterUrl(osoProxyUrl).withOauthToken(cheServiceAccountToken);
+      LOG.info("Using Che SA token for '{}'", userId);
+      config =
+          configBuilder.withMasterUrl(osoProxyUrl).withOauthToken(cheServiceAccountToken).build();
+      LOG.info("Adding Impersonate Header '{}'", userId);
```
garagatyi: Do we need those 2 logs entries on an info level?

ibuziuk: just for a short period of time. Eventually, those log entries are going to be removed. For now those two on the `info` level are handy due to investigation of proxy issues like https://github.com/openshiftio/openshift.io/issues/3884

ibuziuk:
tested on `minishift` and impersonate header processing works - `ImpersonatorInterceptor` kicks in and adds `subject.getUserId()` as a an `Impersonate-User` to every request

PR #738, opened 2018-07-09
Changing default pull policy to `IfNotPresent` for rh-che minishift addon

ibuziuk:
Currently if I want to use custom image with `rhche` addon I have to do smth. like that:

```
minishift addons apply \
    --addon-env RH_CHE_DOCKER_IMAGE=ibuziuk/che-server \
    --addon-env RH_CHE_VERSION=rhche1 \
    --addon-env PULL_POLICY=Never \
    rhche
```

It would be great to change default pull policy from "Always" to "IfNotPresent" by default imo

### DIFF
--- a/openshift/minishift-addons/rhche/templates/rh-che.app.yaml
+++ b/openshift/minishift-addons/rhche/templates/rh-che.app.yaml
@@ -314,8 +314,8 @@ parameters:
   value: 'Recreate'
 - name: PULL_POLICY
   displayName: Che server image pull policy
-  description: Always pull by default. Can be IfNotPresent
-  value: 'Always'
+  description: Pull the image only if not present in the local registry by default. Can be 'Always' and 'Never'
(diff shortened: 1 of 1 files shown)


ibuziuk:
currently PR check fails with 

```
export_env_variables: line 10: rhche: command not found
export_env_variables: line 10: rnminishift: command not found
export_env_variables: line 10: export: `smth.': not a valid identifier
export_env_variables: line 10: export: `that:rnrnrnrnIt': not a valid identifier
export_env_variables: line 10: export: `"Always"': not a valid identifier
export_env_variables: line 10: export: `"IfNotPresent"': not a valid identifier
```

https://ci.centos.org/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/270/console
Need to investifgate

ibuziuk:
[test]

Issue #749, opened 2018-07-11
Add tracking of factories in Telemetry


davidfestal:
We should track how workspaces are created from factories. See this issue comment https://github.com/redhat-developer/rh-che/issues/709#issuecomment-404136882

- [x] When a workspaces is started from a factory
- [x] What factory has been started
- [x] When an account is created on OSIO from a user who clicked on a factory.

This seems to be a pre-requisite to finalizing the migration of factories (https://github.com/redhat-developer/rh-che/issues/729).

slemeur:
Thanks @davidfestal !

ibuziuk:
> When an account is created on OSIO from a user who clicked on a factory.

@slemeur @l0rd  I also commented in the original issue that this event is not really che related [1], but rather related to the registration flow - https://github.com/openshiftio/openshift.io/issues/3814 

Which team do you think should be in charge of the processing and adding to woopra registration events ?  

IMO the woopra registration event should not be limited to Che factory usecase but rather track in general, from which page the user was redirected to the registration page (I believe it would be quite easy to do via `document.referrer`[2] on the registration side and super useful from the marketing perspective) 

[1] https://github.com/redhat-developer/rh-che/issues/709#issuecomment-404141100
[2] https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

slemeur:
> IMO the woopra registration event should not be limited to Che factory usecase but rather track in general, from which page the user was redirected to the registration page (I believe it would be quite easy to do via document.referrer[2] on the registration side and super useful from the marketing perspective)

I agree with you @ibuziuk . 
In fact we are interested by all accounts created from the page here: https://github.com/openshiftio/openshift.io/issues/3814 and should be using the referrer to know if the user was coming from a factory or not.

I can create a dedicated issue for that. I believe Auth team should probably be the appropriate team to handle that. wdyt @alexeykazakov ?

ibuziuk:
cc: @sbose78 since Alexey is on PTO till the end of the month

davidfestal:
Let's summarize what has been done for this issue:

PR https://github.com/redhat-developer/rh-che/pull/760 has added some meaningful properties to Telemetry workspace events, that allow tracking factory and stack usage.
More precisely it provides:
- the *__stack Id__* when the started workspace was initially created from a Dashboard stack,
- the *__factory Id, factory owner and factory name__* when the started workspace was initially created from a factory,
[...]

PR #760, opened 2018-07-17
Add properties to Telemetry events to track factory and stack usages

davidfestal:
### What does this PR do?

This PR adds some meaningful properties to Telemetry workspace events, that allow tracking factory and stack usage.
More precisely it provides:
- the *__stack Id__* when the started workspace was initially created from a Dashboard stack,
- the *__factory Id, factory owner and factory name__* when the started workspace was initially created from a factory,
- a *__first start__* boolean property that indicates if this is the first time a workspace is started after its creation
- the *__age__* of a started workspace (last start date - creation date) which is important to track for workspace *creations* from factories
- the *__return delay__* of started workspaces, which is the timeframe during which a workspace has been left unused before being restarted, and can give an insight of how frequently workspaces are restarted. 

The 2 last properties can probably be useful more generally to measure user engagement / on-boarding, for all types of workspaces, not only factories.

https://github.com/redhat-developer/rh-che/issues/749

Yes, new properties have been tested on dev-cluster with the `test.openshift.io` Woopra instance and new reports that leverage them.


davidfestal:
[test]

ibuziuk:
[test]

PR #761, opened 2018-07-18
[FT] Fist simple suite created.

Katka92:
### What does this PR do?
The first simple suite was created. It contains only one test class (DialogAboutTest) from upstream. Serves as base template for other tests and test suites.

Tested locally against production us-east-2 cluster.


Katka92:
[test]

Katka92:
[test]

Katka92:
[test]

functional-tests/src/test/java/com/redhat/che/functional/tests/simpleTestSuite.xml
```diff
@@ -0,0 +1,49 @@
...
+       verbose="10"
+       parallel="none"
+       thread-count="1">
+
+    <test name="all">
+        <method-selectors>
+            <method-selector>
+                <script language="beanshell">
+                    <![CDATA[
+    boolean isTestGroupsAgreedWithRequestedGroups() {
```
garagatyi: Can you elaborate why we need to hardcode real code in XML?

rhopp: I was asking myself the same question. And after thinking this through, I think we should keep this part in here.

In upstream, they have the same code in their test suites. This code ensures, that no testcases annotated with `@Test(groups="<someTestGroup>")` are executed. 

As we want to use upstream test classes to run tests against rh-che, this mechanism can come in handy to exclude some groups of testcases from our suite.

garagatyi: Thank you for the explanation, but the approach is awful. We shouldn't have such an approach upstream.

ibuziuk:
[test]

ibuziuk:
@rhopp hmm.. why che-server route is still using TLS ^ ? and what is more interesting why the tests are passing ?

rhopp:
@ibuziuk The PR check is not using TLS anymore... The status message reported by jenkins in here, saying che is deployed on `https://che....` is old one. I've already opened PR to almighty jobs, which should fix that: https://github.com/openshiftio/openshiftio-cico-jobs/pull/727

functional-tests/src/test/java/com/redhat/che/functional/tests/simpleTestSuite.xml
Katka92: Ok so should I remove it or leave it here until some better approach is found? As @rhopp said, this functionality can be handy.

rhopp: I'll try to reach to upstream Che QE today (or tomorrow) and discuss this one and see if we can find another (nicer) solution ;-)

rhopp: After discussing with upstream che-qe I would go with removing this code from here as we don't `need` it right now and upstream is thinking about changing that too, so hopefully the new solution will be in place, when we need it.
Che-qe team is having some troubles with this approach also and they promised to invite us to the meeting, where we could discuss this and try to solve
> iokhrimenko: We have planned to discuss our groups approach with the Dmytro at Monday because we have several problems with this functionality, and I created some temporary workaround. So if you wish I can invite you to our planning and you will take part in discussion

Katka92:
@rhopp @garagatyi I've removed this part with "awful" code. :) Should it be merged now or is there something else that needs any adjustment?

garagatyi:
@Katka92 I didn't mean to reject this PR because of the solution you took from the upstream project. It is not our fault that the approach is not good enough from the supportability point of view. And I was completely OK to merge with that hardcoded script. Sorry If I didn't communicate this clearly.

Katka92:
@garagatyi You and rhopp agreed that the approach is not good (I agree with it) and as rhopp said, we don't "need" it now. As there is a talk about this scheduled on Monday, I agree with removing this code and replacing it later with new solution. I'm glad you share your opinion and experience with us!

functional-tests/src/test/java/com/redhat/che/functional/tests/simpleTestSuite.xml
dmytro-ndp: Removal of *beanshell* script from the test suite xml breaks exclusion of tests which are not relevant to current Che infrastructure. At the same time that script doesn't allow to define test dedicated to the several Che infrastructure (not all) at a time (for example, just for _OpenShift_ and _OSIO_).
I have created an issue https://github.com/eclipse/che/issues/10583 to fix this problem.

PR #767, opened 2018-07-19
Removed code starting che-starter

rhopp:
### What does this PR do?
Removed code, which was starting che-starter, from codebase. Now the tests expect the che-starter to be started before executing tests (most possibly by running che-starter docker image - not part of this PR).

This allows us to get rid of few, now unnecessary, dependencies.

https://github.com/redhat-developer/che-functional-tests/issues/305

build locally fine, Run dummy test case with che-starter started (passed) & Run dummy test case with che-starter stopped (failed as expected)

### DIFF
--- a/functional-tests/pom.xml
+++ b/functional-tests/pom.xml
@@ -42,16 +42,6 @@
                 <!-- <version>${che.version}</version> -->
                 <version>6.8.0</version>
             </dependency>
-            <dependency>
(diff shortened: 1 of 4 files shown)


rhopp:
I've added a note to https://github.com/redhat-developer/rh-che/pull/733 to document the new parameter introduced in this PR.

PR #777, opened 2018-07-27
Adding steps after Rh-Che build to prepare dependencies for pr-check

ScrewTSW:
### What does this PR do?
Prepares files for new jenkins job that builds PR-Check dependency docker file

https://github.com/redhat-developer/che-functional-tests/issues/312

no testing done yet.


ScrewTSW:
This new job takes cares of pre-building a docker image which will have the dependencies used in the project already downloaded and it will also have latest-at-the-time packages installed.  
  
It will be triggered after every merge to master in rh-che to keep the versions up to date.  
This should take the bulk of downloads off every pr-check run. If there is a new version introduced in the PR, only that library, and it's dependencies will be downloaded.

ScrewTSW:
[test]

ScrewTSW:
@garagatyi Can do, though this PR only adds the extra step at the bottom of the diagram. Will create a separate PR with the default diagram in that case :)

ScrewTSW:
@garagatyi From the manual testing I have done so far it seems it's somewhere around 10 minutes of time saved when using the modified dependency image :) We will see when the new modified PR check runs more often, however that's something will will be able to monitor more closely with https://github.com/redhat-developer/rh-che/pull/783.

This PR only provides the Dockerfile and scrips required for building https://ci.centos.org/view/Devtools/job/devtools-rh-che-prcheck-build-dep-dev.rdu2c.fabric8.io.openshift.io/

PR #787, opened 2018-08-01
Cleaning up stacks

eivantsov:
### What does this PR do?

**New stacks**:

* php (LS supported)
* golang (LS supported)
* ruby
* python (LS supported)
* ceylon
* Theia
* Java + MySQL (2 containers in 1 pod)

**Fixes**:

* wildfly - rename artifact
* java-centos - add missing server
* most stacks - remove unnecessary tags so that UD can match the right stacks with the right sample apps
* updating components to actual versions of Java, Maven, Node etc.

### Removing stacks with compose recipe type

Since Che server cannot remove stacks that are already in DB, we should manually get rid of:

* https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L3
* https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L83

@riuvshin do you have credentials of rh-che admin user?

## Testing

I have tested most of the stacks + sample apps, however, there might be little things like missing server or a sample app that does not work. I hope we're talking about 1-2 such cases. Maybe zero though :)

### DIFF
--- a/assembly/fabric8-stacks/src/main/resources/stacks.json
+++ b/assembly/fabric8-stacks/src/main/resources/stacks.json
@@ -1,36 +1,96 @@
 [
   {
-    "id": "java-centos-mysql",
+    "id": "simple-theia",
     "creator": "ide",
-    "name": "Java-MySQL CentOS",
-    "description": "Multi-machine environment with Java CentOS Stack and MySQL database",
+    "name": "Theia IDE",
+    "description": "Theia IDE With JS and Java Tooling",
     "scope": "general",
     "tags": [
-      "Java 1.8, Tomcat 8, MySQL 5.7"
+      "Theia",
+      "Alpine",
+      "CentOS"
     ],
-    "components": [
-      {
-        "name": "JDK",
-        "version": "1.8.0_45"
-      },
-      {
-        "name": "Maven",
-        "version": "3.2.2"
-      },
-      {
-        "name": "Tomcat",
-        "version": "8.0.24"
-      }
-    ],
-    "source": {
-      "type": "image",
-      "origin": "registry.devshift.net/che/centos_jdk8"
-    },
     "workspaceConfig": {
       "environments": {
         "default": {
           "machines": {
-            "dev-machine": {
+            "theia": {
+              "servers": {
+                "theia": {
+                  "attributes": {
+                    "type": "ide"
+                  },
+                  "protocol": "http",
+                  "port": "3000",
+                  "path": "/"
+                },
+                "theia-dev": {
+                  "attributes": {
+                    "type": "ide-dev"
+                  },
+                  "protocol": "http",
+                  "port": "3030"
+                }
+              },
+              "installers": [],
+              "volumes": {
+                "projects": {
+                  "path": "/projects"
+                }
+              },
+              "env": {
+                "CHE_MACHINE_NAME": "theia",
+                "HOSTED_PLUGIN_HOSTNAME": "0.0.0.0"
+              },
+              "attributes": {
+                "memoryLimitBytes": "2147483648"
+              }
+            }
+          },
+          "recipe": {
+            "contentType": "",
+            "type": "dockerimage",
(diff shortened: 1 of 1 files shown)


rhopp:
[test]

rhopp:
[test]

eivantsov:
@ibuziuk should they? I mean they are available here https://github.com/eclipse/che-dockerfiles/tree/master/recipes

Not a problem though to move over centos_go, android or ceylon ones.

I think your point is to have one CI for all che.osio stacks, right?

ibuziuk:
@eivantsov yeah, I think having all che osio related images being built as part of redhat-developer/che-dockerfiles used to be a requirement.  @l0rd @riuvshin could you please confirm it is the case ?

eivantsov:
@ibuziuk I am fine with moving them over. I never did that since images already exist. And I created Dockerfiles only for new images.

l0rd:
@ibuziuk @eivantsov I think you should consider a couple of things:
1. We created `redhat-developer/che-dockerfiles` in order to be able to run upstream images on che.openshift.io. At the time upstream images would not start on openshift online. We had to modify them: 
    - inject some tools like `oc` (via [this script](https://github.com/redhat-developer/che-dockerfiles/blob/master/recipes/context/general-deps.install))
    - setup permissions on folders (via [this script](https://github.com/redhat-developer/che-dockerfiles/blob/master/recipes/context/grant-access-arbitrary-UID.sh))
    - make sure that the container will be able to run with arbitrary UID
    - pre-install the boosters dependencies
    - ...
2. Stack images of rh-che are not based on dockerhub images. They are all based on [registry.centos.org](https://registry.centos.org/che-stacks/) images. The reasons are that we own that registry and that the build/push of these images happens automatically when the dockerfiles are changed.

So, as far as the image can run on openshift online, the right tooling is included and we don't need to pre install boosters dependencies, we can avoid having a corresponding Dockerfile in `redhat-developer/che-dockerfiles`.

But the new stacks should be based on the registry.centos.org ones (not dockerhub):
- registry.centos.org/che-stacks/centos-go
- registry.centos.org/che-stacks/centos-ceylon-nodejs-dart

eivantsov:
All of the upstream images can run on che.openshift.io. Some of them do not have oc. But do Android or say Golang need it? As far as I understand oc is required for maven projects that use fabric8 plugin.

l0rd:
Yes originally `oc` was added to make some configmap test pass. It may help to interact with openshift online API too tough (to deploy the application beeing developed for example). 

But I agree that's not really required at this stage.

ibuziuk:
@l0rd thanks for clarifications. 
So, it seems to be fine that some of the images are build as part of centos container pipeline and others are build via devtools centos CI as long as `registry.centos.org` is used
@eivantsov for me the only concern currently is `simple-theia` which is based on dockerhub[1] image. 
Do you think we can exclude it for now and add later once the theia image would be pushed to `registry.centos.org` ?
https://hub.docker.com/r/eclipse/che-theia/tags/

eivantsov:
@ibuziuk I think having Theia stack is important but not critical, since we will have Factories with Theia as an IDE anyway.

While I own runtime upstream images, Theia image has own CI. @benoitf may shed some light on if it's possible not to use Docker Hub.

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

ScrewTSW:
[test]

rhopp:
[test]

eivantsov:
@rhopp thanks for checking

@ibuziuk the only thing left is Theia image hosted on DockerHub.. but it's @benoitf and IDE2 team that own that image and CI for it.

ibuziuk:
@eivantsov I'm ok to merge it in the current state and have separate issue for hosting Theia image on `registry.centos.org`

Issue #791, opened 2018-08-01
Add telemetry events + telemetry user identification in the automatic registration flow script.


davidfestal:
related to https://github.com/redhat-developer/rh-che/issues/773 and https://github.com/redhat-developer/rh-che/issues/790

slemeur:
Needed for https://github.com/redhat-developer/rh-che/issues/701 and https://github.com/redhat-developer/rh-che/issues/765

davidfestal:
Deployed to production.

Issue #790, opened 2018-08-01
Transform the direct user registration flow POC to a real implementation integrated into RhChe


davidfestal:
The POC was this issue: https://github.com/redhat-developer/rh-che/issues/773

slemeur:
Needed for https://github.com/redhat-developer/rh-che/issues/701

davidfestal:
Deployed to production.

Issue #546, opened 2018-02-12
Retrieve the `keycloak.js` file from the official NPM package instead of bundling our own


davidfestal:
Retrieve the `keycloak.js` from the official Keycloak NPM package to include it into upstream Che, now that PR https://github.com/keycloak/keycloak/pull/4978 has been merged.

Of course, we should wait for a new Keycloak version to be released first.

skabashnyuk:
Is it published on NPM registry already?

ibuziuk:
@davidfestal the PR https://github.com/keycloak/keycloak/pull/4978 has been merged, should we prioritize this issue to make the auth flow more sustainable ?

davidfestal:
Not sure it's really high-priority currently.

ibuziuk:
@davidfestal sure it might be not a priority, just wondering if this issue is still relevant so that we could prioritize it (I'm doing issue triage and want to close irrelevant issues)

ibuziuk:
Closing as won't fix

PR #799, opened 2018-08-08
End to end user registration flow with telemetry support.

davidfestal:
### What does this PR do?

This PR adds an acceptable end-to-end registration flow for brand new users that should come into a Che Workspace from a precise Dashboard link, such as a factory link.

The additional required provisoning / setup steps are chained if necessary during the Dashboard initial user authentication.

This allows also adding some telemetry steps to identify the user into the telemtry, as well as track provisioning and dashboard start events.
New telemetry events introduced by the PR:
- `Provision User For Che`
- `Enter Che Dashboard`

The flow from a factory link looks like this for an existing RHD user that wasn't already provisioned for OSIO:

https://youtu.be/JqD6D27_nvE

Flow would be similar for a brand new RHD user (apart from the initial registration + email verification steps that are not a problem since they the factory link in the final redirection).

#790 
#791 

Yes, on dev cluster and through PR checks as well.

### DIFF
--- a/.ci/cico_rhche_prcheck.sh
+++ b/.ci/cico_rhche_prcheck.sh
@@ -125,8 +125,7 @@ if ./dev-scripts/deploy_custom_rh-che.sh -u "${RH_CHE_AUTOMATION_RDU2C_USERNAME}
                                          -t nightly-"${RH_TAG_DIST_SUFFIX}" \
                                          -e che6-automated \
                                          -s \
-                                         -U \
-                                         -z;
+                                         -U;
 then
   echo "Che successfully deployed."
 else
--- a/assembly/fabric8-ide-gwt-app/src/main/resources/org/eclipse/che/public/keycloak/OIDCKeycloak.js
+++ b/assembly/fabric8-ide-gwt-app/src/main/resources/org/eclipse/che/public/keycloak/OIDCKeycloak.js
@@ -1,0 +1,1409 @@
+/*
+ * Copyright (c) 2016-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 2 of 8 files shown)


davidfestal:
@garagatyi My bad, I should have explained that:
- `OIDCKeycloak.js` is mainly the same as the one in upstream project (which is turn is nearly the same as the Keycloak upstream project `keycloak.js` file: see issue #546), apart from a very light fix that I will also propose in a PR on the upstream Keycloak project.
- `provision.html` is the source of the HTML page that provides UI for the OSIO registration request, which is mainly based on a copy of the OSIO login page.
- `RhCheKeycloak.js` is the main place where things happen, and that might / should be reviewed...

Here is the overall behavior:

When using the `RhCheKeycloak.js` as the new JS OIDC adapter file, it will delegate the real Keycloak initialization to the original `OIDCKeycloak.js` adapter file, but add the necessary steps to:
- ask the user to provision for OSIO if necessary (there it displays the `provision.html` page)
- chain the required user preparation steps (account linking, namespace setup, etc ...) as soon as the user can be authenticated (since it is provisioned for OSIO).

davidfestal:
BTW I think we might open issues for next sprints about:
- creating the corresponding PR on upstream Keycloak for the required change in `OIDCKeycloak.js`
- explore how the `provision.html` page could written in a way that would ensure it keeps in sync with the future state of the RHD Openshift.io Keycloak customization.

rhopp:
[test]

rhopp:
The PR check failed, but I guess it's unrelated to this PR... Version from this PR is deployed here: http://rhche-che6-automated.dev.rdu2c.fabric8.io/ (until another PR check replaces that).
I've done some VERY quick testing - workspace startup works, no changes in normal login workflow, ... I would say we're OK to merge this (since @garagatyi is ok with this PR also) and we can do more testing tomorrow on prod-preview, before promoting to prod.

davidfestal:
@garagatyi Do you think you could review / approve it again: I had to rebase the PR to include the downgrade to the 6.7.1 upstream release ?

ScrewTSW:
[test]

Issue #727, opened 2018-06-28
Impossible to import from github account from the dashboard


slemeur:
**Issue problem:** 

**Red Hat Che version:** 
version: 6.7.1

**Reproduction Steps:**
1. Login to che.openshift.io
2. Create a new workspace
3. Select a stack
4. Go into the section to add a new project. Select "Github"
5. Click on the button connect to your github account
![screen shot 2018-06-28 at 20 30 31](https://user-images.githubusercontent.com/1636769/42053671-71169886-7b12-11e8-9d98-0d959db266ed.png)

**Expected Behavior:**
User is already authorized on github (because asked by osio early in the flow) and is able to browse the list of his github projects.

**Current Behavior:**
I have a failure and github showing a 404.
![screen shot 2018-06-28 at 20 32 42](https://user-images.githubusercontent.com/1636769/42053669-6d3d8d50-7b12-11e8-896a-280674351ca8.png)

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [x] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

slemeur:
This behaviour is also observed when trying to use the GitHub button in Preferences > SSH

slemeur:
Looking for some news on this issue: will it be addressed during the sprint? I don't see an assignee on it.

sbose78:
You could use 
```
GET https://auth.openshift.io/api/token/link?for=openshift + token
```
to trigger 'connect github'

and 

```
GET https://auth.openshift.io/api/token?for=openshift + token
```
to get the github token, and/or check if user has connected her github.

PR #810, opened 2018-08-21
rh-che Issue #727 - Enable linking github account from Che dashboard

amisevsk:
### What does this PR do?
Enables communication with fabric8 auth API to allow linking of github accounts from the dashboard and obtaining a github token in Che. All token management is done through fabric8 auth server, so linking a github account will link it for user's openshift.io account.

#727 

Manually on dev-cluster using prod-preview's auth server. Deployable image is available at `amisevsk/rhche-server:dev`.

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/pom.xml
+++ b/plugins/fabric8-multi-tenant-manager/pom.xml
@@ -52,6 +52,10 @@
             <artifactId>kubernetes-model</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt</artifactId>
+        </dependency>
+        <dependency>
             <groupId>javax.annotation</groupId>
             <artifactId>javax.annotation-api</artifactId>
         </dependency>
@@ -60,12 +64,32 @@
             <artifactId>javax.inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>no.finn.unleash</groupId>
             <artifactId>unleash-client-java</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-auth</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-auth-shared</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
@@ -88,12 +112,24 @@
             <artifactId>che-core-commons-inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-lang</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.che.infrastructure</groupId>
             <artifactId>infrastructure-kubernetes</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.infrastructure</groupId>
             <artifactId>infrastructure-openshift</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.multiuser</groupId>
+            <artifactId>che-multiuser-keycloak-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.multiuser</groupId>
+            <artifactId>che-multiuser-keycloak-shared</artifactId>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8AuthServiceClient.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8AuthServiceClient.java
@@ -1,9 +1,10 @@
 /*
  * Copyright (c) 2016-2018 Red Hat, Inc.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
  *   Red Hat, Inc. - initial API and implementation
@@ -14,6 +15,7 @@ import com.google.common.io.CharStreams;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.inject.name.Named;
 import io.jsonwebtoken.Jwt;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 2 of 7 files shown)


amisevsk:
There are a few issues I would like input on:
1. How much should we decouple fabric8 auth from upstream Che's multiuser keycloak functionality? A significant portion of upstream is incompatible with fabric8-auth it seems.
2. I'm not entirely sure my approach is ideal, as it kind of overrides upstream classes to pretty much replace their functionality entirely. Also classes like `KeycloakSettings` are reused even though the fit isn't ideal.
3. I'd like opinions on how to proceed with getting the correct properties into che (e.g. the account linking URL) -- we can add a property for the linking URL but then we should probably separate out all of the fabric8 auth endpoint information (this is related to (1))
4. This implementation leaves some things undone, e.g. the `/api/token/github` Che endpoint still returns a 500 error due to not being able to process the auth server's response (content type is `application/vnd.externaltoken+json`, which is rejected by Che). I'm not entirely sure how/when this endpoint is used.

rhopp:
[test]

plugins/fabric8-multi-tenant-manager/pom.xml
```diff
@@ -86,6 +116,11 @@
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-commons-inject</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-lang</artifactId>
+            <version>6.7.1</version>
```
ibuziuk: @amisevsk am I correct that in final PR it is not planned to hardcode the dependency versions, but rather use the one provided from parent ?

amisevsk: Yeah sorry, I quickly added the dependencies to fix the enforce phase of the build to get comments sooner. I'll fix this.

amisevsk: Fixed.

rhopp:
It seems there is something fishy with MavenTestCase. I've created issue for us to investigate&fix that (https://github.com/redhat-developer/che-functional-tests/issues/344). I don't think this test failure is connected to this PR, so feel free to merge, if you think this PR is OK.

ibuziuk:
good job. I do not think it would be easy, if possible, to fully decouple fabric8-auth flow from upstream multiuser keycloak for github account linking and overriding upstream functionality with fabric8-auth linking logic seems to be the only solution. 

> we can add a property for the linking URL but then we should probably separate out all of the fabric8 auth endpoint information (this is related to (1))

However, probably it would be better to have a separate properties in place for auth endpoint linking to make the code more clear and easy to maintain. I believe it is possible to reuse https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.config.yaml#L31 and have subpath harcoded in the code. 


> This implementation leaves some things undone, e.g. the /api/token/github Che endpoint still returns a 500 error due to not being able to process the auth server's response (content type is application/vnd.externaltoken+json, which is rejected by Che). I'm not entirely sure how/when this endpoint is used.

I have a feeling that this code is not used anymore. It would be great to investigate as part of the - https://github.com/redhat-developer/rh-che/issues/814

@amisevsk what I'm a bit worried about and not fully understand is how github account linking would work for the `brand-new` users registered via `che.openshift.io` and who never visited  openshift.io  profile page ?

amisevsk:
@ibuziuk I'll work on adding props for the auth stuff; I think something like `che.fabric8.auth.base` plus `che.fabric8.auth.github_token`, `che.fabric8.auth.github_link` might work and be clean (we're currently using the `che.keycloak.github.endpoint` which is a little unclear). The reason I didn't do it initially is that I think if we take a property for the linking URL we should also not reuse the upstream property for the github endpoint, as all of the auth config should be in one place (upstream props or rhche props).

Regarding brand new users, I'm not sure of the flow for registering via `che.openshift.io` directly -- I imagine they will still have openshift.io accounts created and their token would still be managed by fabric8-auth. It's worth testing, however.

slemeur:
Have you validated that it works when the user is in the IDE and opens the PR Panel too? 

The use case:
(New User)
- User is starting with the factory: https://che.openshift.io/f?id=factorymtyoro1y0qt8tq2j
- He is doing some changes in the code
- He goes to the PR panel and tries to make the PR from there.

ibuziuk:
@amisevsk I would propose not to create extra properties and just reuse - https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.config.yaml#L31 (subpath would be added in the code e.g - `/api/token/link?for=openshift`) less properties we have the better.
In terms of the flow that Stevan described I believe user will need to set committer name / email at very least before providing PR - this could also be automated but it would be better to do it in a separate PR (retrieve & set committer username / email during github account linking) - https://github.com/redhat-developer/che-starter/blob/master/src/main/java/io/fabric8/che/starter/client/github/GitHubClient.java#L87

amisevsk:
@slemeur I've tested it with an already created user and was able to make PRs / commits on github without issue, but I will test it further. One issue with this implementation is that the linking is not automatic -- the `link github` button still appears even if a user's account is linked on the osio side. For opening PRs from the IDE, I suspect that the user may have to link their github in the preferences page, but I'm not fully sure. @ibuziuk is also correct, users will have to fill out the commiter name and email before making any commits, but that seems out of the scope of the current issue. I believe committer details would have to be obtained from the auth endpoint separately.

@ibuziuk That still leaves auth endpoints hard-coded, and is not altogether different from the implementation we have now -- it would just be moving from `che.keycloak.token.endpoint` to `che.fabric8.auth.endpoint`. I'm not too sure the best approach here, as the `che.fabric8.auth.*` properties are not picked up by `KeycloakSettings`, so it might cause confusion down the line. In general I think an important issue in the near future is to make the properties we use more uniform -- we currently have e.g.
```
  che-fabric8-user-service-endpoint: "https://api.prod-preview.openshift.io/api/user/services"
  che-fabric8-auth-endpoint: "https://auth.prod-preview.openshift.io"
  che-keycloak-auth-server-url: "NULL"
  che-keycloak-realm: "NULL"
  che-keycloak-client-id: "740650a2-9c44-4db5-b067-a3d1b2cd2d01"
  che-keycloak-oidc-provider: "https://auth.prod-preview.openshift.io/api"
  che-keycloak-js-adapter-url: "/_app/keycloak/RhCheKeycloak.js"
  che-keycloak-use-nonce: "false"
```
This is pretty confusing, as it's not clear which properties are important / which are used at all -- as an example, `che-keycloak-oidc-provider` caused me a lot of issues in working on this, as the docs suggest it can be used if `che-keycloak-auth-server-url` is `null`, but this is not the case. For now I think your approach may be the cleanest, so I'll go with that.

amisevsk:
@ibuziuk To add to the issue I mentioned above, this is the current state of `che-fabric8-auth-endpoint`: 
```
$ grep -rnwi 'che.fabric8.auth.endpoint' .
openshift/minishift-addons/rhche/templates/rh-che.app.yaml:213:          - name: CHE_FABRIC8_AUTH_ENDPOINT
openshift/minishift-addons/rhche/templates/rh-che.app.yaml:214:            value: ${CHE_FABRIC8_AUTH_ENDPOINT}
openshift/minishift-addons/rhche/templates/rh-che.app.yaml:323:- name: CHE_FABRIC8_AUTH_ENDPOINT
openshift/rh-che.app.yaml:154:          - name: CHE_FABRIC8_AUTH_ENDPOINT
openshift/rh-che.app.yaml:157:                key: che-fabric8-auth-endpoint
openshift/rh-che.config.yaml:31:  che-fabric8-auth-endpoint: "https://auth.prod-preview.openshift.io"
```
Setting it does nothing at the moment, and the value is not picked up anywhere.

amisevsk:
I've pushed the changes to use the `che.fabric8.auth.endpoint` property, but haven't been able to test it due to issues on the dev cluster.

ibuziuk:
@amisevsk approved. We probably need to created a separate issue for clean-up of obsolete not used properties, but this is out of the scope of the issue.

ibuziuk:
[test]

amisevsk:
Rebased onto current master, should fix build. Still need to test it more thoroughly.

(Forgot about the approve dismissal)

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8AuthServiceClient.java
```diff
@@ -0,0 +1,206 @@
...
+   * Return account linking url for user. Note that this does not return the fabric8 auth linking
+   * URL, but instead makes a request to that url and returns the redirect obtained from there. This
+   * is because requests to fabric8 auth need the Authorization header, and a user's browser will
+   * not include this header by default, so returning the direct URL does not work.
+   */
+  @Override
+  public String getAccountLinkingURL(
+      @SuppressWarnings("rawtypes") Jwt token, String oauthProvider, String redirectAfterLogin) {
+    String linkingEndpoint =
+        UriBuilder.fromUri(githubLinkEndpoint)
```
davidfestal: Don't you think it would be better to get the `tokenEndpoint` and `linkEndpoint` from `fabric8.*` properties whose name would derive from the `oauthProvider` name ? That would make it much more generic instead of having hard-wired for the GitHub provider ?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8DelegatedOAuthAPI.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+    return Response.temporaryRedirect(URI.create(accountLinkingUrl)).build();
+  }
+
+  @Override
+  public OAuthToken getToken(String oauthProvider)
+      throws ForbiddenException, BadRequestException, NotFoundException, ServerException,
+          UnauthorizedException {
+    try {
+      GithubToken token = authServiceClient.getGithubToken();
```
davidfestal: Don't you think it would be better to get the `tokenEndpoint` and `linkEndpoint` from `fabric8.*` properties whose name would derive from the `oauthProvider` name ? That would make it much more generic instead of having hard-wired for the GitHub provider ?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8AuthServiceClient.java
ibuziuk: @davidfestal do you mean adding extra properties to cm for:

- https://auth.openshift.io/api/token/link?for=openshift 
- https://auth.openshift.io/api/token?for=openshift

davidfestal: yes, more precisely, I mean something like:
```
che.fabric8.oauth.provider.github.endpoints.link=api/token/link?for=https://github.com
che.fabric8.oauth.provider.github.endpoints.token=api/token?for=https://github.com
```
And of course it would be available for Openshift as well.

davidfestal: see this code: https://github.com/redhat-developer/rh-che/blob/master/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/OpenshiftUserTokenProvider.java#L74
seems that this property is in fact already available as
```
 @Nullable @Named(OIDC_PROVIDER_SETTING) String oidcProvider
```
The only difference is that it already contains the `/api` suffix

ibuziuk:
[test]

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8AuthServiceClient.java
ibuziuk: @amisevsk @davidfestal I would propose to merge the PR in the current state and update / cleanup properties in a separate issue / PR. I believe there is smth. we could definitely cleanup according to the - https://github.com/redhat-developer/rh-che/pull/810#issuecomment-415852628

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8DelegatedOAuthAPI.java
amisevsk: I agree that we should have properties for the endpoints, but as we discussed above, I think cleaning up the properties used in rh-che should happen first. There are already 3-4 properties that do similar things but don't seem to be used. Before this PR, the `che.fabric8.auth.endpoint` property wasn't used anywhere at all, and setting it did nothing. 

E.g. the oidc provider property is used direclty in `OpenShiftUserTokenProvider`, but is ignored in [`KeycloakServiceClient`](https://github.com/eclipse/che/blob/a59db0a176a82387c78b76201d42050960acced6/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakServiceClient.java#L103). The documentation around the different properties and how they are used is pretty unclear, so we should take a separate issue to clean up properties and make everything consistent.

davidfestal: It is used / checked in the [KeycloakSettings](https://github.com/eclipse/che/blob/a59db0a176a82387c78b76201d42050960acced6/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakSettings.java) class though. But it seems that `KeycloakServiceClient` was implemented after the OIDC provider support change, but without taking it into account, which explains why the `OIDC_PROVIDER_SETTING` property wasn't used.

davidfestal:
> as an example, che-keycloak-oidc-provider caused me a lot of issues in working on this, as the docs suggest it can be used if che-keycloak-auth-server-url is null, but this is not the case

@amisevsk what do you mean exactly ? to summarize, `che-keycloak-auth-server-url` and `che-keycloak-oidc-provider` are somehow mutually-exclusive: even if we don't enforce it (having both properties might be accepted, which we might want to fix), at least we check that at least one of both properties is non-null : see [here](https://github.com/eclipse/che/blob/a59db0a176a82387c78b76201d42050960acced6/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakSettings.java#L67).

ibuziuk:
[test]

ibuziuk:
@amisevsk @davidfestal going to merge this PR and we could continue the discussion around properties in a separate issue - https://github.com/redhat-developer/rh-che/issues/836

PR #813, opened 2018-08-21
FT - workspace deletion fix

Katka92:
### What does this PR do?
Test will not fail if workspace is not found while sending deletion request for che-starter. Only warning is shown.

che-functional-tests#324

Tests were run locally against production.


rhopp:
@Katka92 Build failed - `Non complying file: /root/payload/functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java`
Please fix ;-)

PR #831, opened 2018-08-27
Mechanism for using created workspace added.

Katka92:
### What does this PR do?
In upstream tests, the possibility of using workspace which was created outside tests was missing. New workspace class ProvidedWorkspace as created and new methods to provide it were added. 

https://github.com/redhat-developer/che-functional-tests/issues/343

Locally against production.

### DIFF
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
@@ -49,7 +49,6 @@ public class RhCheSeleniumSuiteModule extends AbstractModule {
   public ProvidedWorkspace getProvidedWorkspace(
       RhCheTestWorkspaceProvider workspaceProvider,
       DefaultTestUser testUser,
-      @Named("workspace.default_memory_gb") int defaultMemoryGb,
       @Named("sys.workspaceName") String givenWorkspaceName)
       throws Exception {
     ProvidedWorkspace ws = workspaceProvider.findWorkspace(testUser, givenWorkspaceName);
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
@@ -17,6 +17,8 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
 import com.google.inject.name.Named;
+import com.google.inject.name.Names;
+import com.google.inject.util.Providers;
 import com.redhat.che.selenium.core.client.RhCheTestWorkspaceServiceClient;
 import com.redhat.che.selenium.core.workspace.ProvidedWorkspace;
 import com.redhat.che.selenium.core.workspace.RhCheTestWorkspaceProvider;
@@ -27,6 +29,7 @@ import org.eclipse.che.selenium.core.configuration.TestConfiguration;
 import org.eclipse.che.selenium.core.provider.DefaultTestUserProvider;
 import org.eclipse.che.selenium.core.user.DefaultTestUser;
 import org.eclipse.che.selenium.core.user.MultiUserCheDefaultTestUserProvider;
+import org.eclipse.che.selenium.core.workspace.TestWorkspace;
 import org.eclipse.che.selenium.core.workspace.TestWorkspaceProvider;
 
 public class RhCheSeleniumSuiteModule extends AbstractModule {
@@ -43,14 +46,35 @@ public class RhCheSeleniumSuiteModule extends AbstractModule {
             .build(TestWorkspaceServiceClientFactory.class));
     bind(TestWorkspaceServiceClient.class).to(RhCheTestWorkspaceServiceClient.class);
     bind(TestWorkspaceProvider.class).to(RhCheTestWorkspaceProvider.class).asEagerSingleton();
+    if (config.getMap().get("che.workspaceName") == null) {
+      bind(String.class)
+          .annotatedWith(Names.named("che.workspaceName"))
+          .toProvider(Providers.<String>of(null));
+    }
+  }
+
+  @Provides
+  public TestWorkspace getWorkspace(
+      TestWorkspaceProvider workspaceProvider,
+      DefaultTestUser testUser,
+      @Named("workspace.default_memory_gb") int defaultMemoryGb)
+      throws Exception {
+    TestWorkspace ws =
+        workspaceProvider.createWorkspace(testUser, defaultMemoryGb, "default.json", true);
+    ws.await();
+    return ws;
   }
 
   @Provides
   public ProvidedWorkspace getProvidedWorkspace(
       RhCheTestWorkspaceProvider workspaceProvider,
       DefaultTestUser testUser,
-      @Named("sys.workspaceName") String givenWorkspaceName)
+      @Named("che.workspaceName") String givenWorkspaceName)
       throws Exception {
+    if (givenWorkspaceName == null) {
+      throw new RuntimeException(
+          "Variable `che.workspaceName` must be set to use ProvidedWorkspace");
+    }
     ProvidedWorkspace ws = workspaceProvider.findWorkspace(testUser, givenWorkspaceName);
     ws.await();
     return ws;
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
@@ -14,8 +14,11 @@ package com.redhat.che.selenium.core;
 import static com.google.inject.name.Names.named;
 
 import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
+import com.google.inject.name.Named;
 import com.redhat.che.selenium.core.client.RhCheTestWorkspaceServiceClient;
+import com.redhat.che.selenium.core.workspace.ProvidedWorkspace;
 import com.redhat.che.selenium.core.workspace.RhCheTestWorkspaceProvider;
 import org.eclipse.che.selenium.core.client.TestWorkspaceServiceClient;
 import org.eclipse.che.selenium.core.client.TestWorkspaceServiceClientFactory;
@@ -40,5 +43,17 @@ public class RhCheSeleniumSuiteModule extends AbstractModule {
             .build(TestWorkspaceServiceClientFactory.class));
     bind(TestWorkspaceServiceClient.class).to(RhCheTestWorkspaceServiceClient.class);
     bind(TestWorkspaceProvider.class).to(RhCheTestWorkspaceProvider.class).asEagerSingleton();
+  }
+
+  @Provides
+  public ProvidedWorkspace getProvidedWorkspace(
+      RhCheTestWorkspaceProvider workspaceProvider,
+      DefaultTestUser testUser,
+      @Named("workspace.default_memory_gb") int defaultMemoryGb,
+      @Named("sys.workspaceName") String givenWorkspaceName)
+      throws Exception {
+    ProvidedWorkspace ws = workspaceProvider.findWorkspace(testUser, givenWorkspaceName);
+    ws.await();
+    return ws;
   }
 }
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
@@ -103,7 +103,6 @@ public class RhCheTestWorkspaceServiceClient extends AbstractTestWorkspaceServic
       waitStatus(workspaceName, owner.getName(), WorkspaceStatus.RUNNING);
       LOG.info("Workspace " + workspaceName + "is running.");
     } catch (Exception e) {
-
       LOG.error("Failed to start workspace \"" + workspaceName + "\": " + e.getMessage(), e);
       throw e;
     }
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
@@ -94,6 +94,10 @@ public class RhCheTestWorkspaceServiceClient extends AbstractTestWorkspaceServic
   @Override
   public void start(String workspaceId, String workspaceName, TestUser workspaceOwner)
       throws Exception {
+    if (getStatus(workspaceId).equals(WorkspaceStatus.RUNNING)) {
+      LOG.info("Workspace is running - no need to start it.");
+      return;
+    }
     try {
       this.cheStarterWrapper.startWorkspace(workspaceId, workspaceName, token);
       waitStatus(workspaceName, owner.getName(), WorkspaceStatus.RUNNING);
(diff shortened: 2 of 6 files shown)


functional-tests/src/test/java/com/redhat/che/functional/tests/EETest.java
```diff
@@ -0,0 +1,34 @@
...
+public class EETest {
+
+  @Inject private ProvidedWorkspace worksapce;
+
+  @Test
+  public void mytest() {
+    try {
+      worksapce.getName();
+    } catch (ExecutionException | InterruptedException e) {
+      // TODO Auto-generated catch block
```
ScrewTSW: hmm :D
I would probably edit that catch block to at least log some kind of a message and the test should probably fail.

Katka92: fixed

Katka92:
New issue found - if request PATCH failed while starting workspace, the method didn't fail. Should be fixed as a part of #833.

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
```diff
@@ -41,4 +44,16 @@ protected void configure() {
     bind(TestWorkspaceServiceClient.class).to(RhCheTestWorkspaceServiceClient.class);
     bind(TestWorkspaceProvider.class).to(RhCheTestWorkspaceProvider.class).asEagerSingleton();
   }
+
+  @Provides
+  public ProvidedWorkspace getProvidedWorkspace(
+      RhCheTestWorkspaceProvider workspaceProvider,
+      DefaultTestUser testUser,
+      @Named("workspace.default_memory_gb") int defaultMemoryGb,
```
rhopp: Is this variable ever used? Do we need it?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
```diff
@@ -0,0 +1,106 @@
...
+                    "Workspace name='{}' id='{}' started in {} sec.",
+                    workspaceName.get(),
+                    ws.getId(),
+                    (System.currentTimeMillis() - start) / 1000);
+              } catch (Exception e) {
+                String errorMessage =
+                    format("Workspace name='%s' start failed.", this.workspaceName.get());
+                LOG.error(errorMessage, e);
+
+                if (e instanceof IllegalStateException) {
```
rhopp: Wouldn't it be nicer to use multiple `catch` instead of one `catch` and then using `instanceof`?

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
```diff
@@ -87,17 +97,41 @@ public Workspace createWorkspace(
...
+        if (expectedStatus == WorkspaceStatus.RUNNING && status == WorkspaceStatus.STOPPED) {
+          break;
+        }
+        WaitUtils.sleepQuietly(1);
+      }
+    }
+
+    throw new IllegalStateException(
+        format(
+            "Workspace %s, status=%s, expected status=%s", workspaceName, status, expectedStatus));
```
ScrewTSW: is it possible to use format `IllegalStateException("Workspace {}, status={}, expected status={}", worksapceName, status, expectedStatus);` ? is the `format()` necessary?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/RhCheTestWorkspaceProvider.java
```diff
@@ -40,7 +40,7 @@ protected RhCheTestWorkspaceProvider(
       @Named("workspace.default_memory_gb") int defaultMemoryGb,
       DefaultTestUser defaultUser,
       WorkspaceDtoDeserializer workspaceDtoDeserializer,
-      TestWorkspaceServiceClient testWorkspaceServiceClient,
+      RhCheTestWorkspaceServiceClient testWorkspaceServiceClient,
```
ScrewTSW: is this correct?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.LoggerFactory;
+
+/** @author Katerina Foniok */
+public class ProvidedWorkspace implements TestWorkspace {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ProvidedWorkspace.class);
+
+  private CompletableFuture<Void> future;
+  private AtomicReference<String> id;
+  private AtomicReference<String> workspaceName;
```
rhopp: Is there a reason why we have this as AtomicReferences, instead of pure String?

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
```diff
@@ -87,17 +97,41 @@ public Workspace createWorkspace(
...
+  @Override
+  public void waitStatus(String workspaceName, String userName, WorkspaceStatus expectedStatus)
+      throws Exception {
+    WorkspaceStatus status = null;
+    for (int i = 0; i < 600; i++) {
+      status = getByName(workspaceName, userName).getStatus();
+      if (status == expectedStatus) {
+        return;
+      } else {
+        if (expectedStatus == WorkspaceStatus.RUNNING && status == WorkspaceStatus.STOPPED) {
```
rhopp: This `if` is the only difference between this method and the one from parent, right? What's the purpose of this?

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
```diff
@@ -40,5 +46,33 @@ protected void configure() {
             .build(TestWorkspaceServiceClientFactory.class));
     bind(TestWorkspaceServiceClient.class).to(RhCheTestWorkspaceServiceClient.class);
     bind(TestWorkspaceProvider.class).to(RhCheTestWorkspaceProvider.class).asEagerSingleton();
+    if (config.getMap().get("sys.workspaceName") == null) {
+      bind(String.class)
+          .annotatedWith(Names.named("sys.workspaceName"))
+          .toProvider(Providers.<String>of(null));
```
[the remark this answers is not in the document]
Katka92: Maybe I should've added comment here - gonna do that. The problem is that Guice has a problem with "null" instances - if we want to bind "default" workspace (not provided), it doesn't receive "sys.workspaceName" parameter. But in the begining of run, Guice checks that he knows all of variables to be bind - event it isn't needed in this particular run. Then guice finds out that this parameter "sys.workspaceName" is missing and failes. It just don't want instance to be null. But if I use provider, which provides null, it seems ok to Guice and accept it as "developer really want it null". As far as I understand, it goes like this.
So fist it tries find it in parameters in bindConstant() method. If it is not specified, it remains null and therefore we need to set it to null via provider.
I'm not aware of any better approach unfortunately. Hope I explained it sufficiently. :)

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
Katka92: I took it from parent method which I'm overriding. Have no strong feelings about that. Will change it.

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
```diff
@@ -40,5 +46,33 @@ protected void configure() {
...
+    return ws;
+  }
+
+  @Provides
+  public ProvidedWorkspace getProvidedWorkspace(
+      RhCheTestWorkspaceProvider workspaceProvider,
+      DefaultTestUser testUser,
+      @Named("sys.workspaceName") String givenWorkspaceName)
+      throws Exception {
+    ProvidedWorkspace ws = workspaceProvider.findWorkspace(testUser, givenWorkspaceName);
```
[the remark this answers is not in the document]
Katka92: Sure, will change it.

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
```diff
@@ -11,10 +11,19 @@
  */
 package com.redhat.che.selenium.core.client;
 
+import static java.lang.String.format;
```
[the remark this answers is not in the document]
Katka92: Because of @ScrewTSW review this import will not be used anymore.

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
Katka92: Yes, this is the only difference. The reason is that the default timeout for starting workspace is set by upstream to 10 mins. As far as I know there is not any workflow when workspace can became "STOPPED" when starting. This ```if``` allows it to be detected much earlier, so time of tests will be shorter and it will be more friendly when testing locally. Next reason why I decided to add it is that we added it in our old tests too.
On the other hand, this behaviour is not needed, so if you prefer, I can remove it.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
Katka92: Not sure - taken from upstream project setting. If you want me to remove it, it needs to be done in RhCheTestWorkspaceImpl too.

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
```diff
@@ -40,5 +46,33 @@ protected void configure() {
...
+        workspaceProvider.createWorkspace(testUser, defaultMemoryGb, "default.json", true);
+    ws.await();
+    return ws;
+  }
+
+  @Provides
+  public ProvidedWorkspace getProvidedWorkspace(
+      RhCheTestWorkspaceProvider workspaceProvider,
+      DefaultTestUser testUser,
+      @Named("sys.workspaceName") String givenWorkspaceName)
```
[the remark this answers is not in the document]
Katka92: Good idea, will change it. :+1:

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
Katka92: Just investigate a bit - if I use {} I still need to use String.format method, don't I? The only thing that I can do is to concatenate it directly with ```+```.

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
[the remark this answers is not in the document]
Katka92: So to be sure, you want to change it this way?
screenshot from 2018-08-30 12-02-41

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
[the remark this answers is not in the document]
ScrewTSW: :+1:

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
ScrewTSW: dismissed - leave it in original format :)

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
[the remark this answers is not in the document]
ScrewTSW: dismissed - remains in original format

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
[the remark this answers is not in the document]
rhopp: I was thinking more like this:
```@Provides
  public ProvidedWorkspace getProvidedWorkspace(
      RhCheTestWorkspaceProvider workspaceProvider,
      DefaultTestUser testUser,
      @Named("sys.workspaceName") String givenWorkspaceName)
      throws Exception {
    if (givenWorkspaceName == null){
        throw new RuntimeException("Variable `che.workspaceName` must be set to use ProvidedWorkspace");
    }
    ProvidedWorkspace ws = workspaceProvider.findWorkspace(testUser, givenWorkspaceName);
    ws.await();
    return ws;
  }
```

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
rhopp: I would prefer removing this re-implementation. If we find ourself in situation, where we need that, I would prefer change upstream (adding parameter, where the timeout can be changed, or changing the logic there...)

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
rhopp: @Katka92 I believe you are correct, `IllegalStateException` does not have any constructor with varargs AFAIK

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
rhopp: @Katka92 Upstream equivalent of this class is `org.eclipse.che.selenium.core.workspace.TestWorkspaceImpl` which has only `id` as AtomicReference.  Because `id` is the only field, which is written into. Other fields are only read (after the initial injection) and that may be the reason...

functional-tests/src/main/java/com/redhat/che/selenium/core/RhCheSeleniumSuiteModule.java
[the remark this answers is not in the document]
Katka92: I see, thanks!

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
ScrewTSW: @rhopp agreed.

Katka92: :+1:

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
Katka92: :+1: Refactored.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/ProvidedWorkspace.java
```diff
@@ -0,0 +1,98 @@
...
+  private RhCheTestWorkspaceServiceClient workspaceServiceClient;
+
+  public ProvidedWorkspace(
+      TestUser owner,
+      RhCheTestWorkspaceServiceClient testWorkspaceServiceClient,
+      String givenWorkspaceName) {
+    this.id = new AtomicReference<>();
+    this.owner= owner;
+    this.workspaceServiceClient = testWorkspaceServiceClient;
+    if (this.workspaceServiceClient == null) {
```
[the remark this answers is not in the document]
Katka92: Hm, I think so. Removed. Hope my PR is perfect now ;)

Katka92:
Failed on maven test case - known issue. @rhopp @ScrewTSW can I merge?

PR #833, opened 2018-08-27
153 refactor functional tests che starter wrapper

ScrewTSW:
### What does this PR do?
Cleans up old code, preparing new RestClient class similar as used in our che-functional-tests
Some other minor changes.

https://github.com/redhat-developer/che-functional-tests/issues/287

manual testing and debugging.

### DIFF
--- a/functional-tests/pom.xml
+++ b/functional-tests/pom.xml
@@ -39,32 +39,39 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>com.jayway.jsonpath</groupId>
-                <artifactId>json-path</artifactId>
-                <version>2.2.0</version>
+                <groupId>javax.ws.rs</groupId>
+                <artifactId>javax.ws.rs-api</artifactId>
+                <version>2.0</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.che.core</groupId>
+                <artifactId>che-core-api-dto</artifactId>
+                <version>${che.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.eclipse.che.selenium</groupId>
                 <artifactId>che-selenium-core</artifactId>
-                <!-- <version>${che.version}</version> -->
-                <version>6.9.0</version>
+                <version>${che.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.eclipse.che.selenium</groupId>
                 <artifactId>che-selenium-test</artifactId>
-                <!-- <version>${che.version}</version> -->
-                <version>6.9.0</version>
+                <version>${che.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.eclipse.che.selenium</groupId>
                 <artifactId>che-selenium-test</artifactId>
-                <version>6.9.0</version>
+                <version>${che.version}</version>
                 <type>test-jar</type>
                 <scope>test</scope>
             </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
@@ -76,10 +83,6 @@
         <dependency>
             <groupId>com.google.inject.extensions</groupId>
             <artifactId>guice-assistedinject</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.jayway.jsonpath</groupId>
-            <artifactId>json-path</artifactId>
         </dependency>
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
@@ -94,8 +97,16 @@
             <artifactId>javax.inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
@@ -11,20 +11,29 @@
  */
 package com.redhat.che.selenium.core.workspace;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonParser;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
-import com.jayway.jsonpath.Configuration;
-import com.jayway.jsonpath.JsonPath;
+import com.redhat.che.selenium.core.model.request.CheServerStartWorkspaceRequest;
+import com.redhat.che.selenium.core.model.request.CheStarterCreateWorkspaceRequest;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.util.stream.Collectors;
+import java.util.HashMap;
+import javax.ws.rs.HttpMethod;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
-import okhttp3.Request.Builder;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import org.eclipse.che.api.core.ApiException;
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.rest.HttpJsonRequestFactory;
+import org.eclipse.che.api.core.rest.HttpJsonResponse;
+import org.eclipse.che.selenium.core.provider.TestApiEndpointUrlProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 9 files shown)


functional-tests/src/test/java/com/redhat/che/functional/tests/TestTestClass.java
```diff
@@ -25,18 +46,62 @@
...
+    // Create workspace
+    String workspaceName;
+    try {
+      workspaceName = cheStarterWrapper.createWorkspace(createWorkspaceRequestJson, token);
+    } catch (Exception e) {
+      LOG.error("Workspace creation failed:"+e.getMessage(),e);
+      throw e;
+    }
+
+//    // Start workspace
```
ibuziuk: why do we need commented code here ?

functional-tests/src/test/java/com/redhat/che/functional/tests/TestTestClass.java
```diff
@@ -25,18 +46,62 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(TestTestClass.class);
 
-  @Inject private TestWorkspace testWorkspace;
-  @Inject private ProjectExplorer projectExplorer;
-  @Inject private Menu menu;
+//  @Inject private TestWorkspace testWorkspace;
```
ibuziuk: IMO we should remove all commented code

ScrewTSW: yes, I will clean it up when I squash the commits.

functional-tests/src/test/java/com/redhat/che/functional/tests/TestTestClass.java
ScrewTSW: I want to keep it as a TODO: for the future rebase. I can put it into a separate branch and clean it up from this PR.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -76,101 +81,74 @@ public String createWorkspace(String pathToJson, String token) throws Exception
...
+  public boolean deleteWorkspace(String workspaceName, String token) throws IOException {
+    String relativePath = "/workspace/" + workspaceName;
+    HashMap<String, String> queryParameters = new HashMap<>();
+    queryParameters.put("masterUrl", this.host);
+    queryParameters.put("namespace", "sth");
+    String authorization = "Bearer " + token;
+    Response response =
+        cheStarterClient.sendRequest(
+            relativePath, HttpMethods.DELETE, null, authorization, queryParameters.entrySet());
+    if (response.isSuccessful()) {
```
Katka92: Shouldn't here be if(!response.isSuccessfull()) ?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -76,101 +81,74 @@ public String createWorkspace(String pathToJson, String token) throws Exception
...
-    RequestBody body = RequestBody.create(null, new byte[0]);
-    Request request = requestBuilder.patch(body).build();
-    try {
-      Response response = client.newCall(request).execute();
+  public void startWorkspace(String WorkspaceID, String workspaceName, String token)
+      throws Exception {
+    String relativePath = "/workspace/" + workspaceName;
+    HashMap<String, String> queryParameters = new HashMap<>();
+    queryParameters.put("masterUrl", this.host);
+    queryParameters.put("namespace", "sth");
```
Katka92: These three lines repeats in each method call. Is there some way how to generalize it?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -76,101 +91,80 @@ public String createWorkspace(String pathToJson, String token) throws Exception
...
-    try {
-      Response response = client.newCall(request).execute();
+  public void startWorkspace(String WorkspaceID, String workspaceName, String token)
+      throws Exception {
+    String relativePath = "/workspace/" + workspaceName;
+    HashMap<String, String> queryParameters = new HashMap<>();
+    queryParameters.put(
+        "masterUrl", this.cheHost); // set by env variables when launching che-starter -- UNUSED
+    queryParameters.put(
+        "namespace", "sth"); // set by env variables when launching che-starter -- UNUSED
```
Katka92: This block is repeated in each method:
 HashMap<String, String> queryParameters = new HashMap<>();
￼     queryParameters.put(
￼         "masterUrl", this.cheHost); // set by env variables when launching che-starter -- UNUSED
￼     queryParameters.put(
￼         "namespace", "sth"); // set by env variables when launching che-starter -- UNUSED

Is there some way how to improve it? E. g. have it as a class variable?

functional-tests/src/main/java/com/redhat/osio/util/RestClient.java
```diff
@@ -0,0 +1,129 @@
...
+    return sendRequest(null, method);
+  }
+
+  public Response sendRequest(String relativePath, HttpMethods method) throws IOException {
+    return sendRequest(relativePath, method, null);
+  }
+
+  public Response sendRequest(String relativePath, HttpMethods method, JsonObject body)
+      throws IOException {
+    return sendRequest(relativePath, method, body, null, null);
```
Katka92: Wouldn't it be better to use this line in methods above too? Instead of sending
(null, method) and (relativePath, method, null) ? There is no logic or setting in these methods so I can't see any reason to go through all of them.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
ScrewTSW: fixed, added into test case.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
ScrewTSW: refactored inot a private method.

functional-tests/src/main/java/com/redhat/osio/util/RestClient.java
ScrewTSW: It clearly shows what is set to null in the specific call. Other values can, but don't have to be null in them. In my opinion it's cleaner like this, but I can change this if requested.

ScrewTSW:
[test]

functional-tests/src/main/java/com/redhat/osio/util/RestClient.java
Katka92: @rhopp what do you think? It seems for me like slowdown - call method with method set which will call next method to set body to null, then next method to set next part to null... Wouldn't it be better to call just (null, method, null, null, null) to avoid this chain of methods (which do only set to null)?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -11,166 +11,209 @@
...
+    Response response = client.newCall(request).execute();
+    if (response.isSuccessful()) {
+      relativePath = "workspace/" + WorkspaceID + "/runtime";
+      request =
+          new Request.Builder()
+              .url(testApiEndpointUrlProvider.get().toString().concat(relativePath))
+              .addHeader("Authorization", authorization)
+              .post(body)
+              .build();
+      response = client.newCall(request).execute();
```
Katka92: Is it needed to use OkHttpClient here? Wouldn't it be better to use jsonRequestFactory? I think OkHttpClient should be used only for PATCH request.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -11,166 +11,209 @@
...
+        LOG.info("Workspace was started successfully. Waiting for state RUNNING.");
+        return;
+      } else {
+        LOG.error("Che server start workspace failed:" + response.body().string());
       }
-    } catch (IOException e) {
-      LOG.error("Workspace start failed : " + e.getMessage(), e);
-      throw e;
+    } else {
+      LOG.error("Che starter start workspace failed:" + response.body().string());
```
Katka92: The more accurate message should be handy. Saying something like "Che starter set environment before worskpace start failed:"?

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -11,166 +11,209 @@
...
-      e.printStackTrace();
+  private String getNameFromResponse(HttpJsonResponse response)
+      throws RuntimeException, JsonParseException {
+    if (response.getResponseCode() == 200) {
+      String responseBody = response.asString();
+      JsonObject workspaceJsonObject = PARSER.parse(responseBody).getAsJsonObject();
+      JsonObject workspaceConfigJsonObject = workspaceJsonObject.get("config").getAsJsonObject();
+      return workspaceConfigJsonObject.get("name").getAsString();
+    } else {
+      String error = "Could not get name from response. Request failed or the response is empty.";
```
Katka92: Is there check that response is empty? Wouldn't it fail in line 199 if it is empty?

ScrewTSW: response can be empty but it still contains a response code. (either 200, 404, 5xx)

functional-tests/pom.xml
```diff
@@ -39,32 +39,44 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>com.jayway.jsonpath</groupId>
-                <artifactId>json-path</artifactId>
-                <version>2.2.0</version>
+                <groupId>com.google.code.gson</groupId>
```
rhopp: Why do we need gson here in dependencyManagement section? Isn't it already specified in parent pom.

ScrewTSW:
[test]

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -133,64 +163,104 @@ public void startWorkspace(String WorkspaceID, String name, String token) throws
...
+    } catch (BadRequestException e) {
+      LOG.error("Workspace failed to start because of a bad request:" + e.getMessage(), e);
+      throw e;
+    } catch (ConflictException e) {
+      LOG.error("Workspace fauled to start because of a conflicting request:" + e.getMessage(), e);
       throw e;
+    } catch (NotFoundException e) {
+      LOG.error("The requested endpoint returned 404:" + e.getMessage(), e);
+      throw e;
+    } catch (UnauthorizedException e) {
```
rhopp: Wouldn't it be nicer to catch `org.eclipse.che.api.core.ApiException`? I don't see much value in catching every exception on it's own just to LOG different message (it seems kind of redundant to me, as the exception already contains information about nature of the failure)

ScrewTSW: will do.

Issue #707, opened 2018-06-11
Disable Organizations on che.openshift.io


l0rd:
After discussing "OSIO permission model as a backend for che permission model on che.openshift.io" [1] during f2f it was decided that currently the 'Organizations'  flow is broken for openshift infrastructure in general e.g.:
- resource limits used in Che are not aligned with openshift quotas in namespaces
- shared team workspace would not always be started in the owner namespace, which essentially breaks the idea of teams

Currently, it was decided to disable organizations on che.openshift.io and update the message to in order to clarify that organizations are currently not supported on `che.openshift.io` + add link to the organizations doc[1]:

![image](https://user-images.githubusercontent.com/1461122/43472743-8c35d960-94ee-11e8-9d19-f93bbd41546a.png)

[1] https://docs.google.com/document/d/1y0wjCKcBhzLFgIR-Qgx3YkqWU0kqy638tSOO4oKa_bE/edit#heading=h.6a6mwnzbufob
[2] https://www.eclipse.org/che/docs/organizations.html

ibuziuk:
also need to investigate possibility of integration of che orgs / teams with osio teams as a part of this issue.

ibuziuk:
Sub tasks:
- test the possibility of workspace sharing (workspace details / tab for sharing)
- also need to investigate possibility of integration of che orgs / teams with osio teams as a part of this issue 
- every user should have a possibility to create org from dashboard

ibuziuk:
- [ ] UD is not part of rh-che build, so we probably need to either re-pack the dashboard or add property in upstream to override the message

amisevsk:
PR #830 enables redirection of `learn more` link to upstream organizations doc.

garagatyi:
We also have share workspace view on UD:
![selection_002](https://user-images.githubusercontent.com/1851863/44645437-49c1b480-a9e0-11e8-850f-13f28caa1478.png)
We will have to disable it too. Not sure whether it would work authomatically after disabling organizations.

ibuziuk:
workspace sharing is still planned to be enabled but with some limitations - https://github.com/redhat-developer/rh-che/issues/798

l0rd:
@ibuziuk if I understand correctly we can close this issue right? In any case #798 is already listed in codenvy.io migration epic (#701)

ibuziuk:
@l0rd I was planning to close it once it would be on prod (currently it is only on prod-preview)

ibuziuk:
[...]

Issue #798, opened 2018-08-02
Add limitations for workspace sharing on che.openshift.io


ibuziuk:
On `che.openshift.io` when workspace is shared the following rules must be applied:
- [ ] Only creators should be able to start/stop workspaces
- [ ] Personal GH token is not available / shared within the workspaces (e.g. users still use their own personal GH tokens when accessing a shared workspace)
- [ ] Workspace creator must be informed via notification message that OSO token will be available to all his collaborators as well as any sensitive information that should be created in the workspace (e.g. private SSH keys)

ibuziuk:
Closing since https://github.com/redhat-developer/rh-che/pull/872 is merged

PR #835, opened 2018-08-28
rh-che #707: Disable organizations on rh-che

amisevsk:
### What does this PR do?
Updates messages given when user navigates to 'Organizations' tab [1] and workspace 'Sharing' tab [2] to explain that these features are unavailable.

#### More detail

Add new module to maven build: fabric8-ide-dashboard-war, that modifies the upstream dashboard build to override files so that we can change upstream in a convenient and simple way.

Modifies the organizations page and share-workspace page to give informative messages about how these features are currently unavailable on osio.

Building the dashboard (which is an angularjs project) is nonstandard, so there is detailed documentation in ./assembly/fabric8-ide-dashboard-war/README.md

#707

Manually on the dev cluster. Image built off this PR is available at `amisevsk/rhche-server:issue-707`

[1] - screenshot from 2018-08-28 14-34-46
[2] - screenshot from 2018-08-28 14-34-59


assembly/fabric8-ide-dashboard-war/src/app/workspaces/share-workspace/share-workspace.html
```diff
@@ -0,0 +1,17 @@
...
+    which is available at https://www.eclipse.org/legal/epl-2.0/
+
+    SPDX-License-Identifier: EPL-2.0
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
+<div class="no-permissions-error" layout="row" layout-align="center center">
+  <span class="material-design icon-ic_error_outline_24px"></span>
```
ibuziuk: @amisevsk workspace sharing should still be enabled but with some limitations - https://github.com/redhat-developer/rh-che/issues/798. So, in this PR I would propose just to disable organizations

amisevsk: Good point. Done.

amisevsk:
@davidfestal I'm not sure if there's a convenient way to do this, while keeping the diff readable -- I will look into it though. Do you think something like using [verbose](https://ant.apache.org/manual/Tasks/copy.html) and logging to a file in the ant task would be sufficient? A plain diff of the two source trees will result in hundreds of lines of `-`, as there are far fewer files in the `./src` directory in general.

amisevsk:
@davidfestal I've added some logging to the build process:
1. A diff is taken between `./src` and `./target/sources/src` and saved to file `source_tree.diff` in the `target` dir
2. Added the verbose flag to the copy command, so that it at least lists how many files are copied.

WDYT?

davidfestal:
> @davidfestal I've added some logging to the build process:
> - A diff is taken between ./src and ./target/sources/src and saved to file source_tree.diff in the target dir
> - Added the verbose flag to the copy command, so that it at least lists how many files are copied.
>
> WDYT?

That seems good to me

Issue #826, opened 2018-08-22
Need to update che to 6.10.0 and update rhche on production


ibuziuk:
6.10.0 has fixes for regressions introduced by the switch from pods to deployments for workspaces. Need to update rh-che to 6.10.0 and promote it to production

PR #838, opened 2018-08-29
[WIP] rh-che Issue #826 - Update upstream Che to 6.10.0

amisevsk:
Update to 6.10 and fix the build. This requires some changes to `ForwardActivityFilter` which could affect analytics; I haven't been able to test this and we should confirm there are no issues there.


davidfestal:
Changes seem correct to me.

You should be able to test your changes with analytics sent to the test woopra instance by adding the 2 following properties in your rh-che config map:
```
analytics.segment_write_key
analytics.woopra_domain
```

PR #852, opened 2018-09-04
FT - workspace test

Katka92:
### What does this PR do?
This PR adds workspace test from old tests. It test if workflow for starting second workspace is not broken.
Two workspaces are created, one is started. Then second is started using PATCH request for preparing environment (stops all other workspaces). When second worksapce is started, the request for starting first workspace is sent - without PATCH request (this should fail).  

I run it locally against production and prod-preview.


Katka92:
[test]

Katka92:
[test]

PR #872, opened 2018-09-11
Add capability to share workspace by factory

amisevsk:
### What does this PR do?
Replace the share tab on a workspace page with a page explaining that sharing is only available via factories at the moment, and give the user an option to create a factory for the current workspace from within the page.

https://github.com/redhat-developer/rh-che/issues/798 (via discussion in housekeeping issue around current workspace sharing issues)

Locally and on dev cluster.

share-workspace-2

### DIFF
--- a/assembly/fabric8-ide-dashboard-war/src/app/workspaces/share-workspace/share-workspace.controller.ts
+++ b/assembly/fabric8-ide-dashboard-war/src/app/workspaces/share-workspace/share-workspace.controller.ts
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2016-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+'use strict';
+import { CheAPI } from '../../../components/api/che-api.factory';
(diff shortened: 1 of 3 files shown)


garagatyi:
@amisevsk can we patch the pages instead of replacing them? This would prevent a situation when a page is updated upstream, but changes are not reflected in RhChe.

slemeur:
Sounds good !

amisevsk:
@garagatyi How do you mean? We used to patch files in dashboard but it meant a potentially large burden for every upstream release, as any changes in modified files meant that patches had to be regenerated. It also significantly complicated the build, and made it difficult to work on some changes. I prefer the current approach for cases where we want to replace files entirely.

While the copying over files approach is much more limited, it's also simpler to reason about -- the files in this repo replace files upstream, so upstream changes to e.g. `share-workspace.controller.ts` will be ignored, since we're not using that page at all (and so upstream changes are not needed downstream). The modularity of the typescript build means that we always get upstream changes for anything except for the specific page we're changing -- here, everything except for the contents of the "share" tab will be pulled from upstream, and this will continue to work as long as the files in this PR are not renamed upstream.

The main limitation of this approach is that it becomes difficult to *modify* a file from upstream. If, for example, we wanted to add a controller/directive to `workspaces-config.ts`, we would have to duplicate the file from upstream, add our changes, and then include that in the repo. In this case a patch would be much more suitable as it would allow us to track upstream changes. There's no real upstream code in this PR, so nothing to track; in other words, if this PR was in the form of upstream patches, the patches would consist of `-`ing all lines in the upstream file, and `+`ing the lines here.

garagatyi:
@amisevsk ok, thanks for the explanation!

PR #873, opened 2018-09-11
Improvements to minishift addon yaml template

amisevsk:
### What does this PR do?
Adds some minor improvements / modifications to the multi-user minishift addon for rhche.

- Update image used in minishift addon to use quay.io instead of registry.devshift.net (old version defaulted to a 6.7.1-based deployment)
- Sort templates used to deploy minishift addon and local deployments to more easily compare addon to what is deployed in `./dev-scripts/deploy_custom_rh-che.sh`. This is helpful for e.g. highlighting differences between local and dev cluster deployments (previously it was impossible to tell what differences were, now a `diff` is possible).
- Add serviceaccount and rolebinding to list of deleted openshift object when removing addon, as they don't seem to be deleted by default if only `rhche` addon is removed
- Update default deploy strategy in minishift addon to "Rolling" to match current use; is still configurable as before

I mostly did this to help myself understand the addon and its differences; I don't see merging it as absolutely necessary but thought I'd share.


rhopp:
# :+1:

PR #878, opened 2018-09-12
Add chromedriver to dockerimage

Katka92:
### What does this PR do?
Chromedriver is installed in docker image and run from entrypoint.

locally

### DIFF
--- a/dockerfiles/functional-tests/Dockerfile
+++ b/dockerfiles/functional-tests/Dockerfile
@@ -14,6 +14,8 @@ ENV LANG=en_US.utf8 \
 
 COPY google-chrome.repo /etc/yum.repos.d/google-chrome.repo
 RUN yum update --assumeyes && \
+	yum install --assumeyes epel-release && \
+	yum install --assumeyes chromedriver && \
     yum install --assumeyes google-chrome-stable && \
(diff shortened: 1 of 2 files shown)


dockerfiles/functional-tests/Dockerfile
```diff
@@ -14,6 +14,9 @@ ENV LANG=en_US.utf8 \
 
 COPY google-chrome.repo /etc/yum.repos.d/google-chrome.repo
 RUN yum update --assumeyes && \
+	curl http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm --output epel.rpm && \
+	yum install --assumeyes epel.rpm && \
```
rhopp: I believe it's enough to do `yum install epel.release` instead of downloading epel.rpm from fedoraproject.

ScrewTSW: :-1: tested on `docker.io/centos:7` image:
```
No package epel.release available.
Error: Nothing to do
```

rhopp: Oops.. misspelling on my side. It's `epel-release`

Issue #839, opened 2018-08-30
Unable to start a workspace from a factory url


benoitf:
**Issue problem:** 
Che allows to use factory URLs like `https://che.openshift.io/f?url=<url-of-a-github-repository>`

When using that mode, workspace fails to start.

<details><summary>Details of stack trace</summary><p>

147 lines are hidden...

at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
at com.google.inject.servlet.GuiceServletContextListener.contextInitialized(GuiceServletContextListener.java:47)
at org.everrest.guice.servlet.EverrestGuiceContextListener.getInjector(EverrestGuiceContextListener.java:140)
at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1085)
at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:734)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:752)
at com.google.inject.internal.InternalInjectorCreator$1.call(InternalInjectorCreator.java:199)
at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1092)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at org.everrest.guice.servlet.EverrestGuiceContextListener.contextInitialized(EverrestGuiceContextListener.java:85)
at org.apache.catalina.startup.HostConfig$DeployWar.run(HostConfig.java:1860)
at org.apache.catalina.startup.HostConfig.deployWAR(HostConfig.java:988)
Caused by: org.eclipse.che.api.core.NotFoundException: Factory with id 'undefined' doesn't exist
at org.eclipse.che.api.core.rest.DefaultHttpJsonRequest.doRequest(DefaultHttpJsonRequest.java:269)
at com.redhat.che.plugin.analytics.wsagent.AnalyticsManager.(AnalyticsManager.java:121)
at com.redhat.che.plugin.analytics.wsagent.AnalyticsManager.class(AnalyticsManager.java:72)
at com.redhat.che.plugin.analytics.wsagent.AnalyticsManager.(AnalyticsManager.java:193)
2) Error injecting constructor, java.lang.RuntimeException: Can't get workspace informations for Che analytics
[...]

Issue #815, opened 2018-08-21
Followup and fixes after the  first test phase of end2end registration flow on production


davidfestal:
fix the required things that might have been spotted during the first test phase (for example I detected that for some users, the factory workspace stops -erroneously it seems- immediately after creation at first start, and the user needs to refresh his factory URL to have it started finally, or something like that. Would have to investigate this case a bit more, since currently there is a lack of telemetry information for these users):

Sub-tasks:
- [x] The factories workflow after login doesn't work on Firefox #832

ibuziuk:
@davidfestal will you be able to take a look at https://github.com/redhat-developer/rh-che/issues/832 as part of this issue ?

davidfestal:
Hi @ibuziuk Yes, I'll look into it. That's a bit strange since I also tested on Firefox. I'll contact @scela to have more details on the precise use-case.

PR #884, opened 2018-09-12
Various enhancements to the Che workspace telemetry in the context of issue #815

davidfestal:
### What does this PR do?

Various enhancements to the Che workspace telemetry in the context of issue #815 :
- Problem with the workspaces whose first start failed, and whose
failure was never reported to Woopra:
    - We now add new event properties to state if a workspace was a
failed state before starting  (last start had failed), and the related
error message if available.
- Better management of the source of the workspace:
    - Fixed the unsupported case of url-based factories (see issue
https://github.com/redhat-developer/rh-che/issues/839) so that now the
sent workspace telemetry events have the `factory id` property set to
`undefined` in those case. This will allow identifying the workspace
starts from URL-based factories.
    - Uses the new workspace attribute `osio_spaceId` (whose value is
always `unknown` for now, see PR
https://github.com/redhat-developer/che-starter/pull/311 ) to create a
new telemetry workspace event property (`osio space id`) which is
non-empty when the workspace has been created from the Openshift.io
codebase UI.

- Problem of visits that were sometimes fragmented without any reasons
inside Woopra:
    - Correctly managed the Woopra ping calls *after* each event
submission, and retry pinging if the first ping failed
    - Correctly configured the `EventDispatcher` cache to avoid unexpected
evictions, which in turn would create a new distinct Woopra cookie, thus
opening a new session.

#815

Yes, on Dev cluster


benoitf:
> Fixed the unsupported case of url-based factories (see issue
#839) so that now the
sent workspace telemetry events have the factory id property set to
undefined in those case. This will allow identifying the workspace
starts from URL-based factories.

does it store URL used for these factories ?

davidfestal:
> does it store URL used for these factories ?

@benoitf no, not for now. I've thought of trying to get the git source location of the project in this case, but finally didn't implement it since it would be very robust IMO:
- We could check that there is *only one* project in the workspace, and that this project has the `git` location source type, and in this case consider the project source location as being the factory URL
- But what I don't like in using project information in workspace events, is that these informations are not constant: a user can remove the project later in the workspace, and in the next starts, we will loose the information that this workspace was initially started from the given URL.

What I could probably do is intercept the start event on the wsmaster side and add a `factoryUrl` attribute to the workspace before start in this case.

davidfestal:
> What I could probably do is intercept the start event on the wsmaster side and add a factoryUrl attribute to the workspace before start in this case.

I propose to do this in a next PR.
@benoitf do you agree ?

davidfestal:
test

benoitf:
@davidfestal yes sure, I was just wondering

PR #887, opened 2018-09-13
rh-che #881: Updating rhche to 6.11.0 upstream version

ibuziuk:

### What does this PR do?
Updating rhche to 6.11.0 upstream version

https://github.com/redhat-developer/rh-che/issues/881

Not tested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
-        <version>6.10.0</version>
+        <version>6.11.0</version>
     </parent>
     <groupId>com.redhat.che</groupId>
     <artifactId>fabric8-ide-parent</artifactId>


ibuziuk:
[test]

ibuziuk:
[test]

garagatyi:
@ibuziuk this PR would restart Che master. We also need to add che-plugin-registry env var (+config map entry), so doesn't it make sense to do add them here?

ibuziuk:
@garagatyi I was thinking that we were planning to do it in a separate PR, since this would require creating HK issue with request for adding new entries to `rhche` cm by SD team. Just want to make 6.11.0 available for testing ASAP wdyt ?

garagatyi:
@ibuziuk OK, I just think that it would be better to update prod 1 time instead of 2

ibuziuk:
@garagatyi well, we could update prod once, even though we should not be afraid of doing it twice ;-) for now just want to have it on prod-preview ;-)

Issue #860, opened 2018-09-06
Need to support ephemeral workspace on che.openshift.io


ibuziuk:
Since currently workspace startup on osio could take sufficient amount of time due to pvc mount [1], it was decided to have support of ephemeral workspace with no pvc attached. There are 2 options for supporting this:

- adding new `none` PVC strategy to the existing `common` & `unique`. When this PVC strategy is set no volume would be mounted for workspace
- having an `ephemeral` option in workspace config, which would allow having it as a per-workspace setting (need to decide how this property would be configured for users in che.openshift.io flow)

[1] https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2169

l0rd:
Please consider [emptyDirs](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) and [local-ephemeral-storage](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#local-ephemeral-storage) when implementing that. These may provide better IOPS than the container COW filesystem.

ibuziuk:
It was decided to have a `mount sources` property in order to disable volume mount for workspaces. Using this property would either not do any volume mount or perform mount via `emptyDirs` approach

ibuziuk:
PR have been sent upstream

ibuziuk:
PR is merged. Issue should be closed one we update rh-che to 6.12.0

PR #894, opened 2018-09-18
WIP rh-che #860 Adding new `common-ephemeral` PVC strategy which allows c…

ibuziuk:
…reating ephemeral workspace with `emptyDir` volumes

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) 2016-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package com.redhat.che.multitenant.pvc;
+
+import static org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesObjectUtil.newVolumeMount;
+import static org.eclipse.che.workspace.infrastructure.kubernetes.provision.LogsVolumeMachineProvisioner.LOGS_VOLUME_NAME;
+
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.kubernetes.api.model.VolumeBuilder;
+import java.util.Map;
+import java.util.Map.Entry;
+import javax.inject.Inject;
+import javax.inject.Named;
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.model.workspace.config.Volume;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.workspace.infrastructure.kubernetes.Names;
+import org.eclipse.che.workspace.infrastructure.kubernetes.environment.KubernetesEnvironment;
+import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesNamespaceFactory;
+import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.pvc.CommonPVCStrategy;
+import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.pvc.PVCSubPathHelper;
+
+/**
+ * Extends {@code CommonPVCStrategy} and provides support of the ephemeral workspaces based on the
+ * `emptyDir` volumes. If workspace config contains `mountSources` attribute which is set to true,
+ * than instead of using regular {@code CommonPVCStrategy}, each volume would be represented as an
+ * `emptyDir`. NOTE: no data would be persisted once between ephemeral workspace restarts, since
+ *
+ * @see <a href="https://kubernetes.io/docs/concepts/storage/volumes/#emptydir">EmptyDir</a>
+ * @author Ilya Buziuk
+ */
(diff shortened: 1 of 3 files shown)


plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
```diff
@@ -0,0 +1,164 @@
...
+   * @return true if workspace config contains `mountSources` attribute which is set to true, false
+   *     otherwise
+   */
+  private boolean isEphemeral(String workspaceId) {
+    try {
+      WorkspaceImpl workspace = workspaceManager.getWorkspace(workspaceId);
+      String mountSources = workspace.getConfig().getAttributes().get(MOUNT_SOURCES_ATTRIBUTE);
+      return !Boolean.parseBoolean(mountSources);
+    } catch (NotFoundException | ServerException e) {
+      throw new RuntimeException("Failed to load workspace info" + e.getMessage(), e);
```
garagatyi: I think that any runtime exceptions in this area are discouraged. Please, use InfrastructureException or InternalInfrastructureException (specifically in this case) instead.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Extends {@code CommonPVCStrategy} and provides support of the ephemeral workspaces based on the
+ * `emptyDir` volumes. If workspace config contains `mountSources` attribute which is set to true,
+ * than instead of using regular {@code CommonPVCStrategy}, each volume would be represented as an
+ * `emptyDir`. NOTE: no data would be persisted once between ephemeral workspace restarts, since
+ *
+ * @see <a href="https://kubernetes.io/docs/concepts/storage/volumes/#emptydir">EmptyDir</a>
+ * @author Ilya Buziuk
+ */
+public class CommonPVCStrategyWithEphemeralWorkspacesSupport extends CommonPVCStrategy {
```
garagatyi: This class has pretty complex logic. Consider adding unit tests to avoid a situation when someone else breaking the prod after changes in a class he is not the author of.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
```diff
@@ -0,0 +1,164 @@
...
+  public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
+      throws InfrastructureException {
+    String workspaceId = identity.getWorkspaceId();
+    if (isEphemeral(workspaceId)) {
+      for (Pod pod : k8sEnv.getPods().values()) {
+        final PodSpec podSpec = pod.getSpec();
+        for (Container container : podSpec.getContainers()) {
+          final String machineName = Names.machineName(pod, container);
+          Map<String, Volume> volumes = k8sEnv.getMachines().get(machineName).getVolumes();
+          addMachineVolumes(workspaceId, pod, container, volumes);
```
garagatyi: What will happen with volumes that might come from the WS recipe or provisioned by the infrastructure?

garagatyi: Our infra can add some additional stuff and PVC strategies should respect it. We have just one such case AFAIK at the moment - logs volume. But it is workarounded in existing strategies (including the strategy from this PR). 
Maybe it is OK and infra should not add anything in addition. Not sure. For the time being this code should work I suppose. 
@sleshchenko might want to comment after he is back from a sick leave. 
@ibuziuk consider this thread as thoughts out loud, it doesn't block the PR

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
ibuziuk: sure

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/pvc/CommonPVCStrategyWithEphemeralWorkspacesSupport.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+      }
+    } else {
+      super.provision(k8sEnv, identity);
+    }
+  }
+
+  @Override
+  public void cleanup(String workspaceId) throws InfrastructureException {
+    super.cleanup(workspaceId);
```
ibuziuk: @garagatyi @sleshchenko guys, it is appeared to be that in the strategy `cleanup` method it is not possible to check if the `mountSources` attribute is presented, since workspace is already deleted. Do you have any ideas how ephemeral / non-ephemeral check should be performed in this case, or for the time being it would be ok if `cleanup` would still be called once ephemeral workspace is deleted (the operation is async and in case of ephemeral workspace the `rm-*` pod would just not do any real cleanup operations against pvc)

garagatyi: Here is the code that calls cleanup [link](https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/WorkspacePVCCleaner.java#L59)
Here where it takes workspace ID from [link](https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/WorkspacePVCCleaner.java#L57)
Here you can see that this object contains whole workspace object and this means that we can pass whole workspace instead of just an ID [link](https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/event/WorkspaceRemovedEvent.java#L26)

Hope it helps.

ibuziuk: thanks, but I was not able to override `WorkspacePVCCleaner` in our assembly so far :-(

Issue #901, opened 2018-09-20
Should we consider increasing memory limit for Che server?


amisevsk:
Back when each user's namespace had its own Che server instance, it was necessary to have a conservative limit on the server's memory limit. Now that Che is multi-tenant and the server does not need to be replicated across namespaces, so its memory limit could likely be increased without significant cost.

Currently, the limit is set at 750MB. I recall the server commonly using near 700MB back when each user's namespace hosted an instance, so it's not unreasonable to suspect that this is no longer sufficient if the server is to handle many users at once.

Related: https://github.com/redhat-developer/rh-che/issues/898

garagatyi:
+1 to increase it to avoid OOM but it is very important to set high priority to the investigation of a source of higher memory consumption. If it is thread/connections pool it is OK, but we might find a memory leak or something else.

garagatyi:
@ibuziuk @davidfestal WDYT?

ibuziuk:
PR has been sent - https://github.com/redhat-developer/rh-che/pull/903

garagatyi:
I do not understand something in out JVM config 
Here is upstream Che default config:
-XX:MaxRAMFraction=2 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true -Xms20m
Here is RhCHe prod config
-XX:+UseParallelGC -XX:MinHeapFreeRatio=25 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true -Xms50m -Xmx180m -Dfile.encoding=UTF8
The question is why they are different? Does anybody know?
In particular, here is difference in those configs:
CHE: -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20  -Xms20m            -XX:MaxRAMFraction=2
RHCHE: -XX:MinHeapFreeRatio=25 -XX:MaxHeapFreeRatio=40  -Xms50m                                                                 -Xmx180m -Dfile.encoding=UTF8
[...]

Issue #898, opened 2018-09-20
rh-che on prod has been restarted a few times due to OOMKilled


ibuziuk:
yesterday che has been updated on prod to the latest version 6.11.0. However, during the day rh-che pod was restarted 4 times with `OOMKilled` exception. 
```
Container rhche
State:
Running since Sep 19, 2018 7:37:28 PM
Last State
Terminated at Sep 19, 2018 7:37:27 PM with exit code 137 (OOMKilled)
Ready:
true
Restart Count:
4
```
Note there was a workshop with 30 - 40 participants during this day, and it looks like we never tested rh-che under sufficient load on osio

- [ ] How can we reproduce the OOM? Starting hundreds of simultaneous workspaces is enough?
- [ ] What is the OOM root cause? What classes/methods are using more memory?
- [ ] With 1.5G memory limit what's the maximum number of simultaneous workspaces Che is able to serve?

ibuziuk:
@rhopp @Katka92 @ScrewTSW will it be possible to setup 100 hundred of tests accounts for prod-preview and run stress tests against it ?

rhopp:
@ibuziuk Katka already started working on this. 
@Katka92 Could you create an issue for this (include it in this sprint) and link it here, so we can track the progress?

garagatyi:
@ibuziuk could Che global SA help us with this? We might be able to trick Che and let master create workspaces in different namespaces without having dozens of clients running in parallel. But in any case, we would need these dozens of accounts to have namespaces.

ibuziuk:
@garagatyi well, I actually thought CHE SA would be impl detail and the tests would be run in parallel using OSIO token against rh-che on prod-preview

garagatyi:
Memory consumption by upstream Che with the grow of the load is unclear. I created an issue for that https://github.com/eclipse/che/issues/11326

ibuziuk:
Upstream issue - https://github.com/eclipse/che/issues/5443
Root cause of the problem - https://github.com/eclipse/che/issues/5443#issuecomment-426578251

ibuziuk:
The problem has been workarounded by increasing RAM to 4GB but according to metrics from prod the memory consumption is still tend to keep growing and could result in new OOM restarts:

![image](https://user-images.githubusercontent.com/1461122/47386569-3d6e8500-d70d-11e8-8e2c-fbb53df4b51b.png)

ibuziuk:
[...]

PR #903, opened 2018-09-21
Increasing memory limit to 1.5Gi for rh-che

ibuziuk:
### What does this PR do?
Increasing memory limit to 1.5Gi for rh-che

https://github.com/redhat-developer/rh-che/issues/901

Not tested


ibuziuk:
[test]

pbergene:
As it's only running one replica (?) this is a net increase of 750MB - I think we should be ok, but let me verify this.  When / if we go to multiple replicas we can re-visit into what makes sense.

ibuziuk:
[test]

pbergene:
4-5G free on rh-idev for the entire cluster..  will have to keep an eye on that.  Starting to cut it close on dsaas, but a few nodes worth left.  +1 to this.

ibuziuk:
@pbergene thanks for confirmation

pbergene:
@ibuziuk Thanks for running it by us :+1:

rhopp:
@pbergene The URL reported back to github is not fixed yet (after dev cluster reinstallation). The correct URL is http://rhche-che6-automated.devtools-dev.ext.devshift.net

l0rd:
@ibuziuk we need to answer the following questions:

1. How can we reproduce the OOM? Starting hundreds of simultaneous workspaces is enough?
2. What is the OOM root cause? What classes/methods are using more memory?
3. With 1.5G memory limit what's the maximum number of simultaneous workspaces Che is able to serve?

ibuziuk:
@l0rd I have added the last comment as a subtask for https://github.com/redhat-developer/rh-che/issues/898

l0rd:
great thank you @ibuziuk

PR #906, opened 2018-09-24
rh-che #901: Adding  'MaxRAMFraction=2' and removing '-Xmx180m' from the 'che-server-java-opts'

ibuziuk:
### What does this PR do?

Currently we are using `che-server-java-opts` that have been initially introduced to the single-tenant che-server 
- https://github.com/fabric8-services/fabric8-tenant-che/commit/ef748b6eb39017764fe489de278eb512642c8420
Those limits where ported to multi-tenant without changes - https://github.com/redhat-developer/rh-che/commit/195007157fdacce07f8e7322c562a564edad037f

In the original PR there is a comment:
```
Remove MaxRAMFraction=2 since this would turn out
to be too inflexible for a memory limit of 300MB (i.e. 150MB).
Explicitly set MaxHeapSize=180m instead.
```

Basically, it was planned that for single-tenant che-server 300MB limit would be sufficiently enough and this is definitely not the case for multi-tenant use-case. I would propose to set  `MaxRAMFraction=2`  and remove `-Xmx180m` as the initial step.

https://github.com/redhat-developer/rh-che/issues/901


ibuziuk:
once PR approved I'm going to create HK issue for cm update on prod / prod-preview

ibuziuk:
New cm values have been applied on prod / prod-preview. Merging PR

Issue #899, opened 2018-09-20
Rolling out deployment / restarting rhche breaks workspace timeout in a bad way


amisevsk:
**Issue problem:** 
When a rhche deployment is rolled out (or restarted e.g. in case of OOM -- haven't tested this exact case), the server seems to be unable to stop running workspaces due to timeout. 

Error message in logs:
```
2018-09-20 17:42:03,681[aceSharedPool-0]  [WARN ] [.Fabric8OpenShiftClientFactory 77]   - Current user ID ('0000-00-0000') is different from the user ID ('3f652e0c-1874-4de5-9c62-602634cb5cc2') that started workspace workspacem4yecw20xrdl9ls8 and subject of user that started workspace is not found in subjects cache
2018-09-20 17:42:03,703[aceSharedPool-0]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 485]   - Error occurs on workspace 'amisevsk-preview/wksp-dv95' with id 'workspacem4yecw20xrdl9ls8' stopped by user 'activity-checker'. Error: {}
org.eclipse.che.api.workspace.server.spi.InfrastructureException: Error(s) occurs while cleaning up the namespace. The anonymous subject is used, and won't be able to perform this action The anonymous subject is used, and won't be able to perform this action The anonymous subject is used, and won't be able to perform this action The anonymous subject is used, and won't be able to perform this action The anonymous subject is used, and won't be able to perform this action
	at org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesNamespace.doRemove(KubernetesNamespace.java:192)
	at org.eclipse.che.workspace.infrastructure.openshift.project.OpenShiftProject.cleanUp(OpenShiftProject.java:99)
	at org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesInternalRuntime.internalStop(KubernetesInternalRuntime.java:482)
	at org.eclipse.che.api.workspace.server.spi.InternalRuntime.stop(InternalRuntime.java:180)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes$StopRuntimeTask.run(WorkspaceRuntimes.java:465)
	at org.eclipse.che.commons.lang.concurrent.CopyThreadLocalRunnable.run(CopyThreadLocalRunnable.java:38)
	at java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1626)
[...]

PR #918, opened 2018-09-24
Fix broken workspace idling after Che rolling update

davidfestal:
### What does this PR do?

This PR provides a fix for broken workspace idling (after Che server restarts). However to be really effective this fix requires that the Che service account token, because in this case the Keycloak token of the user is not required anymore. 
 
This PR fixes issue https://github.com/redhat-developer/rh-che/issues/899

Partly tested on the dev cluster, but the full test should be done on prod-preview since the Che SA token is required.

### What does this PR work?

After a rolling update, the content of the [`WorkspaceSubjectRegistry`](https://github.com/redhat-developer/rh-che/blob/44fc4d1eca5ca75fe786341eb3e3afdbbeb2494e/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/WorkspaceSubjectsRegistry.java) is empty, and [so cannot help when trying to find the user Id / keycloak token of the user that started a workspace](https://github.com/redhat-developer/rh-che/blob/414a44ffa3da05d8a765305429777c4732bb031b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftClientFactory.java#L75), in order to idle it for example.
So without any user Id + Keycloak token, it's not possible to contact the OSIO API to get the user tenant data nor simply connect to the Openshift namespace through the OSO proxy.

However if the workspace is currently running, we can try to *__guess__* the user Id and the related Openshift namespace by looking into the corresponding `KubernetesRuntimeState`. And in the case the Che SA token is used to authenticate to the OSO proxy, user id and namespace will be enough to have access to the user Openshift resources.

So the following changes are done in the PR:
- When [impossible to find the user `Subject` associated to a workspace](https://github.com/redhat-developer/rh-che/pull/918/files#diff-6d6e190c7bb66102654d199a5397f276R87),
- if the Che SA token is used to authenticate the user, then search the user Id + related namespace information in the `KubernetesRuntimeState` associated to the workspace
- if it is found there, then [create a `GuessedSubject`](https://github.com/redhat-developer/rh-che/pull/918/files#diff-6d6e190c7bb66102654d199a5397f276R99) containing the user ID and the namespace, and and use it for Openshift configuration build instead of the current `Subject`
- Then if the a `GuessedSubject` has been used, [`Fabric8WorkspaceEnvironmentProvider` will detect it and use the `GuessedSubject` user Id and namespace](https://github.com/redhat-developer/rh-che/pull/918/files#diff-f2821ac62d8158e3bafa443f0876cddeR159) in order to populate the user tenant data.


davidfestal:
test

rhopp:
[test]

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/GuessedSubject.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package com.redhat.che.multitenant;
+
+import org.eclipse.che.api.core.ForbiddenException;
+import org.eclipse.che.commons.subject.Subject;
+
+public class GuessedSubject implements Subject {
```
ibuziuk: @davidfestal could you please add javadoc with explanation for this class and when it is supposed to be used ?

garagatyi:
@davidfestal can you add an explanation of what actually was changed in the PR (what instead of why)? It would significantly help us to thoughtfully review the PR.

davidfestal:
@garagatyi sure !

davidfestal:
@garagatyi I complete the PR description with what you requested.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/GuessedSubject.java
davidfestal: Is it OK for you ?

davidfestal:
test

davidfestal:
[test]

davidfestal:
The last build failure doesn't seem related to PR code anymore.
Do you think you could approve the PR if the description and code implementation is OK for you ?

garagatyi:
@davidfestal thanks!

Issue #845, opened 2018-09-03
Che does not show bayesain recommendations for package.json (only pom.xml) 


ibuziuk:
Placeholder for https://github.com/openshiftio/openshift.io/issues/3878

PR #932, opened 2018-09-27
Fix file regexes used in fabric8 analytics lsp server

amisevsk:
### What does this PR do?
Modifies the regexes used for watching file changes for the fabric8 analytics language server.

Although the problem is easy to see (the server doesn't get activated for package.json files), matching the `package.json` regex with the `pom.xml` regex (which does work) does not solve the problem. With these changes, the server is initialized correctly.

As far as I can tell, using the regex `.*/package\\.json$` (cf `.*/pom\\.xml` for pom files) *should* work, but the leading `/` prevents the server from being activated on package.json files. I've opted to bring the regexes in line with [`JsonLanguageServerConfig`](https://github.com/eclipse/che/blob/18cba8b55ad7bbff710902a48c05cba4312b22a2/plugins/plugin-json/che-plugin-json-server/src/main/java/org/eclipse/che/plugin/json/languageserver/JsonLanguageServerConfig.java#L38) instead, but would like to understand why this issue is happening at all.

#845 

Locally; I can't seem to import any projects on the dev-cluster.


amisevsk:
@ibuziuk Finally got around to testing this. The language server is activated for pom.xml and package.json files anywhere in the source tree, as expected.

I still have absolutely no idea why the original regex didn't work though.

ibuziuk:
@amisevsk thanks for checking. +1 to merge the PR. Maybe we could create an issue upstream for discussing this problem and get some feedback from other teams

Issue #920, opened 2018-09-25
Implement 'native' build profile for dashboard from upstream


ScrewTSW:
**Issue problem:** 
Currently as of 30th of August rh-che dashboard is being built with `docker`.
This poses an issue where rh-che needs to be build inside a container.
Usually we can get around this by mounting `-v /var/run/docker.sock:/var/run/docker.sock` but in some cases like `docker build` context, this is not possible.
This profile is critical for PR-Check dependency image builds as the idea is to migrate rh-che builds into a docker container to speed up the process significantly.

upstream maven profile: https://github.com/eclipse/che/blob/master/dashboard/pom.xml#L174

**Red Hat Che version:** 

version: since 6.10.x

 - [X] I can reproduce it on latest official image

**Reproduction Steps:**

+ pull latest rh-che master
+ `cd rh-che`
+ `docker build dockerfiles/prcheck/`
build fails on dashboard:  
```
[INFO] Fabric8 IDE :: Docs ................................ SUCCESS [  0.527 s]
[INFO] Fabric8 IDE :: Assemblies :: Dashboard ............. FAILURE [  4.023 s]
[INFO] Fabric8 IDE :: Assemblies :: Main/Tomcat ........... SKIPPED
[INFO] functional-tests ................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 17:53 min
[INFO] Finished at: 2018-09-25T13:34:03+00:00
[INFO] Final Memory: 152M/1521M
[INFO] ------------------------------------------------------------------------
[WARNING] The requested profile "native" could not be activated because it does not exist.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (build-image) on project fabric8-ide-dashboard-war: An Ant BuildException has occured: exec returned: 1
[ERROR] around Ant part ...<exec failonerror="true" dir="/tmp/rh-che/assembly/fabric8-ide-dashboard-war" executable="docker">... @ 4:101 in /tmp/rh-che/assembly/fabric8-ide-dashboard-war/target/antrun/build-main.xml
```

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [ ] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

PR #937, opened 2018-10-01
Add native build profile to dashboard submodule

amisevsk:
### What does this PR do?
Adds native build profile as in upstream, activated by arg `-Pnative`, to support PR-check and CI builds.

https://github.com/redhat-developer/rh-che/issues/920

Builds with and without `-Pnative` succeed locally.

### DIFF
--- a/assembly/fabric8-ide-dashboard-war/pom.xml
+++ b/assembly/fabric8-ide-dashboard-war/pom.xml
@@ -94,46 +94,7 @@
                             </target>
                         </configuration>
                     </execution>
-        <!--   3. Build dashboard as in upstream, using docker image   -->
-                    <execution>
-                        <id>build-image</id>
-                        <phase>generate-sources</phase>
-                        <goals>
-                            <goal>run</goal>
-                        </goals>
-                        <configuration>
-                            <target>
-                                <!-- build user dashboard with maven -->
-                                <exec dir="${basedir}" executable="docker" failonerror="true">
-                                    <arg value="build" />
-                                    <arg value="-t" />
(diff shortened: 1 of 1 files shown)


assembly/fabric8-ide-dashboard-war/pom.xml
```diff
@@ -170,4 +131,124 @@
...
+        </profile>
+        <profile>
+            <!-- Upstream native build. Activate with -Pnative -->
+            <id>native</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <executions>
+                            <execution>
```
Katka92: Maybe these steps should be commented as 3. and 4. as in "docker" profile? Anyway seems good to me :)

rhopp:
@ScrewTSW Could you try that and confirm this is what you needed?

ScrewTSW:
@rhopp Tested and it's exactly what we need :) PR-Check works flawlessly with this change

assembly/fabric8-ide-dashboard-war/pom.xml
amisevsk: You're probably right, I largely copied over config.

Worth noting: The native build is *slightly* different from upstream: we skip the clean step (since all artifacts are in /target already) and we defer the href update to the main build process. I'll open a new PR for documentation on this.

amisevsk: PR here: https://github.com/redhat-developer/rh-che/pull/939

PR #940, opened 2018-10-02
Use custom timeout in waitStatus

Katka92:
### What does this PR do?
Uses upstream method for custom timeout - logic moved, replaced only by function call.

Upstream change - https://github.com/eclipse/che/pull/11430

localy


Katka92:
[test]

PR #942, opened 2018-10-02
Adding test for bayesian

Katka92:
### What does this PR do?
Adds test for bayesian in pom.xml files. Test opens pom.xml file, adds vulnerable dependency and waits for error to appear. Once error appears, test closes pom.xml file. When reopening file, test checks if error is still present. 

https://github.com/redhat-developer/che-functional-tests/issues/291

Localy against prod and prod-preview. Prod-preview is failing (known issue - added catch to code to pass).


Katka92:
@amisevsk thanks! Yes sure, we are counting on it. We already have an issue for it, maybe we will add it into next sprint.

PR #954, opened 2018-10-05
Temporary increasing memory limits to 4Gi to workaround pod restarts with OOM on production

ibuziuk:
### What does this PR do?
Temporary increasing memory limits to 4Gi to workaround pod restarts with OOM on production

https://github.com/redhat-developer/rh-che/issues/898


ibuziuk:
@jfchevrette @pbergene could you please comment if we could temporary afford extra 2.5Gb on rh-idev/dsaas for decreasing the number of pod restarts with OOM ?

jfchevrette:
@ibuziuk +1 to this workaround until we can find a proper solution

Issue #941, opened 2018-10-02
Protect the End2end flow from bots


davidfestal:
We should protect the end2end provisioning flow of `che.openshift.io` against bots, so that they cannot sucessfully pass the regsitration flow as it is the case currently.

That's related to the following HK issues:
https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2351
https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2356

garagatyi:
@davidfestal Could you add more details on what exactly is needed to be done to close this issue? Could you also label the complexity of this issue? I think it might be useful to let someone but you in the team to work on that to share knowledge about e2e flow and avoid bus-factor.

davidfestal:
@garagatyi I created this issue after discussing with @ibuziuk and the SD team about the problem.
And it's on purpose that I didn't add technical details directly in this issue description.

Please see
https://chat.openshift.io/developers/pl/jhrpu5xfj7yqjnfrph6kirf69c

See also:
https://chat.openshift.io/developers/pl/wcbjirbottyq5mttr31sbtoqdh
https://chat.openshift.io/developers/pl/dceqbe3p83fgzjm5b8azof7bch

I didn't assign myself to the issue, so even if I'm OK to start immediately on this as a critical issue, I see no problem if another team member would work on it, and I will provide as much information as necessary for him to start with.

So let's put `intermediate` as a level, unless you have another feeling.

davidfestal:
This has been deployed to production, but we're waiting for HK issue https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2377 to be fixed in order to have the protection enabled.

ibuziuk:
@davidfestal HK issue was closed. I guess we can close this issue now ?

PR #960, opened 2018-10-09
End2end flow protection

davidfestal:
### What does this PR do?

This PR protects the activation page of the end2end provisioning flow from bots.

For more detailed desciption, please see the following HK issue: https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2356#note_439944

This PR fixes issue https://github.com/redhat-developer/rh-che/issues/941

Yes, on Dev-cluster.


sbose78:
Nice. How do we test the invisible recaptcha? Any way to trigger the `I am not a human` flow?
Also, how do we turn it off for a specific environment?

davidfestal:
@sbose78 let me answer in the HK issue

l0rd:
Cool!

davidfestal:
[test]

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
```diff
@@ -0,0 +1,217 @@
...
+        try (BufferedReader in = new BufferedReader(new InputStreamReader(con.getInputStream()))) {
+          String inputLine;
+          while ((inputLine = in.readLine()) != null) {
+            response.append(inputLine);
+          }
+        }
+        return Response.ok(response.toString()).build();
+      } else {
+        LOG.error(
+            "reCaptcha verification failed with the fllowing response code: {}", responseCode);
```
garagatyi: Might be helpful to add response body to the log in a case some unexpected error is received

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
```diff
@@ -0,0 +1,217 @@
...
+      try (DataOutputStream wr = new DataOutputStream(con.getOutputStream())) {
+        wr.writeBytes(urlParameters);
+        wr.flush();
+      }
+
+      responseCode = con.getResponseCode();
+
+      if (responseCode == 200) {
+        StringBuffer response = new StringBuffer();
+        try (BufferedReader in = new BufferedReader(new InputStreamReader(con.getInputStream()))) {
```
garagatyi: You can use `IoUtil.readAndCloseQuietly()` instead of the bunch of the code

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
```diff
@@ -0,0 +1,217 @@
...
+  @POST
+  @Path("error")
+  @Consumes("text/plain")
+  public void error(String message, @Context HttpServletRequest servletRequest) {
+    LOG.error(getLog(message, servletRequest));
+  }
+
+  @GET
+  @Path("{path: files\\/.*}")
+  public Response staticResources(@PathParam("path") final String path) {
```
garagatyi: It is not very common to host files using a REST service. Why not use servlet for that? I suppose there is ready servlet somewhere in Che for this purpose.

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
```diff
@@ -0,0 +1,217 @@
...
+    return Response.status(responseCode).build();
+  }
+
+  private String getLog(String message, HttpServletRequest servletRequest) {
+    String ip = servletRequest.getHeader("X-Forwarded-For");
+    if (ip == null) {
+      ip = servletRequest.getRemoteAddr();
+    }
+
+    return new StringBuffer("[E2E Registration Flow - IP = ")
```
garagatyi: We have 124k events like that on prod already. Can you elaborate on why we need to log all these events and what profits of having them compared to not having?

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
davidfestal: Is it a problem ? It seems much easier to configure / maintain than using a Servlet, no ?

Additionally this was done this way in a similar context, and merged on upstream: https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakConfigurationService.java#L54

So if you don't mind, I'd rather let it like that.

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
davidfestal: This seems important to track the registration errors.
This huge number is obviously due to bots aggressively and constantly trying to register.
It might also indicate some erroneous code flow in the end2end automatic provisioning, which is important to track as well.

And even without bots / flow errors to track, it seems critical to be able to track the cases when true users didn't succeed the `che.openshift.io` registration fully.

davidfestal:
@garagatyi I fixed 2 of your comments, and answered the 2 remaining ones.
Is it OK for you now ?

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
garagatyi: I wouldn't block the PR because of that, but I do consider binding of a component with a couple of lines easier to support than 60 lines of code.

davidfestal: I can change it if you point me to the right Che ready-to-use file-serving Servlet component.
However let me note that the content of the files is optionally filtered before returning it. Would the out-of-the-box Che Servlet component provide this ability ?

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
garagatyi: I doubt that we will analyze 124k log entries (and this is only for 2 weeks), but if you believe we need them let's leave them here.

plugins/fabric8-end2end-flow/src/main/java/com/redhat/che/end2end/End2EndFlowService.java
garagatyi: I'm OK to leave it. Don't see much sense to argue about it.

garagatyi:
@rhopp @ScrewTSW @Katka92 when CI says that server was successfully deployed does it also mean that we did some automatic tests and they passed?

rhopp:
@garagatyi Yes. So far, only [this test class](https://github.com/redhat-developer/che-functional-tests/blob/master/tests/src/test/java/redhat/che/functional/tests/MavenTestCase.java) is executed. It starts vert.x workspace with project imported (using che-starter) and does maven build on the project. It's just a basic check, that workspace is able to start and operate.

garagatyi:
@rhopp OK, it is what I expected. Can we change the wording in the PR comment that CI posts after a successful run of tests to reflect that it is not just deployed but actually tested?

Issue #911, opened 2018-09-24
Add and track Che telemetry events for 'push to remote', 'create project', and 'import project' 


davidfestal:
Related to EPIC https://github.com/redhat-developer/rh-che/issues/765

ibuziuk:
@slemeur could you please clarify if it is a priority for the next sprint

slemeur:
Yes, I think this one is pretty important as it allows us to better know what the users are doing and also qualify the level of engagement. 

This issue has an higher priority than #908 .

davidfestal:
Here are some guidelines to add the support of a new event:
- Add specific properties of the new event (ie `PROGRAMMING_LANGUAGE` is a property specific to the `edit workspace file in Che` event):
https://github.com/redhat-developer/rh-che/blob/master/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/EventProperties.java#L15
- Add the declaration for the new event, with the estimated duration of the event, and the related specific properties:
https://github.com/redhat-developer/rh-che/blob/master/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java#L22
- Add the submission of the new event when the related action is intercepted.
https://github.com/redhat-developer/rh-che/blob/master/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/UrlToEventFilter.java#L116
As shown by the code example above, for now tracked actions are intercepted through a filter on the corresponding wsagent API. Though limited and subject to become obsolete (because wsagent will disappear with ws.next), this works quite well for the already defined events, and should work also for `create project`, `project import` and (possibly) for `push to GitHub` events.
So just have a look at the corresponding endpoint calls in the wsagent access log that can be found in the `/workspace_logs/ws-agent/logs/` directory, to find the precise API call that would best fit.

Of course, on could also subscribe to published events  or even intercept websockets activity to catch much more events. However I don't think it is worth the work for now because of the following upcoming  issue: https://github.com/redhat-developer/rh-che/issues/690.

garagatyi:
@davidfestal thanks for the details!

amisevsk:
[...]

PR #963, opened 2018-10-10
Add 'commit locally' and 'push to remote' events in analytics

amisevsk:
### What does this PR do?
Adds commit locally and push to remote analytics events, that fire when a user tries to do either action in Che.

#911 (the parts that are easy to implement, at least)

On dev cluster; two events are correctly displayed.


slemeur:
Good !

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
```diff
@@ -19,7 +19,9 @@
   WORKSPACE_USED("Use Workspace in Che"),
   WORKSPACE_INACTIVE("Keep Workspace Inactive in Che"),
   WORKSPACE_STOPPED("Stop Workspace in Che"),
-  EDITOR_USED("Edit Workspace File In Che", 30, new String[] {PROGRAMMING_LANGUAGE});
+  EDITOR_USED("Edit Workspace File In Che", 30, new String[] {PROGRAMMING_LANGUAGE}),
+  PUSH_TO_REMOTE("Push to remote"),
```
davidfestal: Isn't there some information from the endpoint call that we could use to add properties to the new event ?

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
```diff
@@ -19,7 +19,9 @@
...
+  COMMIT_LOCALLY("Commit locally");
```
davidfestal: I would probably add an expected duration to both commit and push event.
Maybe something like 10 seconds ? That seems reasonable as the duration of such an action.

Adding an expected duration also avoids sending 2 events too close from one another in time, which, afaict, Woopra doesn't like much.

davidfestal: Also could you add `in Che` at the end of the event names please ?

This makes it very easy to group all the events that happen in a Che workspace, which is necessary to sum-up the whole usage duration (since there's only 1 current action at a given time).

davidfestal:
@l0rd Do you think you could merge this ? @rhopp confirmed that we should be able to merge, because the failing build is related to https://github.com/openshiftio/openshift.io/issues/4020.

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
amisevsk: The only info is project path, which is IMO something we shouldn't pull as telemetry. We could attempt to resolve project type (e.g. maven) but I don't know how useful that is.

plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsEvent.java
amisevsk: Fixed

amisevsk: Done.

amisevsk:
@davidfestal Added tests. I forgot :)

davidfestal:
@amisevsk Is there a reason why we don't merge this ?

amisevsk:
[test]

Issue #958, opened 2018-10-09
Fix Che dashboard build


garagatyi:
**Issue problem:** 
UD is no longer builds in RH-CHE (against latest snapshot) after the changes in latest upstream.
The error in build logs:
<details>

```
10:37:31      [exec] [91mnpm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules/chokidar/node_modules/fsevents):
10:37:31      [exec] npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
10:37:31      [exec] [0m[91mnpm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.2.3 (node_modules/sane/node_modules/fsevents):
10:37:31      [exec] [0m[91mnpm WARN notsup[0m[91m SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
10:37:31      [exec] [0m[91mnpm WARN angular-gravatar@0.2.4 requires a peer of grunt-cli@0.1.13 but none was installed.
10:37:31      [exec] [0m[91mnpm WARN[0m[91m file-loader@1.1.11 requires a peer of webpack@^2.0.0 || ^3.0.0 || ^4.0.0 but none was installed.
10:37:31      [exec] npm WARN ts-jest@22.4.6 requires a peer of typescript@2.x but none was installed.
10:37:31      [exec] [0m[91mnpm WARN che-dashboard@4.0.0 No description
10:37:31      [exec] [0m[91mnpm WARN optional SKIPPING OPTIONAL DEPENDENCY: bufferutil@1.1.0 (node_modules/bufferutil):
10:37:31      [exec] npm WARN optional[0m[91m SKIPPING OPTIONAL DEPENDENCY: bufferutil@1.1.0 install: `node-gyp rebuild`
10:37:31      [exec] npm WARN optional[0m[91m SKIPPING OPTIONAL DEPENDENCY: Exit status 1
10:37:31      [exec] [0m[91mnpm WARN optional SKIPPING OPTIONAL DEPENDENCY: utf-8-validate@1.1.0 (node_modules/utf-8-validate):
[...]

Issue #944, opened 2018-10-03
Dashboard build broken when using native profile


ScrewTSW:
Recently after issues with master build went away, the dependency build started failing at the dashboard submodule.

command used for build:
`scl enable rh-maven33 rh-nodejs4 "mvn -B -f /tmp/rh-che/ -Pnative clean install"`

Build ends with an error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (copy-dist-from-sources-dir) on project fabric8-ide-dashboard-war: An Ant BuildException has occured: /tmp/rh-che/assembly/fabric8-ide-dashboard-war/target/sources/target does not exist.
[ERROR] around Ant part ...<copy todir="/tmp/rh-che/assembly/fabric8-ide-dashboard-war/target/" overwrite="true" verbose="true">... @ 4:104 in /tmp/rh-che/assembly/fabric8-ide-dashboard-war/target/antrun/build-main.xml
```

* possibly missing mkdir step?
* is the path correct?

**Red Hat Che version:** 

version: 6.11.1

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**

+ clone repo into the temp directory
+ run `docker build ./dockerfiles/pr-check/`
+ build fails at dashboard step

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [ ] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

PR #964, opened 2018-10-10
Fix pr-check build issues when building dashboard

amisevsk:
### What does this PR do?
Fixes a few issues in building dashboard using native profile
- Bower update step requires non-root user; fails silently otherwise. Build now creates a new user and executes the git clone and maven build as them
- It seems a higher verison of nodejs is required to build Che; there is some issue in the `gulp build` step that occurs (module not found) when nodejs4 is used. Uses the rh-nodejs8 package instead.
  - I'm guessing nodejs8 will also be required for https://github.com/redhat-developer/rh-che/issues/958
- Running `npm install -g gulp bower` is no longer necessary.

#944 

Build succeeds locally executing `docker build ./dockerfiles/pr-check`


dockerfiles/pr-check/Dockerfile
```diff
@@ -6,11 +6,31 @@
...
+        java-1.8.0-openjdk-devel \
+        centos-release-scl \
+ && yum install --assumeyes \
+        rh-maven33 \
+        rh-nodejs8 \
+ && yum clean all \
+ && rm -rf /var/cache/yum
+
+# Bower update step won't work for root user
+RUN useradd -ms /bin/bash newuser
```
ScrewTSW: Nice :) We should still try to keep the RUN commands minimized into only single call because every time RUN is called, it creates a new docker layer, thus increasing the docker image size

amisevsk: Creating new docker layers isn't a big problem if the `RUN` commands affect different files; the issue arises when e.g. files are created in one `RUN` command and then deleted in another. `docker inspecting` images with and without the `useradd` step being combined with the previous `RUN` gives
- `"Size": 2304547111` (2304.55MB) for current PR
- `"Size": 2304254909` (2304.25MB) for dockerfile with useradd combined with previous command.

I also think this is more maintainable and makes more sense, but I'm happy to make the change if you prefer.

ScrewTSW: :+1: in that case. Thank you for explaining the inner workings of docker :) I think we are ready for merge then.

PR #970, opened 2018-10-15
Update upstream Che version to 6.12.2

garagatyi:
### What does this PR do?
Update to the next upstream bugfix version to check compatibility.
It brings a lot of new [commits](https://github.com/eclipse/che/commits/6.12.2). There are other fixes in master that we also need. I'll work on releasing one more bugfix version to deliver them.


garagatyi:
[test]

garagatyi:
[test]

Issue #971, opened 2018-10-16
Rh-che cannot recover workspaces coming from OpenShift recipe types


amisevsk:
### Issue problem:
At che server startup, workspace recovery fails for workspaces based off of OpenShift recipe types.

#### Details:
When OpenShift recipe workspaces are recovered, [OpenShiftEnvironmentFactory uses an OpenShift client to parse the yaml of the recipe](https://github.com/eclipse/che/blob/c7c697f1e3dfc918acd149d128eeeccc7df4b2c2/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java#L114). This requires creating an OpenShift client. Creating the client requires calling [buildConfig](https://github.com/redhat-developer/rh-che/blob/7ac14815f44bbc80e89d3ec23c5b333ef32dbc42/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftClientFactory.java#L77), which fails due to anonymous subject because there are no subjects in the cache at startup, and a workspaceId is not specified. 

### Red Hat Che version: 
Current master

### Reproduction Steps:
1. Create and start workspace using the "Java + MySQL DB" stack type (which is an OpenShift recipe)
2. After workspace is started, scale down or deploy new version of rhche
3. Message
   ```
   An error occurred while attempting to recover runtimes using infrastructure 'openshift'. Reason: 'Couldn't recover runtime '<workspace-id>:default'. Error: The anonymous subject is used, and won't be able to perform this action'
   ```
   is logged

amisevsk:
Fixed by #972

PR #972, opened 2018-10-16
Use default openshift config for client when we can't get a subject

amisevsk:
### What does this PR do?
Return default OpenShiftConfig when we cannot get a subject and no workspaceId is available. There are situations where it is necessary to create an OpenShiftClient without a subject available.

During the workspace recovery step in startup, the server creates a client and uses it to parse the yaml of an OpenShift recipe. We don't necessarily need a subject for this, as the client is used only to `load` and does no interaction with the cluster. 

#971 

Tested only locally for the moment, I'm having issues on dev cluster.


davidfestal:
@amisevsk Good catch !

However, it seems that the case we have to support is the `KubernetesClientFactory.create()` method (without a `workspaceId` parameter), which in turn calls the `buildConfig(Config config, String workspaceId)` with a `null` value for `workspaceId`.
So It seems to me that, in `Fabric8OpenShiftClientFactory.buildConfig(Config defaultConfig, String workspaceId)`, we should return the `defaultConfig` when:
```
workspaceId == null && (subject == null || subject.isAnonymous())
```
without even emitting a warning.

But, afaict, you still return the `defaultConfig` in the block where the `workspaceId` is __not__ `null` and `runtimeIdentity` is not present: https://github.com/redhat-developer/rh-che/pull/972/files#diff-6d6e190c7bb66102654d199a5397f276R111

I would move your code in the outer `else` block, but I might be missing some point.

garagatyi:
Good catch, indeed! 
I agree with David - that `else` should be for the outer `if` clause. But I would also emit a warning in all the cases when workspaceID is present but the default config is used. This is a situation that should not happen and this warning should identify a bug in the code.

davidfestal:
> But I would also emit a warning in all the cases when workspaceID is present but the default config is used. This is a situation that should not happen and this warning should identify a bug in the code.

+1

I would even say that, in the case the `workspaceId` is present but the subject is `null` or anonymous, we should keep it like it was currently, and not return the default config at all, since it will finally not allow connecting to the right Openshift namespace, and the error would be even more confusing.
Until now there was an exception thrown at least telling the anonymous subject could not perform the workspace-related operation..

That doesn't prevent us to log a warning at this place to help troubleshooting this unexpected case.

amisevsk:
@davidfestal @garagatyi Thanks for catching that guys, I moved the check to the bottom right before opening the PR and mixed up my indentation levels -- I meant for the `else if` to be an else to `workspaceId != null`.

Sorry for any confusion there. I'm emitting the warning because I think it could be useful if a similar issue happens later, as the "Anonymous subject is used" message is hard to trace.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftClientFactory.java
```diff
@@ -108,6 +115,12 @@ public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
             }
           }
         }
+      } else { // !runtimeIdentity.isPresent()
+        LOG.warn(
```
garagatyi: Shouldn't we throw an exception here? If we can't get user identity OSO Proxy would not allow any operation, so we can just fail here.
@davidfestal WDYT?

amisevsk:
[test]

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftClientFactory.java
amisevsk: From discussion on mattermost, we're deferring this until later to avoid risking regressions.

PR #973, opened 2018-10-16
Rh-che compatibility test

Katka92:
### What does this PR do?
Adding scripts for testing compatibility of latest che and rh-che. Should deploy rh-che with latest che on dev cluster and run basic tests.


.ci/cico_build_deploy_test_rhche.sh
```diff
@@ -0,0 +1,166 @@
...
+set -x
+set -e
+set +o nounset
+
+/usr/sbin/setenforce 0
+
+export BASEDIR=$(pwd)
+export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
+export OC_VERSION=3.9.33
+export TARGET=${TARGET:-"rhel"}
```
davidfestal: Are you sure you want to have `rhel` be the default target ?
What's going to happen if we try to run the script locally on a RHEL laptop ?

Katka92: @ScrewTSW What do you think about that?

Katka92:
[test]

.ci/cico_build_deploy_test_rhche.sh
```diff
@@ -0,0 +1,166 @@
...
+yum install --assumeyes \
+            rh-maven33 \
+            rh-nodejs4
+
+systemctl start docker
+pip install yq
+
+set +x
+#change version of used che
+if [[ "${USE_CHE_LATEST_SNAPSHOT}" == "true" ]]; then
```
garagatyi: If we have a job that changes the version of upstream Che it should call a specific script (this one). We can have all the custom logic here and then execute only common logic with a default script. 
In that case, the default script would be simpler (less if clauses) and this script would encapsulate all the information about the change it brings. If some info, such as image tag or namespace is needed in the default script we can pass all the info using env vars.

.ci/cico_build_deploy_test_rhche.sh
```diff
@@ -0,0 +1,166 @@
...
+fi
+
+# Build and push image to docker registry
+source ./config
+BuildTagAndPushDocker
+
+# Deploy rh-che image
+echo "Deploying nightly-${RH_TAG_DIST_SUFFIX} from ${DOCKER_IMAGE_URL}"
+
+if ./dev-scripts/deploy_custom_rh-che.sh -o "${RH_CHE_AUTOMATION_DEV_CLUSTER_SA_TOKEN}" \
```
garagatyi: As far as I understand in this case all the output of that script is suppressed. Shouldn't we let script log all its logs to be able to trace it in a job output in a case something breaks?

Katka92: Sorry, I'm not sure what you mean about that. The messages that are printed are readable in log, e.g.:
```
Deploying upstream-check-latest from quay.io/openshiftio/rhel-rhchestage-rh-che-automation
[1mLogging in to openshift...[0m
Login successful, creating project compatibility-check
Switched to project compatibility-check
[92;1mProject created successfully.[0;38;5;238m
[0;1mGetting deployment scripts...[0m
[92;1mGetting deployment scripts done.[0m
Waiting for postgres deployment to be deleted...................
Postgres deployment successfully deleted.
[1mDeploying PostgreSQL[0;38;5;60m
[0;38;5;60mPostgres status: Available:[0;1m"False" [0;38;5;60mProgressing:[0;1m"Unknown" [0;38;5;60mTimeout:[0;1m180[0m
[0;38;5;60mPostgres status: Available:[0;1m"False" [0;38;5;60mProgressing:[0;1m"Unknown" [0;38;5;60mTimeout:[0;1m179[0m
[0;38;5;60mPostgres status: Available:[0;1m"False" [0;38;5;60mProgressing:[0;1m"Unknown" [0;38;5;60mTimeout:[0;1m178[0m
...
```
You mean that this is not enough and there should be any other logs?

.ci/prepare_env_utils.sh
```diff
@@ -0,0 +1,72 @@
...
+	            golang \
+	            make \
+	            jq \
+	            java-1.8.0-openjdk \
+	            java-1.8.0-openjdk-devel \
+	            centos-release-scl
+	
+	yum install --assumeyes \
+	            rh-maven33 \
+	            rh-nodejs4
```
amisevsk: Does it make sense to use `rh-nodejs8` here? The docs are a bit in disagreement, but I had to use Node 8 for the pr-check plugin.

https://github.com/eclipse/che/tree/master/dashboard#quick-start

ScrewTSW: I'm not sure what effects does switching to nodejs8 have, however if it's working with the prcheck, I don't see any issues there. we can use rh-nodejs8 instead.

ScrewTSW: I'm not sure what effects does switching to nodejs8 have, however if it's working with the prcheck, I don't see any issues there. we can use rh-nodejs8 instead.

ScrewTSW: I'm not sure what effects does switching to nodejs8 have, however if it's working with the prcheck, I don't see any issues there. we can use rh-nodejs8 instead.

ScrewTSW: I'm not sure what effects does switching to nodejs8 have, however if it's working with the prcheck, I don't see any issues there. we can use rh-nodejs8 instead.

ScrewTSW: I'm not sure what effects does switching to nodejs8 have, however if it's working with the prcheck, I don't see any issues there. we can use rh-nodejs8 instead.

Katka92:
[test]

PR #974, opened 2018-10-16
Adding Che 7 Preview stacks

l0rd:
### What does this PR do?

Add a couple of Che 7 stacks: a first one to simple use a che7 workspace and add plugins, a second one with the tooling needed to build and publish a plugin.

https://github.com/eclipse/che/issues/11502

I have run them on openshift.io.


l0rd:
cc @skabashnyuk

benoitf:
@l0rd you need to apply formatter. Let me know if you can't do it

[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (check-headers) on project fabric8-ide-stacks: Some files do not have the expected license header -> [Help 1]

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -64,6 +64,152 @@
...
+    ],
+    "workspaceConfig": {
+      "name": "che7-preview-plugin-dev",
+      "defaultEnv": "default",
+      "environments": {
+        "default": {
+          "machines": {
+            "ws/dev": {
+              "attributes": {
+                "memoryLimitBytes": "536870912"
```
garagatyi: 500+ MB might be too much. We would want to have extra space for several sidecars. WDYT?

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -64,6 +64,152 @@
...
+    ],
+    "workspaceConfig": {
+      "name": "che7",
+      "defaultEnv": "default",
+      "environments": {
+        "default": {
+          "machines": {
+            "ws/dev": {
+              "attributes": {
+                "memoryLimitBytes": "536870912"
```
garagatyi: We don't need both an attribute and memory limit in the recipe. Can we remove one of them?

assembly/fabric8-stacks/src/main/resources/stacks.json
garagatyi: The same for a second stack

assembly/fabric8-stacks/src/main/resources/stacks.json
garagatyi: The same for a second stack

l0rd:
@garagatyi I have already merged. I need to do another PR anyway to update those stacks in the next days and will tests your suggestions (reduce memory and remove the attribute).

Issue #976, opened 2018-10-16
Regressions and errors following the deployment of the end2end flow protection


davidfestal:
**Issue problem:** 

- The End2End flow protection recently released to production introduced a regression that makes a new requirement for the user to disable popups on his browser. 
- There are a number of log entries in Sentry that show some cleaning / enhancements to be done in the way errors are caught on the client-side and sent back on the server-side.

**Red Hat Che version:** 

OSIO production and prod-preview

ibuziuk:
@davidfestal do you think we can close this issue now ?

ibuziuk:
I believe this issue can be closed now

davidfestal:
This HK issue should be fixed before deploying the last PR to production:

https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2414

PR #977, opened 2018-10-16
Fix regressions and errors due to the e2e flow protection deployment

davidfestal:
### What does this PR do?

The PR fixes a regression introduced by the e2e flow protection deployment, and cleans / enhances the way errors are caught on the client-side and sent back to the server-side. 

#976

On dev-cluster against the production OSIO infrastructure.

### Demo

https://youtu.be/W_u-yACab10


plugins/fabric8-end2end-flow/src/main/resources/end2end/files/RhCheKeycloak.js
```diff
@@ -244,7 +206,7 @@ function initAnalytics(writeKey){
         return new Promise((resolve, reject) => {
             var request = new XMLHttpRequest();
             request.onerror = request.onabort = function(error) {
-                reject(error);
+                reject(this);
```
benoitf: it's unsual to reject this ( usually it's only errors)

davidfestal: Well, in the `get` and `post`, when an error occurs (or a non-successful status code), I just return the request itself. This way, in the calling code, I can extract whatever information I need from the request.
Is it a problem ?

PR #983, opened 2018-10-17
Last fixes to the end2end flow

davidfestal:
### What does this PR do?

Fixes some light remaining bugs after the last end2end flow PRs

This PR is just a followup of the last end2end flow -related PRs


davidfestal:
@garagatyi Do you think we could wait for this fix to be on prod-preview before releasing to production ?

davidfestal:
@l0rd That would be great if we could wait for this fix to be on prod-preview before releasing to production.
Let me merge now, since the changes won't impact the build and tests at all, OK ?

PR #991, opened 2018-10-18
Add Quay imagepullsecrets to rhche ServiceAccount

aditya-konarde:
### What does this PR do?
This PR will add an entry with the quay imagepullsecret set in the SA definition.

When this template processed and applied, the existing `rhche` SA that has the `quay.io` imagepullsecrets linked would get overwritten, and unlink the image pull secret. This leads to a permission denied error pulling the image.


davidfestal:
yes, let me take a look

PR #993, opened 2018-10-19
Change the title of Che7 stacks

benoitf:
### What does this PR do?
Change title

N/A

Change-Id: I2512e7cfa8bb3f747c8c5ad71d6b434e9f1ef07b


assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -141,7 +141,7 @@
   {
     "id": "che7-preview-plugin-dev",
```
l0rd: Why haven't you updated the id here? Is it because it would duplicate the stack in prod?

benoitf: I was able to see this one into prod so I kept the id
you think that I need to rename id here ?

l0rd: I would change it just to be consistent with the dev stack. And maybe to be sure that it gets updates. But honestly I am not familiar with the stack update mechanism.

PR #996, opened 2018-10-21
Add better implementation of Theia plugin broker

garagatyi:
### What does this PR do?
It processes .theia files instead of che specific yamls. It supports both regular theia plugins and remote theia plugins. Not to depend on changes in upstream plugins processed by this broker should be called `Theia plugin`.


garagatyi:
[test]

PR #997, opened 2018-10-21
Added commands for building and publishing a formatter plugin

sleshchenko:
### What does this PR do?
Adds commands for building and publishing a formatter plugin.


assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -217,6 +217,18 @@
           "name": "Get OpenShift Console URL",
           "type": "",
           "attributes": {}
+        },
+        {
+          "commandLine": "cd /projects/formatter && yarn",
+          "name": "Build Plugin",
```
sleshchenko: @l0rd Would it better to name these commands `Build Formatter Plugin` and `Publish Formatter Plugin` since information about formatter is hard-coded?

l0rd: Can you automatically get the project folder name (you can assume that there is just one folder in /projects/). That would allow to use the same command for all plugins.

sleshchenko: Sure, `cd /projects/*` should work fine in this case I suppose. I will try

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -217,6 +217,18 @@
...
+          "type": "",
+          "attributes": {}
+        },
+        {
+          "commandLine": "publish_plugin formatter 1:0:0 theia",
```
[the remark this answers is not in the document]
sleshchenko: well, I think `Che Command` is not the best place to do that.
If we need publishing without arguments - maybe publish_plugin may do that?

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -217,6 +217,18 @@
           "name": "Get OpenShift Console URL",
           "type": "",
           "attributes": {}
+        },
+        {
+          "commandLine": "cd /projects/* && yarn",
```
garagatyi: Please remove asterisk. Otherwise it would fal

assembly/fabric8-stacks/src/main/resources/stacks.json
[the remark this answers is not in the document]
l0rd: Correct. I have done it in this [branch here](https://github.com/ws-skeleton/che-plugin-dev-tooling/blob/publish-docker-build-imprmts/publish_plugin.sh#L6). Will merge when eclipsecon is over.

assembly/fabric8-stacks/src/main/resources/stacks.json
l0rd: I don't think it will work. You may check if there is a `package.json` in current folder and run yarn:
`[ -f ./package.json ] && yarn || echo ERROR: no package.json found in current folder $(pwd)`

sleshchenko:
`Build Plugin` command should be reworked when macro `{current.project.path}` will be available.
`publish_plugin` script should be reworked to evaluate name, version and type of plugin automatically.

l0rd:
@sleshchenko I have already worked on the improvements to `publish_plugin` script [here](https://github.com/ws-skeleton/che-plugin-dev-tooling/commit/404b847411be0fd5f82420f8974db32f9324b031). Will provide a PR later today.

PR #1021, opened 2018-10-26
Start workspaces like in upstream tests

Katka92:
### What does this PR do?
Rh-che specific tests uses che-starter for handling workspaces. To be able to run upstream tests, it is needed to handle workspaces directly (without che-starter). This PR adopts upstream behaviour.

localy


functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
```diff
@@ -23,19 +24,24 @@
 public class WorkspaceTest {
 
   private static final Logger LOG = LoggerFactory.getLogger(TestTestClass.class);
+  // When creating workspace with che-starter, all others are stopped. Therefore it is better to
```
rhopp: This comment should not be true anymore - https://github.com/redhat-developer/che-starter/pull/313
(at least on prod-preview)

Katka92: Good point, I've created issue for that. https://github.com/redhat-developer/che-functional-tests/issues/386

PR #1033, opened 2018-10-31
Implement custom templates for running tests on different stacks

ScrewTSW:
### What does this PR do?
Adds functionality to switch between test stacks by using an annotation
Enables test cases BayesianPom and BayesianJson test cases
Adds a new test case for building and running a project.

~~This PR needs to wait for https://github.com/eclipse/che/pull/11792 to be merged.~~ MERGED
~~This PR is waiting for the upstream master to be ported downstream~~ PORTED

https://app.zenhub.com/workspace/o/redhat-developer/che-functional-tests/issues/294

manual testing.


functional-tests/pom.xml
```diff
@@ -51,17 +51,20 @@
...
                 <groupId>org.eclipse.che.selenium</groupId>
                 <artifactId>che-selenium-core</artifactId>
-                <version>${che.version}</version>
+                <version>6.14.0-SNAPSHOT</version>
+                <!--<version>${che.version}</version>-->
             </dependency>
             <dependency>
                 <groupId>org.eclipse.che.selenium</groupId>
                 <artifactId>che-selenium-test</artifactId>
-                <version>${che.version}</version>
```
garagatyi: Is it just for testing purposes or you are going to merge it?

ScrewTSW: That depends on whether we will be able to have those chances backported, or if we will have to wait until rh-che goes to 6.14.0

ibuziuk: I believe it would be better to wait. Making our build snapshot dependent does not sound like a good idea. If you need some particular changes from upstream you could also request a service release and we would update rh-che to `6.13.1` which would contain all changes required for this PR

rhopp: There was no plan from our side to depend on SNAPSHOT versions... For every change in upstream we depend on, we should wait for proper release which includes the change and then make relevant chagnes in our code.

We may (temporarily) break this rule for some extra urgent cases, but this one is not such case.

If I understand this PR correctly, it's just for collecting early feedback - and in that case, it's missing "WIP" label in description.

garagatyi:
@ScrewTSW please take a look whether CodeFactor CI shows any interesting suggestions on the code cleanness

ScrewTSW:
> @ScrewTSW please take a look whether CodeFactor CI shows any interesting suggestions on the code cleanness

Well, the stuff CodeFactor is complaining about is code duplication between test classes. I don't think it's a good idea to do a generalization on a test class.

garagatyi:
> > @ScrewTSW please take a look whether CodeFactor CI shows any interesting suggestions on the code cleanness
> 
> Well, the stuff CodeFactor is complaining about is code duplication between test classes. I don't think it's a good idea to do a generalization on a test class.

No pressure. Just saw him complaining about something. I added that CI to check whether it adds any value to the project, we don't have to follow any rules about it.

PR #1044, opened 2018-11-05
Add sys adm env

riuvshin:
Add possibility to define system admin

### DIFF
--- a/openshift/rh-che.config.yaml
+++ b/openshift/rh-che.config.yaml
@@ -32,6 +32,7 @@ data:
   che-openshift-precreate-subpaths: "false"
   che-openshift-secure-routes: "true"
   che.predefined.stacks.reload_on_start: "true"
+  che-system-admin-name: "admin"
   che-secure-external-urls: "true"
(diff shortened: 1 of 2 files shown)


ibuziuk:
[test]

openshift/rh-che.config.yaml
```diff
@@ -32,6 +32,7 @@ data:
   che-openshift-precreate-subpaths: "false"
   che-openshift-secure-routes: "true"
   che.predefined.stacks.reload_on_start: "true"
+  che-system-admin-name: "admin"
```
garagatyi: Do we have a user with this name already?

ibuziuk: I believe it would be applied via cm "on dsaas / dsaas-stg"

garagatyi: yeah, you are right. It doesn't have to match this one

ibuziuk:
[test]

openshift/rh-che.config.yaml
rhopp: Is it a problem, if such user does not exist? If so, it could be a problem for PR check deployment.

ibuziuk: no, I believe there should not be any problems if the user does not exist - no admin user is going to be assigned.

riuvshin: no, but I set it to be matching [this](https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties#L27)

Issue #1013, opened 2018-10-24
Remove unnecessary authentication steps in client-side IDE loading on openshift.io


ibuziuk:
Need to understand what makes Theia IDE loading slow on openshift.io and provide a solution that would improve Theia IDE startup time. Even when workspace is already started switching from dashboard to workspace requires sufficient amount of time:

![image](https://user-images.githubusercontent.com/1461122/47447097-0d7eba80-d7bd-11e8-825a-374e90737350.png)

ibuziuk:
Some notes from @garagatyi:

>  FYI I did some investigation before the conference, so here are somethings to consider. UD requests all the workspaces by ID. If I'm not mistaken it also does it 2 times (get all and  get one by one). The same for factories even though factories view is not opened. You can also check whether it fetches all the stacks when it is not needed.
> And it might be useful to check time needed to load fonts and other pages resources since they are not downloaded from a CDN.

davidfestal:
I've changed the description of the issue and will close it since the 2 related PRs removed the unnecessary authentication steps that were taking between 5 and 10 seconds, and are not necessary when starting the IDE frame in the context of the User Dashboard.
Most of the network calls shown in the [issue description image](https://user-images.githubusercontent.com/1461122/47447097-0d7eba80-d7bd-11e8-825a-374e90737350.png) were related to these unnecessary authentication steps.

However, the analysis work done in this issue also resulted in the design of an additional solution for optimizing Theia IDE loading time, by allowing the Theia main JS bundle files to be:
- cached correctly (which implies having them available from a fixed URL instead of the always-changing workspace route URL)
- retrieved from a configured CDN when available from there, with a fallback to local versions (which implies changing the HTML page accordingly)
- pre-loaded into the browser cache from the HTML page of the workspace loader (= previous step), so that the first client-side loading and caching of these big JS files would occur while the Theia container is still starting on the server-side.

However the work involved in the implementation of this caching support will be covered by new issues that will be added in the following Epic: https://github.com/redhat-developer/rh-che/issues/1011

ibuziuk:
@davidfestal am I correct that issue is planned to be closed once prod will be updated to 6.13.1 / 6.14.0 version ?

davidfestal:
yes.

ibuziuk:
prod has been updated to 6.13.1. Closing

PR #1055, opened 2018-11-09
Optimize IDE loading - End-2-end provisioning flow enhancements

davidfestal:
### What does this PR do?

This PR reduces a bit IDE load times on the client-side with the following changes:
 - Only run end-2-end provisioning flow during the dashboard authentication, and skip it when the authentication is done for IDEs, workspace loader or JWT proxy loader page.
This avoids a number of verification steps involved in the end-2-end provisioning flow that are not necessary for IDE frames since, to come into the IDE window, a user should come from either the Dashboard or OSIO, and so would have been already provisioned there.
- Add a 2 minutes `max-age` header to the end-2-end flow and authentication Javascript resources.
This allows caching to memory and avoids having to reload the `RhCheKeycloak.js` and `OIDCKeycloak.js` files when the page is refreshed after the authentication redirection.

These issues are part of the work done in the context of issue https://github.com/redhat-developer/rh-che/issues/1013

Yes, on Dev cluster.


davidfestal:
@benoitf should we wait for your approval, or can we go on and merge ?

PR #1063, opened 2018-11-13
Changes necessary to make minishift addon work with OpenShift 3.11.0

amisevsk:
Changes necessary to make the rhche minishift addon work with the
current version of minishift, which supports OpenShift 3.10.0+.

This commit breaks the addon for OpenShift 3.9.0 and below, due to some
changes on the minishift side in the move.

### What does this PR do?
Modifies the rhche minishift addon so that it works with OpenShift 3.11.0 (tested) and 3.10.0 (assumed).

Note that the current release of minishift uses 3.11.0 by default, and 3.9.0 is no longer supported. If you have an older VM running 3.9.0 it is probably necessary to `minishift delete` and probably `rm -rf ~/.minishift`.

Locally, with
```
$ minishift version
minishift v1.26.1+1e20f27

$ oc version
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://192.168.42.164:8443
kubernetes v1.11.0+d4cacc0
```


amisevsk:
Note, the SA issue we were having is on me -- I copied it over from the regular template in #1048

PR #1078, opened 2018-11-16
Fix nodejs stack run command

jarifibrahim:
### What does this PR do?
In order to run the nodejs boosters (eg https://github.com/nodeshift-starters/nodejs-rest-http-redhat) the `Run` command has to be changed.
The previous command would look for `app/app.js` file which would fail since `app.js` is in the root directory.

I have tested this PR by importing the source code in che and then changing the `RUN` command manually.


ibuziuk:
@jarifibrahim if we merge this PR, it will break existing quickstarts. I would recommend to contribute the https://github.com/nodeshift-starters/nodejs-rest-http-redhat to existing upstream samples [1]

[1] https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json#L405

sbose78:
Thanks @ibuziuk . 

A couple of questions:
1. Does the json file get hot loaded ? Or do we need a new version upgrade of Che to reflect the changes?
2. The booster would be opened in Che by a call from WIT to che-starter. Would having the booster in the list https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json imply that all workspaces triggered from the WIT codebase API would have the custom `run` command?

ibuziuk:
> Does the json file get hot loaded ? Or do we need a new version upgrade of Che to reflect the changes?
We need new version of Che

> The booster would be opened in Che by a call from WIT to che-starter. Would having the booster in the list https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json imply that all workspaces create from the WIT codebase 
No, run command would be defined only based on stack. 

@eivantsov I started to think that probably we should use `cd ${current.project.path} && node .` as a default command for node.js stack upstream and update samples accordingly if needed. For me it sounds like `app/app.js` is the default stack command makes it really project dependent. WDYT ?

sbose78:
> No, run command would be defined only based on stack.
Ok, so there's no way to override that for a specific project/codebase?

eivantsov:
@ibuziuk yes, I think a macro with current project path should be used

sbose78:
@eivantsov  ,
`cd ${current.project.path} && node .` 

Are you saying that we start using this as the `run` command and update che samples accordingly?

eivantsov:
@sbose78 yes, sound ok to me

jarifibrahim:
> @eivantsov ,
> `cd ${current.project.path} && node .`
> 
> Are you saying that we start using this as the `run` command and update che samples accordingly?

@sbose78 We have a lot of boosters in https://github.com/nodeshift-starters . Should we add these to the samples.json https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json ?

sbose78:
@jarifibrahim , 

Phase 1:
- Fix the che-samples to use the new run command
- One by one: get the missions working , and enabled in launcher backend.

Phase 2.
- Adopt all boosters as che-samples.

jarifibrahim:
> @jarifibrahim ,
> 
> Phase 1:
> 
> * Fix the che-samples to use the new run command

@sbose78 The existing che samples work fine with the existing command. We don't need to change them.
`node app/app.js` would be the correct command for the already existing samples

sbose78:
> The existing che samples work fine with the existing command. 

@jarifibrahim Yes, but they don't with the new command. That's what we would need to fix.

ibuziuk:
@jarifibrahim thanks, I will wait until https://github.com/eclipse/che/pull/11983/files is merged and than merge this one

ibuziuk:
@jarifibrahim could you please also send a PR for updating default command for node.js stack upstream ?

jarifibrahim:
> @jarifibrahim could you please also send a PR for updating default command for node.js stack upstream ?

@ibuziuk Can you point me to the repository, please?

ibuziuk:
@jarifibrahim https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L2024

jarifibrahim:
I've raised https://github.com/eclipse/che/pull/11989 to fix the `Run` command in upstream.

ibuziuk:
@jarifibrahim thanks, I'm going to merge this PR once eclipse/che#11989 is applied

Issue #1075, opened 2018-11-15
Update to 6.14.0 on prod-preview break usage of Che SA token required for che-* namespace lockdown and introduced issue with k8s client usage


ibuziuk:
`6.14.0` upstream version of che have updated the version of kubernetes-client form `3.1.12` to `4.1.0` break usage of Che SA token that is required for removing edit rights from *-che namespace [1]. Currently all prod-preview accounts have been switched to use OSIO token instead of Che SA token via toggle. Currently using Che SA token for communication with oso-proxy results in the following errors which prevents creation of resources in *-che namespace:

> {"@timestamp":"2018-11-15T14:07:53.744+00:00","@version":1,"message":"Workspace 'rh-che-automation-preview/wksp-muq9' with id 'workspacetmpa1qatnigx0kix' created by user 'rh-che-automation-preview'","logger_name":"org.eclipse.che.api.workspace.server.WorkspaceManager","thread_name":"http-nio-8080-exec-6","level":"INFO","level_value":20000,"identity_id":"0c08fee4-5bae-4063-ad65-091bb80a9df5"}
> {"@timestamp":"2018-11-15T14:08:01.444+00:00","@version":1,"message":"Starting workspace 'rh-che-automation-preview/wksp-muq9' with id 'workspacetmpa1qatnigx0kix' by user 'rh-che-automation-preview'","logger_name":"org.eclipse.che.api.workspace.server.WorkspaceRuntimes","thread_name":"http-nio-8080-exec-2","level":"INFO","level_value":20000,"identity_id":"0c08fee4-5bae-4063-ad65-091bb80a9df5"}
[...]

PR #1083, opened 2018-11-19
rh-che #1075: Returning back passing identity_id via query parameter to oso-proxy

ibuziuk:
### What does this PR do?
rh-che #1075: Returning back passing identity_id via query parameter to oso-proxy

https://github.com/redhat-developer/rh-che/issues/1075


ibuziuk:
[test]

PR #1086, opened 2018-11-21
Add new nodejs-postgres stack

jarifibrahim:
### What does this PR do?
In order to run multi-container booster in che (https://github.com/nodeshift-starters/nodejs-rest-http-crud) we need to add a new stack which has 2 machines, first one is the dev machine with node and the second machine has the postgres database.

N/A

Yes. I have tested this stack locally and it works as expected.


jarifibrahim:
The nodejs stack is failing with
```
Loading...
pulling image "eclipse/node"
Successfully pulled image "eclipse/node"
Created container
Started container
pulling image "registry.centos.org/postgresql/postgresql:9.6"
Successfully pulled image "registry.centos.org/postgresql/postgresql:9.6"
Error: Error response from daemon: create bf02f38276729574b6586e049129587cdeb36a951e4d1d38129ec8b8bb8a955d: error while creating volume path '/var/lib/docker/volumes/bf02f38276729574b6586e049129587cdeb36a951e4d1d38129ec8b8bb8a955d/_data': mkdir /var/lib/docker/volumes/bf02f38276729574b6586e049129587cdeb36a951e4d1d38129ec8b8bb8a955d: permission denied
```

How can I fix these permissions?

jarifibrahim:
> @jarifibrahim could you switch to `quay.io/openshiftio/che-centos-nodejs` ?

Done. See https://github.com/redhat-developer/rh-che/pull/1086/commits/ba45c6f8848f6e1d8f3b056222f5026ebcece886

ibuziuk:
[test]

jarifibrahim:
@ibuziuk I've changed the docker image for the Postgres and that's why the build is failing.

ibuziuk:
[test]

ibuziuk:
@jarifibrahim I do not think it is related

jarifibrahim:
I see the same error even after changing the Postgres docker image.

ibuziuk:
@jarifibrahim could you please point to the dockerfile of the postgres image used in the stack ?

jarifibrahim:
@sbose78 @ibuziuk Looks like volume mount doesn't work for mysql container as well https://github.com/sclorg/mysql-container/blob/master/5.7/Dockerfile#L73 because of https://github.com/sclorg/httpd-container/issues/30.

What else can we do to get Postgres container running?

sbose78:
Does this image work `registry.centos.org/postgresql/postgresql:9.6` ?
I had use it in a recipe, from which I had created a workspace on che.openshift.io https://gist.github.com/sbose78/5c4ad4e3fd2fea7efb4efb3849ecdea4#file-gistfile1-txt-L24

jarifibrahim:
According to https://github.com/openshift/origin/issues/16211#issuecomment-350819976 we need to mount the volume on the specific path. @ibuziuk Is there a way to do this? I tried adding a volume via UI and updating the recipe but it doesn't seem to work.

jarifibrahim:
I've raised https://github.com/sclorg/postgresql-container/pull/301 to remove volume from the centos-postgresql image.

jarifibrahim:
I managed to fix the volume issue using the correct PV claim https://github.com/redhat-developer/rh-che/pull/1086/commits/dc8659f30e686000ec8d2bd7d8a567f369f92a68 . 
The stack works out of the box with https://github.com/nodeshift-starters/nodejs-rest-http-crud now :tada:

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -64,6 +64,104 @@
...
+      }
+    ],
+    "workspaceConfig": {
+      "defaultEnv": "default",
+      "environments": {
+        "default": {
+          "machines": {
+            "database": {
+              "attributes": {
+                "memoryLimitBytes": "1147483648"
```
jarifibrahim: I've assigned 1.1GB of memory. Not sure if we need less/more memory.

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -64,6 +64,104 @@
...
+              "env": {
+                "POSTGRESQL_ADMIN_PASSWORD": "adminpassword",
+                "POSTGRESQL_DATABASE": "my_data",
+                "POSTGRESQL_USER": "user",
+                "POSTGRESQL_PASSWORD": "password"
+              }
+            },
+            "nodejs": {
+              "attributes": {
+                "memoryLimitBytes": "1147483648"
```
jarifibrahim: I've assigned 1.1GB of memory. Not sure if we need less/more memory.

jarifibrahim:
@ibuziuk Can you please review this PR?

assembly/fabric8-stacks/src/main/resources/stacks.json
sbose78: 1.1GB is a lot. How much does the booster need?

jarifibrahim: It takes about 870mb on `npm install && node .`

ibuziuk:
@jarifibrahim great, that the sample is finally working! A few questions:
- are you planning to contribute this sample upstream ?
- what happens if you create a few workspaces against this stack ? I suspect that data in `/var/lib/pgsql/data` will be corrupted.

jarifibrahim:
> @jarifibrahim great, that the sample is finally working! A few questions:
> 
> * are you planning to contribute this stack upstream ?

I could. Do you suggest I should contribute this upstream?

> * what happens if you create a few workspaces against this stack ? I suspect that data in `/var/lib/pgsql/data` will be corrupted.

I guess it would corrupt it since the volume would be shared. Is there a way to  make this workspace specific?

ibuziuk:
> I could. Do you suggest I should contribute this upstream?

yes, I believe it would be great to contribute this stack upstream.

> I guess it would corrupt it since the volume would be shared. Is there a way to make this workspace specific?

I do  not have a good idea how this could be done in a sustainable way. @eivantsov maybe you could comment ? Also, wondering why no volume is required for mysql db stack ?

jarifibrahim:
> I do not have a good idea how this could be done in a sustainable way. @eivantsov maybe you could comment ? Also, wondering why no volume is required for mysql db stack ?

See https://github.com/sclorg/mysql-container/blob/master/5.7/Dockerfile#L73 . The mysql stack doesn't mount a volume. Once the pod is removed, the data would be lost.

ibuziuk:
@jarifibrahim so, we should probably opt for the same approach with not mounting volume for postgresql. WDYT ?

jarifibrahim:
@ibuziuk A database should have a persistent volume. Pods are often killed and removed. A user would lose data if don't use a persistent volume :)

jarifibrahim:
@amisevsk Thanks for the thorough review. I have a couple of questions
1. If we can't (don't) want to mount a volume, how can we use the `PostgresSQL` image? The container will try to mount a volume (because of [this](https://github.com/sclorg/postgresql-container/blob/generated/9.6/Dockerfile#L68)) and fail. Since, removing the `volume` declaration might now be possible from the original image, shall I create a new `Dockerfile` in `che-dockerfiles` repo which is **exactly same** as [centos/postgresql-96-centos7](https://github.com/sclorg/postgresql-container/tree/generated/9.6) except the volume declaration and use the new one?

amisevsk:
@jarifibrahim Sorry, I didn't fully grok the discussion above. One workaround for the time being would be to instead use an emptyDir volume instead of `claim-che-workspace`, e.g.:
```
     volumeMounts:
      - 
       mountPath: /var/lib/pgsql/data
       name: ephemeral-che-postgres
   volumes:
    - 
     name: ephemeral-che-postgres
     emptyDir: {}
```

That *should* work, but I haven't tested it.

l0rd:
A couple of considerations:
- I think we have a limit of 2 PV per che namespace on osio and we are currently using only one. We may use the second PV for these use cases.
- We should be consistent with what is documented https://www.eclipse.org/che/docs/che-6/volumes.html#Adding-volumes

Independently from this PR I think we should document how to persist some folders (outside /projects) on an osio che workspace.

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -140,7 +144,7 @@
...
+            "content": "kind: List\nitems:\n - \n  apiVersion: v1\n  kind: Pod\n  metadata:\n   name: nodejs-postgres-pod\n   annotations:\n    org.eclipse.che.container.nodejs.machine_name: nodejs\n    org.eclipse.che.container.database.machine_name: database\n  spec:\n   containers:\n    - \n     image: quay.io/openshiftio/che-centos-nodejs\n     name: nodejs\n     ports:\n      - \n       containerPort: 8080\n       protocol: TCP\n     resources:\n      limits:\n       memory: 1024Mi\n    - \n     image: centos/postgresql-96-centos7\n     name: database\n     ports:\n      - \n       containerPort: 5432\n       protocol: TCP\n     resources:\n      limits:\n       memory: 512Mi\n     env:\n      - \n       name: POSTGRESQL_USER\n       value: user\n      - \n       name: POSTGRESQL_PASSWORD\n       value: password\n      - \n       name: POSTGRESQL_DATABASE\n       value: my_data\n      - \n       name: POSTGRESQL_ADMIN_PASSWORD\n       value: adminpassword\n",
```
amisevsk: It seems as though volume names must be lowercase - I get the error `Volume name 'postgresData' in machine 'database' is invalid` using this stack, but no error if I change the volume name to `postgresdata`. Could you please update this value?

amisevsk: Just realized I put this comment at the wrong line -- should be on 109.

jarifibrahim:
Thank you so much @amisevsk. That fix worked perfectly :)

assembly/fabric8-stacks/src/main/resources/stacks.json
```diff
@@ -64,6 +64,108 @@
...
+                "memoryLimitBytes": "1147483648"
+              },
+              "servers": {
+                "node": {
+                  "attributes": {},
+                  "port": "8080",
+                  "protocol": "http"
+                }
+              },
+              "volumes": {},
```
jarifibrahim: @ibuziuk @amisevsk Do we need to mount a volume here for the user data? Like `/projects` ?

ibuziuk: no, project is mounted by default. No need to explicitly specify if

jarifibrahim:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

ibuziuk:
@rhopp @ScrewTSW @Katka92 seems to be failing with:

> OpenShift username and password and Keycloak token are not empty.
> Tests are executed with user: rh-ch e-automation-preview
> OSO_MASTER_URL found and set to: https://api.starter-us-east-2a.openshift.com
> Active token has 1 sections.
> JWT token parse failed!

Any ideas what is the reason ?

Katka92:
I think that the problem is in a way obtaining active token for test user. I'll have a look and try to fix it ASAP.

ScrewTSW:
@ibuziuk Fix for the issue. I have deleted keycloak token from the vault as it's not being used anymore. Sorry for the issues caused
https://github.com/redhat-developer/rh-che/pull/1137

ibuziuk:
[test]

PR #1097, opened 2018-11-26
Remove usage of offline token

Katka92:
### What does this PR do?
As offline token was mandatory, script expected it as a mandatory parameter. Now this token is not used at all.


rhopp:
I'm fine with this quickfix... Last week I've created issue to remove all occurrences of offline token & offline token endpoint: https://github.com/redhat-developer/che-functional-tests/issues/403 But we can tackle that one in another PR.

Katka92:
Good to know. Thank you for sharing :+1:

PR #1105, opened 2018-11-29
Implement custom templates for running tests on different stacks

ScrewTSW:
### What does this PR do?
Adds functionality to switch between test stacks by using an annotation

~~This PR needs to wait for eclipse/che#11792 to be merged.~~ MERGED
~~This PR is waiting for the upstream master to be ported downstream~~ PORTED

https://app.zenhub.com/workspace/o/redhat-developer/che-functional-tests/issues/294

manual testing.


functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
```diff
@@ -27,10 +28,14 @@
...
-  @InjectTestWorkspace(startAfterCreation = false)
+  @InjectTestWorkspace(
+      startAfterCreation = false,
+      template = RhCheWorkspaceTemplate.RH_STARTER_DEFAULT)
   private RhCheTestWorkspaceImpl workspaceToPass;
 
-  @InjectTestWorkspace(startAfterCreation = false)
+  @InjectTestWorkspace(
+      startAfterCreation = false,
+      template = RhCheWorkspaceTemplate.RH_STARTER_NODEJS)
```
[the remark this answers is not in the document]
ScrewTSW: It was mostly for testing purposes, whether che-starter can create a workspace with project successfully. I don't think this hurts the test speed/performance in any significant way

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/RhCheWorkspaceTemplate.java
```diff
@@ -0,0 +1,23 @@
...
+package com.redhat.che.selenium.core.workspace;
+
+public final class RhCheWorkspaceTemplate {
+
+  // Upstream way
+  public static final String RH_DEFAULT = "vertx_default.json";
+  public static final String RH_NODEJS = "nodejs_default.json";
+
+  // Che-starter way
+  public static final String RH_STARTER_DEFAULT = "che-starter_vertx.json";
```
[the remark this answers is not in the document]
ScrewTSW: I can rename them, but is this really necessary as VERT.X is used by default?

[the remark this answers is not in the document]
Katka92: That information is common for us, not for somebody else who can look into our tests. If I didn't know the code, I would say that "default" should be blank or something like that. WDYT @rhopp?

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
```diff
@@ -40,7 +40,7 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(RhCheTestWorkspaceServiceClient.class);
   private static final String CREATE_WORKSPACE_REQUEST_JSON_PATH =
-      "/configs/create-workspace-request.json";
+      "/requests/workspace/osio/che-starter_vertx.json";
```
rhopp: Is there a reason why you chose `/requests` path instead of `/templates/workspace` as is used in upstream? Wouldn't make sense to put json templates for using with che-starter into `/templates/workspace/che-starter` or something like that? Currently I'm not entirely sure which way I prefer, but I'm slightly leaning towards `/templates/...` side.

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
```diff
@@ -73,8 +73,18 @@ public RhCheTestWorkspaceServiceClient(
...
+      throw new IllegalStateException("Workspace does not have an owner.");
+    }
+    this.cheStarterWrapper.checkIsRunning(this.token);
+    String name =
+        this.cheStarterWrapper.createWorkspace(
+            "/requests/workspace/osio/" + templateFileName, token);
+    return requestFactory
+        .fromUrl(getNameBasedUrl(name, owner.getName()))
+        .request()
+        .asDto(WorkspaceDto.class);
```
rhopp: This whole method seems to me like a duplication of code (see method `createWorkspace(String, int, MemoryMeasure, WorkspaceConfigDto)`)

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/RhCheWorkspaceTemplate.java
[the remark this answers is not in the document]
rhopp: Yep... I do think it should be obvious from the name, what it does... Thus putting vert.x into the var name seems like a good step to me.

functional-tests/src/test/resources/suites/simpleTestSuite.xml
```diff
@@ -24,7 +24,7 @@
             <!--<class name="org.eclipse.che.selenium.dashboard.CreateAndDeleteProjectsTest"/>-->
             <!--<class name="org.eclipse.che.selenium.dashboard.RenameWorkspaceTest"/>-->
             <class name="org.eclipse.che.selenium.miscellaneous.DialogAboutTest"/>
-            <!--<class name="com.redhat.che.functional.tests.WorkspaceTest"/>-->
+            <class name="com.redhat.che.functional.tests.WorkspaceTest"/>
```
rhopp: It would be nice to have changes not related to the issue you are trying to solve in different PRs.

functional-tests/src/test/resources/templates/workspace/openshift/default.json
```diff
@@ -1,41 +0,0 @@
-{
```
rhopp: I'm not sure, but don't we need this file to run some basic upstream tests? (The ones which requires default workspace?)

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
ScrewTSW: I've put it into `/requests` because it's effectively a request template for che-starter. this is not contacting the master API directly

ScrewTSW: Ok, I can move the location to `/templates/che-starter` rather than `/templates/workspace/che-starter` because the `workspace` directory contains a `JSON` version of the `YAML` DC. The che-starter templates just don't fit there.

functional-tests/src/test/resources/suites/simpleTestSuite.xml
ScrewTSW: Well, the test is using the che-starter templates, which are being injected by the `@Workspace` annotation. It's effectively testing my changes, so I thought it wouldn't be a bad idea to enable it.

functional-tests/src/test/resources/templates/workspace/openshift/default.json
ScrewTSW: the file was renamed and moved to `vertx_default.json`

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
rhopp: files in `/templates/workspace/<infra>` are just workspace configs... json files describing workspaces to be created... I think, that adding `che-starter` into the workspace folder, it is more fitting, as it contains json files describing what workspace should be cerated ;-)

ScrewTSW: :+1:

functional-tests/src/main/java/com/redhat/che/selenium/core/client/RhCheTestWorkspaceServiceClient.java
ScrewTSW: This method allows us to use a template. `createWorkspace` doesn't take a template as a parameter (because it overrites super class method) and it uses a vert.x stack by default.

ScrewTSW:
LEEERRRROYYY JENKIIINS!!!

PR #1107, opened 2018-11-29
Implement PackageJSON test case

ScrewTSW:
### What does this PR do?
*Adds Bayesian test for JSON-based projects
*Refactors BayesianPomXml test case

PR depends on https://github.com/redhat-developer/rh-che/pull/1105 - MUST BE MERGED FIRST

https://app.zenhub.com/workspaces/osio-che-team-59411dba86156437e0727c12/issues/redhat-developer/che-functional-tests/294

manual testing against minishift and prod-preview

### DIFF
--- a/functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
+++ b/functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
@@ -30,11 +30,11 @@ import org.testng.annotations.BeforeClass;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
-public class BayesianPomXml {
+public class BayesianPackageJson {
 
-  private static final Logger LOG = LoggerFactory.getLogger(TestTestClass.class);
+  private static final Logger LOG = LoggerFactory.getLogger(BayesianPackageJson.class);
 
-  @InjectTestWorkspace(template = RhCheWorkspaceTemplate.RH_VERTX)
+  @InjectTestWorkspace(template = RhCheWorkspaceTemplate.RH_NODEJS)
   private TestWorkspace workspace;
 
   @Inject private NavigateToFile navigateToFile;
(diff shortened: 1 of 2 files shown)


functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
```diff
@@ -0,0 +1,136 @@
...
+    editor.closeAllTabs();
+  }
+
+  @Test
+  public void createBayesianError() {
+    try {
+      editor.waitTextInToolTipPopup(ERROR_MESSAGE);
+    } catch (Exception e) {
+      LOG.error(
+          "Bayesian error not present after adding dependency - known issue for prod-preview.");
```
Katka92: Would you like to show this error even if this fails on prod? I don't see any switch or check if the test runs on prod-preview. Same for ```checkErrorPresentAfterReopenFile``` method.

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
```diff
@@ -0,0 +1,136 @@
...
+      return;
+    }
+    LOG.info("Bayesian error message was present after adding dependency.");
+  }
+
+  @Test
```
Katka92: Are you sure this test would ALWAYS run as a seconds one? I would suggest to add ```priority``` here to be sure about an order as far as it depends on ```createBayesianError```. @rhopp are you OK that these methods aren't atomics?

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPomXml.java
```diff
@@ -21,68 +24,91 @@
...
+  }
+
+  @AfterMethod
+  public void closeFiles() {
+    removeDependency();
+    editor.closeAllTabs();
   }
 
-  @Test(priority = 2)
+  @Test
```
Katka92: Why did you remove priority from here?

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
ScrewTSW: fixed

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
ScrewTSW: Those tests are re-written to be atomic. the order doesn't matter anymore.

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPomXml.java
ScrewTSW: Both tests were re-written to be atomic, the order therefore doesn't matter

PR #1125, opened 2018-12-05
Temporarily disabling WorkspaceTest for periodical tests

ScrewTSW:


ibuziuk:
@ScrewTSW could do explain in the PR description why are you planning to disable the test ?

ScrewTSW:
> @ScrewTSW could do explain in the PR description why are you planning to disable the test ?

The test has been consistently failing for the past 4 builds and the periodical job only runs twice a day. We would like to see whether the rest of our suite is stable or not.

rhopp:
@ScrewTSW I personally woudn't exclude this testcase right now... I would leave this here and try to figure out what's causing that. This does not prevent other tests from running AFAIK. 
We should prioritize archiving logs, so we can take a look at che-starter logs and continue from there.

ScrewTSW:
closing

PR #1132, opened 2018-12-07
Implement archiving artifacts for periodic tests (new suite)

ScrewTSW:
# What does this PR do?

Adds artifacts collection (test logs, che-starter logs, test screenshots) for periodic tests running our new suite

# What issues does this PR fix or reference?

https://github.com/redhat-developer/che-functional-tests/issues/382

# How have you tested this PR?
manual testing with modified prcheck

### DIFF
--- a/.ci/prepare_env_utils.sh
+++ b/.ci/prepare_env_utils.sh
@@ -42,7 +42,7 @@ function checkAllCreds() {
 	CREDS_NOT_SET="false"
 	curl -s "https://mirror.openshift.com/pub/openshift-v3/clients/${OC_VERSION}/linux/oc.tar.gz" | tar xvz -C /usr/local/bin
 
-	if [[ -z "${QUAY_USERNAME}" && -z "${QUAY_PASSWORD}" ]]; then
+	if [[ -z "${QUAY_USERNAME}" || -z "${QUAY_PASSWORD}" ]]; then
 	  echo "Docker registry credentials not set"
 	  CREDS_NOT_SET="true"
 	fi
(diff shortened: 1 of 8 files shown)


.ci/prepare_env_utils.sh
```diff
@@ -67,7 +68,16 @@ function checkAllCreds() {
 	set +x
 }
 
-
-
-
-
+# JOB_TYPE=che-functional-tests|functional-tests|periodic-tests
```
Katka92: Now we have three jobs, so I would suggest something like ```rh-che/${JOB_NAME}``` - to have periodic and pr-check in one rh-che folder. There are e2e and bdd test so I would avoid creating many rh-che related folders in that level of hierarchy.

.ci/prepare_env_utils.sh
```diff
@@ -67,7 +68,16 @@ function checkAllCreds() {
...
+function archiveArtifacts() {
+  echo "Archiving artifacts from ${DATE} for ${JOB_NAME}/${BUILD_NUMBER}"
+  ls -la ./artifacts.key
+  chmod 600 ./artifacts.key
+  chown $(whoami) ./artifacts.key
+  mkdir -p ./${JOB_TYPE}/${JOB_NAME}/${BUILD_NUMBER}/surefire-reports
+  #cp -R ./logs/artifacts/surefire-reports/*.txt ./${JOB_TYPE}/${JOB_NAME}/${BUILD_NUMBER}/surefire-reports
```
Katka92: If there are no surefire reports maybe we should remove these two lines.

functional-tests/devscripts/run_tests.sh
```diff
@@ -78,6 +88,9 @@ else
     RESULT=$?
 fi
 
+JOB_TYPE="periodic-tests"
+archiveArtifacts
```
Katka92: What about moving archiving artifacts at the end of cico_rhche_prcheck.sh? It would make sense from a flow view - firstly tests take place, secondly artifacts are sent. And it wouldn't be needed to source ```prepare_env_utils.sh``` here.

.ci/prepare_env_utils.sh
```diff
@@ -67,7 +68,16 @@ function checkAllCreds() {
...
+  cp ./logs/*.log ./${JOB_TYPE}/${JOB_NAME}/${BUILD_NUMBER}/
+  cp -R ./logs/artifacts/screenshots/ ./${JOB_TYPE}/${JOB_NAME}/${BUILD_NUMBER}/
+  #cp -R ./logs/artifacts/videos/ ./${JOB_TYPE}/${JOB_NAME}/${BUILD_NUMBER}/
```
Katka92: I don't see any advantage of this comment - as far as I understand we are not planning to add videos to tests.

Issue #1116, opened 2018-12-04
Need to test new CORS configuration in 6.15 that will be applied by default in 6.16 release 


ibuziuk:
from @mkuznyetsov email:

We have some news regarding to CORS filter configuration changes in 6.16.

In Che 6.16 we plan to upgrade Tomcat from 8.5.23 to 8.5.35. This however will require us to update our CORS filter configuration for Che WS Master and WS agent, since it would not allow our current CORS configuration which is deemed to be unsecure.
More information about original Tomcat issue is at https://bz.apache.org/bugzilla/show_bug.cgi?id=62343
 
This is the configuration that will be applied for Che CORS filters in Che 6.16:
 
- WS Master - CORS Filter disabled
- WS Agent - CORS Filter enabled, allowing requests with credentials, providing Domain of WS Master as an allowed origin

Before we will merge these changes in 6.16, there will be added options to test this configuraion in Che 6.15 (https://github.com/eclipse/che/pull/12046)
We encourage you to try this configuration in 6.15 version to see if it works for your Che installation.

Here are the steps to enable this configuration, using following environment variables for Che deployment:

1) `CHE_CORS_ENABLED=false` - disable CORS on WS Master.
2) `CHE_WSAGENT_CORS_ALLOWED__ORIGINS=<wsmaster-domain>` - set the default allowed origin of WS Agent CORS pointing to Domain of WS Master (replace `<wsmaster-domain>` with actual WS master domain )

If you discover an regression related to this configuration (for example, certain cross origin requests are not working in your Che installation).
Please reach out to us describing the issues you have, so we may address them before Che 6.16 release.

ibuziuk:
blocked by https://github.com/redhat-developer/rh-che/issues/1110

garagatyi:
LOL see https://github.com/redhat-developer/rh-che/issues/1117
You were 2 minutes ahead!

ibuziuk:
PR for adding env vars which allow to disable cors

ibuziuk:
@rhopp @Katka92 @ScrewTSW CORS are currently disabled on prod-preview 

![image](https://user-images.githubusercontent.com/1461122/49727959-ce73ce00-fc71-11e8-826a-0e9a69ff0dba.png)

Will you be able to test openshift.io, che.openshift.io and e2e registration flow before we would disable CORS on prod ? Also please test functionality of adding custom registry via UD

Katka92:
[...]

PR #1133, opened 2018-12-07
rh-che #1116: Adding env var which allows to disable CORS

ibuziuk:
The plan is the following:

- disable CORS on prod-preview, but not on prod
- test if everything works fine on prod-preview
- disable CORS  on prod

### What does this PR do?
Adding env var which allows to disable CORS

https://github.com/redhat-developer/rh-che/issues/1116

### HK issue
https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2470

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -85,6 +85,11 @@ objects:
               configMapKeyRef:
                 key: che-jsonrpc-max-pool-size
                 name: rhche
+          - name: CHE_CORS_ENABLED
+            valueFrom:
+              configMapKeyRef:
+                key: che-cors-enabled
+                name: rhche
           - name: CHE_DEBUG_SERVER
             valueFrom:
               configMapKeyRef:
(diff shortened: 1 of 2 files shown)


ibuziuk:
Blocked until cm would be updated as part of the https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2470

openshift/rh-che.app.yaml
```diff
@@ -85,6 +85,16 @@ objects:
               configMapKeyRef:
                 key: che-jsonrpc-max-pool-size
                 name: rhche
+          - name: CHE_CORS_ENABLED
+            valueFrom:
+              configMapKeyRef:
+                key: che-cors-enabled
+                name: rhche
+          - name: CHE_WSAGENT_CORS_ALLOWED__ORIGINS
```
amisevsk: (Totally optional) suggestion: should we move this env entry down to keep the env var names alphabetically ordered? It doesn't matter too much for just one yaml but I've found it useful when comparing against the minishift addon template.

ibuziuk: no IMO, since those env var would be removed in the `6.16.0` release - CORS is planned to be always disabled

ibuziuk: ah, sorry last comment was for `Should we also add these values to the minishift template yaml? (I don't know if it's interesting/useful to do testing there)`. Sure thing I will move the env var down ;-)

ibuziuk: @amisevsk should be better now

ibuziuk:
[test]

ibuziuk:
[test]

PR #1135, opened 2018-12-10
Abstract common behavior of test classes.

Katka92:
### What does this PR do?
Add abstraction for test classes. Abstract check for running workspace as is used in almost all test classes as @Before method. Abstract method for opening file.

Tested localy against prod-preview.

### DIFF
--- a/functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
+++ b/functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
@@ -16,6 +16,7 @@ import org.eclipse.che.selenium.core.provider.TestApiEndpointUrlProvider;
 import org.eclipse.che.selenium.core.workspace.TestWorkspace;
 import org.eclipse.che.selenium.pageobject.CodenvyEditor;
 import org.eclipse.che.selenium.pageobject.ProjectExplorer;
+import org.openqa.selenium.TimeoutException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testng.SkipException;
@@ -129,11 +130,15 @@ public abstract class BayesianAbstractTestClass extends RhCheAbstractTestClass {
   }
 
   protected void editorCheckBayesianError() {
-    if (testApiEndpointUrlProvider.get().getHost().equals(CHE_PROD_PREVIEW_URL)) {
-      throw new SkipException("Skipping test for prod-preview - known issue.");
-    }
     editor.setCursorToLine(EXPECTED_ERROR_LINE);
     editor.moveCursorToText(EXPECTED_ERROR_TEXT);
-    editor.waitTextInToolTipPopup(ERROR_MESSAGE);
+    try {
+      editor.waitTextInToolTipPopup(ERROR_MESSAGE);
+    } catch (TimeoutException e) {
+      if (testApiEndpointUrlProvider.get().getHost().equals(CHE_PROD_PREVIEW_URL)) {
+        throw new SkipException("Skipping test for prod-preview - known issue.");
+      }
+      throw e;
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


Katka92:
[test]

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
```diff
@@ -0,0 +1,66 @@
...
+      LOG.error(
+          "Could not obtain workspace name and id - worskape was probably not successfully injected.");
+      throw e;
+    } catch (Exception e) {
+      LOG.error("Could not open workspace IDE.");
+      throw e;
+    }
+  }
+
+  public void openDefinedClass(String projectFile) {
```
rhopp: Couldn't this method be used instead? https://github.com/eclipse/che/blob/bb8e6161b53616aedefad18c066998d7cd0e5c5f/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java#L1092

Katka92: I've removed this method and used upstream one. I decided to use this method: 
https://github.com/eclipse/che/blob/bb8e6161b53616aedefad18c066998d7cd0e5c5f/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java#L1052
Although it requires more code added, it seems more stable when testing localy.

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  protected void removeDependency() {
+    editor.setCursorToLine(INJECTION_ENTRY_POINT);
+    editor.deleteCurrentLine();
+    editor.waitTabFileWithSavedStatus(PROJECT_FILE);
+  }
+
+  protected void editorCheckBayesianError() {
+    if (testApiEndpointUrlProvider.get().getHost().equals(CHE_PROD_PREVIEW_URL)) {
```
rhopp: I don't think this is the right approach... In this case, when bayesian is fixed in prod-preview, this test will be still skipped...
IMHO we want to skip the test only if it fails AND the env is prod-preview.

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPackageJson.java
```diff
@@ -11,139 +11,38 @@
...
+    super.setExpectedErrorLine(JSON_EXPECTED_ERROR_LINE);
+    super.setExpectedErrorText(JSON_EXPECTED_ERROR_TEXT);
+    super.setInjectionEntryPoint(JSON_INJECTION_ENTRY_POINT);
+    super.setProjectFile(PROJECT_FILE);
+    super.setPathToFile(PATH_TO_FILE);
+    super.setErrorMessage(ERROR_MESSAGE);
+    super.setProjectName(PROJECT_NAME);
+    super.setProjectDependency(PROJECT_DEPENDENCY);
+    super.setWorksapce(workspace);
+    super.openTestFile();
```
rhopp: I don't think these `super` are necessary...

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianPomXml.java
```diff
@@ -57,102 +39,30 @@
...
+    super.setExpectedErrorLine(POM_EXPECTED_ERROR_LINE);
+    super.setExpectedErrorText(POM_EXPECTED_ERROR_TEXT);
+    super.setInjectionEntryPoint(POM_INJECTION_ENTRY_POINT);
+    super.setProjectFile(PROJECT_FILE);
+    super.setPathToFile(PATH_TO_FILE);
+    super.setErrorMessage(ERROR_MESSAGE);
+    super.setProjectName(PROJECT_NAME);
+    super.setProjectDependency(PROJECT_DEPENDENCY);
+    super.setWorksapce(workspace);
+    super.openTestFile();
```
rhopp: Same thing.

PR #1136, opened 2018-12-10
Fix compatibility test

Katka92:
### What does this PR do?
Add check for obtaining CHE_VERSION. Fail if version is not found.


Katka92:
[test]

PR #1137, opened 2018-12-10
Hotfix - remove keycloak token logic

ScrewTSW:
### What does this PR do?
Keycloak token is not being used anymore - deleted from vault, removed keycloak token check logic

### DIFF
--- a/.ci/cico_rhche_prcheck.sh
+++ b/.ci/cico_rhche_prcheck.sh
@@ -8,7 +8,7 @@
 export PR_CHECK_BUILD="true"
 export BASEDIR=$(pwd)
 export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
-export OC_VERSION=3.9.33
+export OC_VERSION=3.10.85
 
 eval "$(./env-toolkit load -f jenkins-env.json -r \
         ^DEVSHIFT_TAG_LEN$ \
--- a/.ci/prepare_env_utils.sh
+++ b/.ci/prepare_env_utils.sh
@@ -52,12 +52,13 @@ function checkAllCreds() {
 	  CREDS_NOT_SET="true"
 	fi
 	
-	if [[ -z "${KEYCLOAK_TOKEN}" || -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_USERNAME}" || -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_PASSWORD}" ]]; then
+	if [[ -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_USERNAME}" ]] ||
+	   [[ -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_PASSWORD}" ]]; then
 	  echo "Prod-preview credentials not set."
 	  CREDS_NOT_SET="true"
 	fi
 	
-	if [ "${CREDS_NOT_SET}" = "true" ]; then
+	if [[ "${CREDS_NOT_SET}" = "true" ]]; then
 	  echo "Failed to parse jenkins secure store credentials"
 	  exit 2
 	else


.ci/prepare_env_utils.sh
```diff
@@ -40,6 +40,7 @@ function installDependencies() {
 function checkAllCreds() {
 	set -x
 	CREDS_NOT_SET="false"
+	OC_VERSION="3.10.85"
```
ibuziuk: why specific OC_VERSION is set ?

.ci/prepare_env_utils.sh
```diff
@@ -52,12 +53,13 @@ function checkAllCreds() {
...
 	
-	if [[ -z "${KEYCLOAK_TOKEN}" || -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_USERNAME}" || -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_PASSWORD}" ]]; then
+	if [[ -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_USERNAME}" ]] ||
+	   [[ -z "${RH_CHE_AUTOMATION_CHE_PREVIEW_PASSWORD}" ]]; then
 	  echo "Prod-preview credentials not set."
 	  CREDS_NOT_SET="true"
 	fi
 	
-	if [ "${CREDS_NOT_SET}" = "true" ]; then
+	if [[ "${CREDS_NOT_SET}" = "true" ]]; then
```
ibuziuk: Why extra `[` are used ?

.ci/prepare_env_utils.sh
ScrewTSW: I didn't find any place where it would be set from the config or machine. This is the latest version of the OC binary for `3.10.x` openshift which is running on dev cluster.

.ci/prepare_env_utils.sh
ScrewTSW: It was recommended by my IDE and shellcheck to fix that

.ci/prepare_env_utils.sh
ibuziuk: ok, but why it needs to be defined explicitly ?

ScrewTSW: it's used in the download URL for OC binary. I've discussed it with @Katka92 and found out I've set it at a wrong place :) Will fix it

.ci/cico_rhche_prcheck.sh
```diff
@@ -8,7 +8,7 @@
 export PR_CHECK_BUILD="true"
 export BASEDIR=$(pwd)
 export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
-export OC_VERSION=3.9.33
+export OC_VERSION=3.10.85
```
ibuziuk: @Katka92 @ScrewTSW what would happen if this `OC_VERSION` will not be set ?

ScrewTSW:
[test]

ScrewTSW:
[test]

.ci/cico_rhche_prcheck.sh
ScrewTSW: tests would fail, because the `oc` binary would not be downloaded.

ibuziuk: hmm.. so this is the reason of the current failures https://ci.centos.org/view/Devtools/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/ ?

ScrewTSW: Not really. Those were caused by the `che-functional-tests` docker image we are using in PR-Check.
Some keycloak token verifying logic is hard-coded in the image. I've temporarily restored the token, so the tests don't fail.

ibuziuk: ok, please explain why it was not failing before and oc `binary` were downloaded without this env var

ScrewTSW: because the env var was already there in the prcheck script, but with old version. devcluster was updated to 3.10 from 3.9, so I also updated the version of the binary

ibuziuk: @ScrewTSW should we remove this env var from old prcheck script ? or is it already done ?

PR #1148, opened 2018-12-17
WIP Updating rh-che RAM limit on prod / prod-preview to 5 GB

ibuziuk:
### What does this PR do?
Increasing RAM limit to 5Gb. Currently RAM calculation is the following:
there is a JVM option - `-XX:MaxRAMFraction=2` which means using 50% of the container RAM for heap
Also there is a `che-jsonrpc-max-pool-size` property which is in charge of the maximum size of the json rpc processing pool which is set to `2000`. 

Basically, current RAM consumption is the following:
- 4 GB for container
- 2 GB for heap
- 2 GB for off heap (1Mb for each of the 2000 threads created during json rpc processing)

However, there are also threads that need to be created by tomcat, which are not taken into account. Currently, QA team is planning to do investigation and empirically identify precise value for `che-jsonrpc-max-pool-size` for particular container RAM of che-server. As a temporary solution we want to increase RAM to 5GB for container:

- 2.5 for heap
- 2 for json rpc processing
- 0.5 for other (tomcat threads etc)  

Related issue for QA team - https://github.com/redhat-developer/che-functional-tests/issues/409

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -426,7 +426,7 @@ objects:
             timeoutSeconds: 60
           resources:
             limits:
-              memory: 4Gi
+              memory: 5Gi
             requests:
               memory: 256Mi
         serviceAccountName: rhche


ibuziuk:
cc: @skabashnyuk 
@rhopp could you please point to the QA issue for investigation of the suitable `che-jsonrpc-max-pool-size` connected to OOM investigation

ibuziuk:
@pbergene @jfchevrette could you please comment if we could temporary afford one extra GB on osd ?

ibuziuk:
marking as WIP until getting +1 from SD

pbergene:
I'm having a look. My main concern is less about the total memory available on the cluster, but rather that the total capacity per node is 16GB, so I'm trying to figure out exactly what is reserved for infra, to ensure that it can schedule at all.

ibuziuk:
@pbergene thanks for looking into this. Let's me know if we can not afford it and this PR should be cancelled

pbergene:
@ibuziuk Oh wait. I was looking at the 10GB in title.. that confused me.  Was wondering what I missed and when it started using 9GB.  Correct numbers are 1GB up from the 4GB current limit

ibuziuk:
@pbergene yeah, sorry I do not know why I put 10 in PR description initially (maybe was thinking about both prod / prod-preview)

pbergene:
We are having some challenges on dsaas at the moment, and a failed montoring pod.  We want to determine if this is only monitoring or if there is a cluster issue, so please do not merge this.

ibuziuk:
@pbergene sure, waiting +1 / -1 from SD side before taking any action

ibuziuk:
@pbergene any news ? AFAIK, osd monitor has been stabilized on prod

ibuziuk:
@pbergene am I correct that due to `Christmas shutdown 21.12.18 - 02.01.18` there will not be +1 from SD team and we are going to put this PR on hold ?

pbergene:
Will sync with @jfchevrette once he is online today.  Did you have a chance to talk to him yesterday?

ibuziuk:
@pbergene no, should I ping @jfchevrette today to got clarity about this PR ?

pbergene:
@ibuziuk I'll sync with him once he's on

jfchevrette:
What I'm seeing at the moment is around 2.2GB memory utilisation for rhche.

rhche

A couple of observations from the above:
- rhche is currently using around 9x the memory request value, so I'd recommend increasing that to match average utilization so that rhche can be scheduled on to a node that has enough memory to sustain normal usage
- it's also only using about 54% of the memory limit, so with that in mind increasing the limit wouldn't help much, unless the above 48h sample is not large enough to show higher memory usage from a prior time window.
- CPU seem to cap at 1 core for extended period of time. Looking at the deployment it doesn't have any CPU limits and no defaults are applied to the pods. So I'm wondering if che is leveraging multicore capabilities of the JVM at all? This is not my area of expertise but I would not be surprised to see an increase in memory utilization if the GC is capped and unable to work fast enough to cope with a lot of activity(workspaces).

riuvshin:
@ibuziuk maybe it worth to try to adjust XMX to fit mem limit better

ibuziuk:
>  rhche is currently using around 9x the memory request value, so I'd recommend increasing that to match average utilization so that rhche can be scheduled on to a node that has enough memory to sustain normal usage

@jfchevrette good point, so you think we should put smth. like 2Gb in the memory request value ?

> it's also only using about 54% of the memory limit, so with that in mind increasing the limit wouldn't help much, unless the above 48h sample is not large enough to show higher memory usage from a prior time window.

48h is not enough, you could take a look at zabbix and find that all 4Gb where used which resulted in OOM ome time ago. Increasing to 5Gb should prevent this situations (reasoning in PR description)

>  So I'm wondering if che is leveraging multicore capabilities of the JVM at all

I'm not sure about that, need to sync with Che Platform team regarding this 

> @ibuziuk maybe it worth to try to adjust XMX to fit mem limit better

@riuvshin yeah, it makes sense but I do not want to play with XMX on the edge of the Christmas shutdown and temporary increasing RAM to 1Gb for sure will not do any harm.

@jfchevrette @pbergene @riuvshin I know this increase is not ideal and we need to do more investigation as part of redhat-developer/che-functional-tests#409 to adjust heap / off heap memory usage. What I'm wondering is could we afford this workaround on dsaas temporary until better RAM / CPU calculations would be done?

ibuziuk:
After 6.16.0 update the situation with RAM seems to be stabilized due to fix of potential memory leak - https://github.com/eclipse/che/issues/12187 


RAM increase is not required anymore. Closing

PR #1151, opened 2018-12-20
DO NOT MERGE

Katka92:
### What does this PR do?
Removing some artifacts from ci.centos.org


Katka92:
All outdated folders and files were removed.

PR #1153, opened 2018-12-20
Removing offline tokens

Katka92:
### What does this PR do?
As far as tests are using CHE_INFRASTRUCTURE=OSIO no offline tokens are needed anymore.
All offline tokens appearances in scripts were removed.


Katka92:
Offline token was used to obtain active token. Now as far as offline tokens are outdated, we have another approach implemented to follow redirects and obtain active token in other way.

ibuziuk:
@Katka92 could you please clarify what is the current approach for obtaining active OSIO token ?

PR #1156, opened 2018-12-20
Refactor scripts

Katka92:
### What does this PR do?
Moving installation of oc. OC_VERSION changed to latest.

### DIFF
--- a/.ci/functional_tests_utils.sh
+++ b/.ci/functional_tests_utils.sh
@@ -35,12 +35,12 @@ function installDependencies() {
 	
 	systemctl start docker
 	pip install yq	
+	curl -s "https://mirror.openshift.com/pub/openshift-v3/clients/${OC_VERSION}/linux/oc.tar.gz" | tar xvz -C /usr/local/bin
 }
 
 function checkAllCreds() {
 	set -x
 	CREDS_NOT_SET="false"
-	curl -s "https://mirror.openshift.com/pub/openshift-v3/clients/${OC_VERSION}/linux/oc.tar.gz" | tar xvz -C /usr/local/bin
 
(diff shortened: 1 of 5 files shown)


.ci/functional_tests_utils.sh
```diff
@@ -35,12 +35,12 @@ function installDependencies() {
 	
 	systemctl start docker
 	pip install yq	
+	curl -s "https://mirror.openshift.com/pub/openshift-v3/clients/${OC_VERSION}/linux/oc.tar.gz" | tar xvz -C /usr/local/bin
```
ibuziuk: this does not look like a really good refactoring:
- why not to have separate function for installing oc ?
- why not to hardcode oc version and avoid calling `export OC_VERSION=3.10.90` in every single script ?

Katka92: I'm not sure how advantageous would be having oc in separate function. I don't see why to have oc in separate function and docker or maven not. These are all dependencies needed for tests to run. 
I though it can be useful to have option to set version in each script. Maybe I could move OC version here.

ibuziuk: > I though it can be useful to have option to set version in each script. 

Could you clarify why / in which cases it might be useful 

> I'm not sure how advantageous would be having oc in separate function

it would improve readability of the script

Katka92: All scripts that uses oc installed this way are run against dev cluster, so there should not be need of setting version independently for each script. I'll change that.
If you think it can be more readable, I'll add this method. :+1: 
Thank you for your suggestions.

Katka92:
[test]

PR #1162, opened 2018-12-26
Enabled job that stops workspaces without pods

sleshchenko:
### What does this PR do?
Sometimes it may happen that workspace Pods goes down when Che Server considers that workspace is RUNNING.
For example, the user stopped it from OpenShift/K8s console, admin can terminate (miner detection) pod or infrastructure node goes down.
This PR enabled a periodical job that will stop such Runtimes without existing pods.

It may work incorrectly on OSIO in some cases. As an example: Che Server doesn't have a cached owner Subject. It happens always after workspace recovering since Subject is cached after workspace starting only. See https://github.com/redhat-developer/rh-che/blob/master/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/WorkspaceSubjectsRegistry.java#L75

Logged errors from `InconsistentRuntimesDetector` should be tracked to make sure that there no other failing cases.

https://github.com/eclipse/che/pull/12257

I tested this job on upstream packaging deployed on minishift where Che Server that creates workspaces objects in the same namespace where Che Server is run. But I haven't tested it on OSIO packaging.


sleshchenko:
Actually, AFAIU it won't fix an issue with 600 cached runtimes while there are about 30 running workspaces.
@garagatyi @ibuziuk Is there any plans to fix an issue with cached subjects?

ibuziuk:
@sleshchenko thanks for the contribution and deep analysis. Am I correct that this PR should NOT be applied until rh-che would be using `6.17.0` / `6.16.1` upstream ?

sleshchenko:
@ibuziuk `CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN` env var will be ignored if PR will be applied before 6.17.0. So, this PR may be applied earlier.

ibuziuk:
sure, but we still need a HK issue for updating cm on dsaas / dsaas-stg. Also, taken into account that it may work incorrectly in rh-che I would probably put it on halt and test  meticulously before merging. WDYT ?

sleshchenko:
@ibuziuk Agree with you. It definitely won't work correctly for restored workspaces. So, what about cached subjects issue? Are there any plans for it?

ibuziuk:
@sleshchenko sure thing we need to address it, but I'm not sure we will be able to do it in the current sprint. Currently che server on prod has been running without restarts for 16 days, so the situation with OOM restarts seems to be stabilized. What I currently want to do is to update prod to the latest `6.16.0`, which contains fix for https://github.com/eclipse/che/issues/12187 and already should improve the situation significantly + create an issue for cached subjects (for the next sprint). Does it makes sense or you think that cached subject issue is more urgent ?

garagatyi:
@ibuziuk I think it should be done upstream. See https://github.com/eclipse/che/issues/12304

sleshchenko:
> Currently che server on prod has been running without restarts for 16 days,

Looks like quite stable. But it is not about OOM but about rolling update. After Che Server update: there is no ability for Che Server to stop workspace because of inactivity, an absence of workspaces pods.

I'm not sure that persisting of user's token that is described in the issue[[1](https://github.com/eclipse/che/issues/12304)] will help us. Like miner blocking use case. As far as I understand: admin marks keycloak user as `unprovisioned` which leads to keycloak user banning and he is not able to retrieve token anymore. I assume that the current token is expired as well. So, in this case, Che Server will have invalid user's token.
So, needed to list all cases when Che Server might need to stop user's workspace and investigate them deeply.

Just note that even if we will persist user's subjects it won't help to get rid of existing `RUNNING` workspaces which are not running anymore. It may be done only with updating Che Server by stopping all workspaces, or stopping such workspaces one by one by admin, or granting Che Server cluster admin token to check pods existence without user's subject at all.

> Does it makes sense or you think that cached subject issue is more urgent ?

So, it makes sense for me. Cached subjects issue is not urgent task but very important.

[1] https://github.com/eclipse/che/issues/12304

ibuziuk:
@sleshchenko @garagatyi  I might be missing some details, but in rh-che we are currently using Che SA token and NOT keycloak / OSIO user token for all the communications with the oso cluster. For me it sounds like this token can be used for recovery without hassle related to persistence (all we need to know is namespace / user associated with the workspace). Does it make sense ?

garagatyi:
@ibuziuk +1

garagatyi:
> It may be done only with updating Che Server by stopping all workspaces, or stopping such workspaces one by one by admin, or granting Che Server cluster admin token to check pods existence without user's subject at all.

@sleshchenko can you elaborate on that. Not sure I following you here

sleshchenko:
Sorry guys, I'm not really aware of the actual state of RhChe.
I thought that RhChe uses a special token for communication with OSO cluster. And last week @garagatyi and I investigated a bit RhChe and discovered that there are in-memory cached subjects https://github.com/redhat-developer/rh-che/blob/master/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftClientFactory.java#L91
If RhChe uses service account token but not user keacloak - why we need persisting subjects at all [1]? 
[1] https://github.com/eclipse/che/issues/12304

garagatyi:
Even though we use SA on prod we still have some issues with access to OS API.
Here are examples:
https://errortracking.prod-preview.openshift.io/openshift_io/che6master/issues/9256/
https://errortracking.prod-preview.openshift.io/openshift_io/che6master/issues/9353/
https://errortracking.prod-preview.openshift.io/openshift_io/che6master/issues/9348/

Not sure whether all these use cases related to an outdated situation with the use of tokens. 
Thoughts?

garagatyi:
@sleshchenko could you remind whether we can apply this PR or not?

sleshchenko:
@garagatyi I think this PR may be applied. I guess it still may not work in some cases when Che Server does not have permissions to access user namespace but it should not cause any new user-visible errors, just logged error in Che Server log.

ibuziuk:
please, do not apply until new cm entries for `CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN` are created on dsaas / dsaas-stg

garagatyi:
@ibuziuk who is working on that?

ibuziuk:
no one AFAIK.

garagatyi:
@sleshchenko could you do this to proceed with your PR?

ibuziuk:
@sleshchenko I will be creating a few issue on HK related to cm updates, so can take care of this one also, if it works for you.

sleshchenko:
@garagatyi Sure, could you please provide more details what should I do? Where ConfigMap for `dsaas` / `dsaas-stg` is located?

sleshchenko:
@ibuziuk It works for me, thanks.

ibuziuk:
cms have been updated. Merging

PR #1165, opened 2018-12-28
Remove dead code

garagatyi:
### What does this PR do?
Remove dead code. Suggested by JDT.LS!


garagatyi:
@davidfestal Please, review

davidfestal:
This code has been initially inspired by [this code](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/WsAgentServerLivenessProbeConfigFactory.java#L58), but it seems that I finally forgot recreating the URL with the real port, as it is done [here](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/HttpProbeFactory.java#L36).
Hopefully it seems it is not necessary (since it appears to be working as it is). However I wonder if the right fix isn't to create the URL with the right port in all cases ?

garagatyi:
@davidfestal I think that server URL should be valid, so this code wouldn't change anything.

garagatyi:
@davidfestal so, do you think that I can merge the PR?

PR #1172, opened 2019-01-08
Fix e2e test

Katka92:
### What does this PR do?
E2e test is sometimes failing because it can not open file in a workspace. That happens because of pop-ups that stole a focus. We solved this problem in bayesian tests by using projectExplorer feature implemented in upstream. Switching into that approach should help here too. 

locally against productin env


ScrewTSW:
@Katka92 Shouldn't we implement this change for our tests as well? Won't our MavenPom and PackageJson tests also face the same issue?

PR #1186, opened 2019-01-14
Clear warnings from tests, simplify test output

ScrewTSW:
### What does this PR do?
* Fixes some warnings present in our test runs
* Shortens log output from preparation of environment (package installs, docker pulls)

manual testing

### DIFF
--- a/functional-tests/devscripts/prepare_env.sh
+++ b/functional-tests/devscripts/prepare_env.sh
@@ -2,9 +2,9 @@
 
 set -e
 echo "Installing docker..."
-yum install -y docker
+yum install -y docker | cat # Suppress multiple-line output for package installation
 systemctl start docker
-docker pull quay.io/openshiftio/rhchestage-rh-che-functional-tests-dep
+docker pull quay.io/openshiftio/rhchestage-rh-che-functional-tests-dep | cat # Suppress multiple-line output for docker pull
 
 export HOST_URL=$HOST_URL
 eval "$(./env-toolkit load -f jenkins-env.json -r  USERNAME PASSWORD EMAIL OFFLINE_TOKEN JOB_NAME BUILD_NUMBER)"
--- a/functional-tests/devscripts/run_tests.sh
+++ b/functional-tests/devscripts/run_tests.sh
@@ -10,7 +10,7 @@ function printHelp {
 	YELLOW="\\033[93;1m"
 	WHITE="\\033[0;1m"
 	GREEN="\\033[32;1m"
-	NC='\033[0m' # No Color
+	NC="\\033[0m" # No Color
 	
 	echo -e "${YELLOW}$(basename "$0") ${WHITE}[-u <username>] [-p <passwd>] [-m <email>] [-r <url>]" 
 	echo -e "\n${NC}Script for running functional tests against production or prod-preview environment."
--- a/functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
+++ b/functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
@@ -22,6 +22,7 @@ import com.redhat.che.selenium.core.model.request.CheStarterCreateWorkspaceReque
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.nio.charset.Charset;
 import java.util.HashMap;
 import javax.ws.rs.HttpMethod;
 import okhttp3.MediaType;
@@ -29,6 +30,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.rest.HttpJsonRequestFactory;
@@ -105,7 +107,9 @@ public class CheStarterWrapper {
     BufferedReader jsonRequestReader;
     try {
       jsonRequestReader =
-          new BufferedReader(new InputStreamReader(getClass().getResourceAsStream(pathToJson)));
+          new BufferedReader(
+              new InputStreamReader(
+                  getClass().getResourceAsStream(pathToJson), Charset.forName("UTF-8")));
     } catch (Exception e) {
       LOG.error("File with json was not found on address: " + pathToJson, e);
       throw e;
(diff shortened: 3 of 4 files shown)


functional-tests/devscripts/run_tests.sh
```diff
@@ -10,7 +10,7 @@ function printHelp {
 	YELLOW="\\033[93;1m"
 	WHITE="\\033[0;1m"
 	GREEN="\\033[32;1m"
-	NC='\033[0m' # No Color
+	NC='\\033[0m' # No Color
```
Katka92: Why there is only one ```\``` and others stay?

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
```diff
@@ -98,6 +98,7 @@ public void closeFiles() {
     closeFilesAndProject();
   }
 
+  @Override
```
Katka92: It doesn't make sense for me to override that. If there is a problem with the method with same name, maybe changing the name would be enough?

functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
```diff
@@ -59,10 +59,12 @@ public void testCorrectWayOfStarting() throws Exception {
   @Test
   public void testWrongWayOfStarting() throws ExecutionException, InterruptedException {
     LOG.info("Try to start " + firstWorksapce.getName() + " without PATCH.");
+    LOG.info("--- ERROR EXPECTED ---");
```
rhopp: What's the purpose of these two log outputs? 
If you want to refactor this method, you are more than welcome, but do it in separate issue... There are definitely things to improve here (order of methods is fixed - if they would run in different order, they will fail. Also I think it would be better to use expectedException, than own handling of exception... And another thing is to use more specific exception than just Exception (this boils further down the code to fix)).
I don't see any value in adding such logs into this method... I think this method deserves some more refactoring (but unfortunately we have more important tasks on our hands)

functional-tests/devscripts/prepare_env.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e
 echo "Installing docker..."
-yum install -y docker
+yum install -y docker | cat # Suppress multiple-line output for package installation
```
amisevsk: I'd clarify that piping to `cat` is to suppress tty output (e.g. progress). There is also the `-q` option to suppress all output.

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
```diff
@@ -208,12 +212,16 @@ private void patchPrepareEnv(String workspaceName, String token) throws IOExcept
 
     try {
       response = client.newCall(request).execute();
-      if (response.isSuccessful()) {
+      boolean responseStatus = response.isSuccessful();
+      int responseCode = response.code();
+      ResponseBody responseBody = response.body();
+      response.close();
+      if (responseStatus) {
```
amisevsk: Suggestion: rename to `responseSuccess`. Status could be confused with status code.

functional-tests/devscripts/run_tests.sh
ScrewTSW: ? the `\` backwards slash was missing. others use doubleslash correctly

ScrewTSW: what's more intriguing to me is why `NC` uses `'` instead of `"`

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
ScrewTSW: I got a warning that the method is overriding a method from `super`, which it is, and that the `@Override` annotation was missing, so I added it.

functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
ScrewTSW: This test case was putting an entire stack trace into stdout, because of method that it uses, so I added the `error expected` test, so that we won't be intrigued by the exception (which it is supposed to throw)

functional-tests/devscripts/prepare_env.sh
ScrewTSW: I don't want to silence it entirely. I just want to avoid it adding 3 new lines into logs every time the progress bar moves. this way it only writes the install log for every package once

functional-tests/src/main/java/com/redhat/che/selenium/core/workspace/CheStarterWrapper.java
ScrewTSW: noted :)

functional-tests/devscripts/run_tests.sh
Katka92: Yes, maybe you can try it and fix that. I think it will not take a lot of time.

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
Katka92: Basically the idea is to add @AfterClass annotation to the method. I'm not sure if it is needed to @Override it as far as the body of the method is exactly the same as in the parent. 
Wouldn't it be better to use something like that?
```
  @AfterClass
  public void closeBrowserAfterTests() {	  
    super.closeBrowser();	    
  }	  
```

functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
rhopp: I still don't see a point in having such log message here. It solves nothing. This method has bigger problems than this and should (IMHO) be refactored. But that should be done as part of another issue/PR.

ScrewTSW: ok

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
ScrewTSW: if it's exactly the same as parent, then maybe the implementation of the `@AfterClass` in the test case is not needed. the `@AfterClass` of parent class should be called, right?

functional-tests/src/test/java/com/redhat/che/functional/tests/WorkspaceTest.java
ibuziuk: agree with @rhopp those terminal approach log entries should be avoided

functional-tests/src/test/java/com/redhat/che/functional/tests/BayesianAbstractTestClass.java
Katka92: But there is no @AfterClass in a parent. And as far as parent should be common for all test cases, I'm not sure if we should set methods like @BeforClass and @AfterClass there. I think only the behaviour should be implemented and exact class inheriting from that should say if it wants to use that method in specific way (e.g. as @AfterClass).

rhopp:
@ScrewTSW Could you also create an issue to refactor `WorkspaceTest`? Thanks!

PR #1194, opened 2019-01-16
Update to 6.17.0

fabric8cd:
Tracking changes for fixing compatibility with upstream 6.17.0-SNAPSHOT. This PR was created automatically by Jenkins from job devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io


ibuziuk:
@Katka92 what is the PR about? It looks like a mixture of doc, scripts and version update

Katka92:
@ibuziuk this is PR created from testing compatibility check. The only was to test it is via PR so there is my code too. Please ignore this PR now as far as compatibility check is not working correctly and it will create PR each time I'll try it, so it doesn't make sense to close it.

PR #1208, opened 2019-01-21
Fix setting of protocol in docker entrypoint.

Katka92:
### What does this PR do?
There was a bug in setting protocol in docker entrypoint - it was set only in rollout test. 
This fix is needed for fixing PR check job.

### DIFF
--- a/dockerfiles/functional-tests/docker-entrypoint.sh
+++ b/dockerfiles/functional-tests/docker-entrypoint.sh
@@ -79,6 +79,10 @@ if [[ -n $RUNNING_WORKSPACE ]]; then
 	MVN_COMMAND="${MVN_COMMAND} -Dche.workspaceName=${RUNNING_WORKSPACE}"
 fi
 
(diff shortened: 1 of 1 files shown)


Katka92:
@rhopp please could you review?

dockerfiles/functional-tests/docker-entrypoint.sh
```diff
@@ -79,6 +79,10 @@ if [[ -n $RUNNING_WORKSPACE ]]; then
 	MVN_COMMAND="${MVN_COMMAND} -Dche.workspaceName=${RUNNING_WORKSPACE}"
 fi
 
+if [[ -n $RHCHE_HOST_PROTOCOL ]]; then
```
rhopp: What about port? https://github.com/redhat-developer/rh-che/blob/master/functional-tests/pom.xml#L29 Here we are setting default value to "443". Is that correct? If not, I guess this PR should take care of this too...

Issue #1210, opened 2019-01-22
Che 7 workspace startup fails on che.prod-preview.openshift.io


ibuziuk:

ibuziuk:
comment form @garagatyi 

> Couldn't connect to endpoint 'wss://che.prod-preview.openshift.io/api/websocket', due to error 'x509: failed to load system roots and no roots provided' seems that there is an error in the init broker dokerfile

PR #1212, opened 2019-01-22
Update version of init broker to 0.7.1

garagatyi:
### What does this PR do?
Update version of init broker to 0.7.1

https://github.com/redhat-developer/rh-che/issues/1210

No, requires SSL powered WS Master


garagatyi:
HK issue https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2529

ibuziuk:
@garagatyi am I correct that we would need to remove this property in the `6.18.0` release? Basically, it is a workaround for not requesting `6.17.2` ?

PR #1218, opened 2019-01-24
Fixing cleanup job.

Katka92:
### What does this PR do?
Fix scripts for clean up job.

### DIFF
--- a/.ci/cico_cleanup_dev_cluster.sh
+++ b/.ci/cico_cleanup_dev_cluster.sh
@@ -12,7 +12,6 @@
 export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
 
 source ./.ci/functional_tests_utils.sh
-installEpelRelease
 installJQ
 installOC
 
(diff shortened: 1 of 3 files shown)


.ci/cico_cleanup_dev_cluster.sh
```diff
@@ -12,6 +12,7 @@
 export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
 
 source ./.ci/functional_tests_utils.sh
+installEpelRelease
```
ibuziuk: could you clarify why do we need this call ?

Katka92: Epel is needed to be installed because it provides JQ which is installed right after.

rhopp: maybe the `installJQ` function can take care of that? But that's just a minor note, not preventing the merge from my POV.

ibuziuk: @Katka92 agree with @rhopp `installEpelRelease` does not really needed if the only reason it is required is for JQ. Please, consider refactoring since installing Epel is just an impl detail in this case

Katka92: I've moved enabling of epel. @ibuziuk could you please review again?

Katka92:
[test]

Katka92:
[test]

Katka92:
[test]

Katka92:
[test]

Issue #1211, opened 2019-01-22
Minishift addon deployment - Login from OKD fails


ScrewTSW:
**Issue problem:** 
I'm using the latest version of minishift-addons form rh-che master branch, with `minishift v1.30.0+186b034` and a brand new virtual machine with 4GB of RAM and 2 CPU cores.

The deployment works as intended (first installing `rhche-prerequisites` addon and then the `rhche` addon)

Upon opening the dashboard I'm redirected to the OKD login page and after successfully entering the credentials I'm redirected back to dashboard with this error:
![screenshot from 2019-01-22 09-06-04](https://user-images.githubusercontent.com/16451875/51541149-601d5080-1e58-11e9-9166-6e9e1ce82a25.png)

There are no errors in the rhche logs, only exception being thrown is in the keycloak logs:
```
08:28:41,450 WARN  [org.keycloak.events] (default task-20) type=LOGIN_ERROR, realmId=che, clientId=null, userId=null, ipAddress=172.17.0.1, error=identity_provider_login_failure
08:31:16,680 ERROR [org.keycloak.broker.oidc.AbstractOAuth2IdentityProvider] (default task-7) Failed to make identity provider oauth callback: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
    at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
    at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1959)
    at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:302)
    at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:296)
...
```
![screenshot from 2019-01-22 15-28-00](https://user-images.githubusercontent.com/16451875/51542044-9b208380-1e5a-11e9-8547-0349ea21f843.png)
![screenshot from 2019-01-22 15-28-21](https://user-images.githubusercontent.com/16451875/51542046-9bb91a00-1e5a-11e9-955a-610c9f4b0eab.png)

**Red Hat Che version:** 

version: minishift quay.io/openshiftio/che-rh-che-server:df0c632
(cannot get exact version as I cannot get to dashboard)

 - [x] I can reproduce it on latest official image

**Reproduction Steps:**
* create new minishift VM with 4GB of RAM and 2 CPUs
[...]

PR #1246, opened 2019-02-12
Fix the Minishift RhChe addon

davidfestal:
### What does this PR do?
 This PR fixes the Minishift addon that was broken after a change in upstream Che Keycloak docker image.

https://github.com/redhat-developer/rh-che/issues/1211

Yes, on Minishift v1.31.0+cfc599c (with oc 3.11)


davidfestal:
@tdancs Can you confirm this works for you now ?

davidfestal:
Just a word about commit https://github.com/redhat-developer/rh-che/pull/1246/commits/8cbafcb1ce389f3bdb9c35ec1bb8eee6f9ec99a7:

In order to have it working property and in a stable way, I had to add the path of the kube config to the `ssh` command, and it might break compatibility with previous versions of Minishift / Openshift.

I only tested it with Minishift v1.31.0+cfc599c  (and oc 3.11)

@ibuziuk : Do you think it is acceptable to expect the addon to be used with the last Minishift release, or should we invest more work in making it compatible with a number of previous releases ?

ibuziuk:
@davidfestal I do believe it is acceptable and making latest version of oc / minishift a requirement is fully ok

PR #1255, opened 2019-02-20
Throw checked exceptions in TenantData cache loader

amisevsk:
### What does this PR do?
Change `TenantDataProvider` to throw checked exceptions when unable to load tenant data.

The API doc for [`LoadingCache#get()`](https://static.javadoc.io/com.google.guava/guava/16.0/com/google/common/cache/LoadingCache.html#get(K)) specifies that an `ExecutionException` is thrown when a *checked* exception is thrown in the load method. We currently throw `RuntimeException`s, resulting in those errors being propagated to the top of the stack unhandled.

Check prod Kibana query `kubernetes.container_name:"rhche" AND kubernetes.namespace_name:"dsaas-production" AND *TenantDataProvider*`

Sample stacktrace:
```
com.google.common.util.concurrent.UncheckedExecutionException: java.lang.RuntimeException: Exception during the user tenant data retrieval
    [<snip>]
	at com.redhat.che.multitenant.TenantDataProvider.getUserCheTenantData(TenantDataProvider.java:92)
	at com.redhat.che.multitenant.Fabric8WorkspaceEnvironmentProvider.getUserCheTenantData(Fabric8WorkspaceEnvironmentProvider.java:168)
	at com.redhat.che.multitenant.Fabric8WorkspaceEnvironmentProvider.getWorkspacesOpenshiftConfig(Fabric8WorkspaceEnvironmentProvider.java:109)
	at com.redhat.che.multitenant.Fabric8OpenShiftClientFactory.buildConfig(Fabric8OpenShiftClientFactory.java:126)
	at org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesClientFactory.create(KubernetesClientFactory.java:101)
    [<snip>]
Caused by: java.lang.RuntimeException: Exception during the user tenant data retrieval
	at com.redhat.che.multitenant.TenantDataProvider.loadUserCheTenantData(TenantDataProvider.java:125)
    [snip]
	... 87 common frames omitted
Caused by: java.io.IOException: Failed access: https://api.openshift.io/api/user/services?token, method: GET, response code: 401, message: {"errors":[{"code":"jwt_security_error","detail":"JWT validation failed: crypto/rsa: verification error","id":"VLh4JIPD","status":"401","title":"Unauthorized"}]}

	at org.eclipse.che.api.core.rest.DefaultHttpJsonRequest.doRequest(DefaultHttpJsonRequest.java:285)
	at org.eclipse.che.api.core.rest.DefaultHttpJsonRequest.request(DefaultHttpJsonRequest.java:164)
	at com.redhat.che.multitenant.TenantDataProvider.getResponseBody(TenantDataProvider.java:178)
	at com.redhat.che.multitenant.TenantDataProvider.loadUserCheTenantData(TenantDataProvider.java:117)
	... 92 common frames omitted
```

Tested on dev-cluster -- creation / start / stop of workspace works as expected.


plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
```diff
@@ -179,6 +118,72 @@ private String getResponseBody(final String endpoint, final String keycloakToken
     return response.asString();
   }
 
+  private class TenantDataLoader extends CacheLoader<CacheKey, UserCheTenantData> {
```
ibuziuk: if we started the refactoring maybe it makes sense to create a separate non-private class ?

amisevsk: I only moved it to a private class because the previous format (`CacheLoader#from(...)`) doesn't allow exceptions to be thrown. I still think that these classes (`TenantDataLoader` and `CacheKey`) only many sense within `TenantDataProvider` for now.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
```diff
@@ -179,6 +118,72 @@ private String getResponseBody(final String endpoint, final String keycloakToken
...
+
+    @Override
+    public UserCheTenantData load(CacheKey cacheKey) throws InfrastructureException {
+      if (standalone) {
```
garagatyi: Do we need this standalone part at all? AFAIK we wanted to remove single user functionality completely. Maybe it worth removing this part now?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
```diff
@@ -179,6 +118,72 @@ private String getResponseBody(final String endpoint, final String keycloakToken
...
+            namespaceToUse = "myproject";
+        }
+        return new UserCheTenantData(
+            namespaceToUse, "https://kubernetes.default.svc", "dummy.prefix.unused", false);
+      }
+
+      String responseBody;
+      try {
+        responseBody = getResponseBody(fabric8UserServiceEndpoint, cacheKey.keycloakToken);
+      } catch (ServerException
```
garagatyi: Can we just use Exception instead of listing all possible exceptions here?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
ibuziuk: AFAIK, standalone = minishift rh-che addon

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
```diff
@@ -179,6 +118,72 @@ private String getResponseBody(final String endpoint, final String keycloakToken
...
+          | UnauthorizedException
+          | ForbiddenException
+          | NotFoundException
+          | ConflictException
+          | BadRequestException
+          | IOException e) {
+        throw new InfrastructureException("Exception during the user tenant data retrieval", e);
+      }
+      try {
+        final JsonArray namespaces =
```
garagatyi: It is legacy code but still this code is very complex for such a simple thing as getting fields of an object from a response. You could drastically improve this code by refactoring parsing. You could create an object with only fields that we need here and parse response as a single operation instead of traversing it manually. You could use Google GSON used here or switch to Jackson - both libraries are heavily used in upstream Che.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
garagatyi: oh, I forgot about this use-case,thanks

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
amisevsk: We can -- I just copy-pasted :)

Fixed.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
```diff
@@ -179,6 +118,72 @@ private String getResponseBody(final String endpoint, final String keycloakToken
     return response.asString();
   }
 
+  private class TenantDataLoader extends CacheLoader<CacheKey, UserCheTenantData> {
+
+    @Override
+    public UserCheTenantData load(CacheKey cacheKey) throws InfrastructureException {
```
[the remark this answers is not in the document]
amisevsk: Good point -- added tests

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/TenantDataProvider.java
amisevsk: Mostly I was avoiding this because I hate dealing with json in Java ;)

amisevsk:
@garagatyi I've added two commits, one refactoring TenantDataProvider into multiple classes, and another adding test cases for the above. See commit messages for more detailed description of changes.

amisevsk:
>  I would love seeing this code refactored

I usually avoid refactoring unless it is necessary, since I think it makes review more complicated, but I agree it's probably helpful in this case.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheKey.java
```diff
@@ -0,0 +1,40 @@
+package com.redhat.che.multitenant.tenantdata;
```
ibuziuk: no license header

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
```diff
@@ -0,0 +1,100 @@
+package com.redhat.che.multitenant.tenantdata;
```
ibuziuk: no license header

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheKey.java
amisevsk: Fixed.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
amisevsk: Fixed.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataProvider.java
```diff
@@ -0,0 +1,78 @@
...
+   * @param namespaceType type of the namespace which tenant data are requested
+   * @throws InfrastructureException when any error occurs or there is no namespace of specified
+   *     type in user's tenant
+   */
+  public UserCheTenantData getUserCheTenantData(Subject subject, String namespaceType)
+      throws InfrastructureException {
+    checkSubject(subject);
+
+    String keycloakToken = subject.getToken();
+    if (keycloakToken == null) {
```
garagatyi: Could we move this into checkSubject method?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
```diff
@@ -0,0 +1,111 @@
...
+              new UserCheTenantData(
+                  ns.getName(),
+                  ns.getClusterUrl(),
+                  ns.getClusterAppDomain(),
+                  ns.isClusterCapacityExhausted());
+          UserCheTenantDataValidator.validate(cheTenantData);
+          return cheTenantData;
+        }
+      }
+    } catch (ValidationException | JsonSyntaxException e) {
```
garagatyi: Is NPE possible here?

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataProvider.java
garagatyi: Could you check whether this case is covered by unit tests?

garagatyi:
> I usually avoid refactoring unless it is necessary, since I think it makes review more complicated, but I agree it's probably helpful in this case.

@amisevsk we wouldn't have an opportunity to refactor on purpose. It's like a dream for an engineer that is unachievable unless code got really shitty. I'm trying to refactor the code on every occasion and recommend doing the same to others. Like TenantDataProvider class - even though I'm listed as the author, I just moved this code out of the another class. Now you are simplifying it hugely and code looks way better after several improvements by different engineers. 
Apart from that, an engineer rarely has an idea of the clean, bug-free and testable code right away, so it is better to improve code iteratively and with small changes. 

And thank you for the refactoring and tests!

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
```diff
@@ -0,0 +1,111 @@
...
+          "Invalid response from Fabric8 user services:" + responseBody, e);
+    }
+    throw new InfrastructureException(
+        format(
+            "No namespace with type '%s' was found in the user tenant",
+            cacheKey.getNamespaceType()));
+  }
+
+  private UserCheTenantData loadStandalone(TenantDataCacheKey cacheKey) {
+    String namespaceType = cacheKey.getNamespaceType();
```
garagatyi: We could move this to the retriever since caching is not needed for this case.

plugins/fabric8-multi-tenant-manager/src/test/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoaderTest.java
```diff
@@ -0,0 +1,158 @@
...
+    UserCheTenantData data = cacheLoader.load(cacheKey);
+
+    // Then
+    assertEquals(data.getRouteBaseSuffix(), ROUTE_PREFIX);
+    assertEquals(data.getClusterUrl(), CLUSTER_URL);
+    assertEquals(data.getNamespace(), NAMESPACE);
+    assertEquals(data.isClusterCapacityExhausted(), false);
+  }
+
+  @Test(expectedExceptions = InfrastructureException.class)
```
garagatyi: From my experience tests that expect exception but do not specify exception message tend to degrade after new changes because some another error is caught instead of the error that was supposed to be caught by the test. Consider adding exception message.

plugins/fabric8-multi-tenant-manager/src/test/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoaderTest.java
```diff
@@ -0,0 +1,158 @@
...
+    TenantDataCacheKey cacheKey = new TenantDataCacheKey("token", "che");
+
+    // When
+    cacheLoader.load(cacheKey);
+  }
+
+  @Test
+  public void standaloneShouldReturnDefaultDataWithCheNamespaceType() throws Exception {
+    // Given
+    TenantDataCacheKey cacheKey = new TenantDataCacheKey("token", "che");
```
garagatyi: Consider adding test that checks case when json is valid but an attribute is null instead of the object. This potentially can emit NPE. For example 
 ```json
{
    "data": {
        "attributes": null
    }
}
```

amisevsk: I've added a test cast to cover the NPE described above, that *should* cover this. Apart from that if any of the fields in namespace is null that should cause a `ValidationException`.

plugins/fabric8-multi-tenant-manager/src/test/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoaderTest.java
amisevsk: Added regexes.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
amisevsk: You're right, that's cleaner.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataProvider.java
amisevsk: Added test case

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoader.java
amisevsk: Good point -- I was looking at the validator but if the response cannot be parsed at all then `ns` could be null.

Added test case to cover this scenario.

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/tenantdata/TenantDataProvider.java
amisevsk: Done.

plugins/fabric8-multi-tenant-manager/src/test/java/com/redhat/che/multitenant/tenantdata/TenantDataCacheLoaderTest.java
amisevsk: Added a javadoc to the `getNamespaces()` method to mention the potential NPE.

amisevsk:
[test]

ibuziuk:
@amisevsk there are some proposals from Codancy (e.g. Avoid unused private fields such as 'standaloneCacheLoader'.)

ibuziuk:
[test]

amisevsk:
Fixed codacy comments (came in late at night) and squashed fixup commits into previous.

amisevsk:
[test]

ibuziuk:
[test]

amisevsk:
[test]

amisevsk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

amisevsk:
postgres issue again :/

ibuziuk:
[test]

ibuziuk:
[test]

amisevsk:
[test]

amisevsk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

PR #1259, opened 2019-02-25
Disabling unstable tests.

Katka92:
### What does this PR do?
After discussion with @rhopp and due to the fact that Che 6 will be deprecated soon, we decided to disable unstable tests for now. We would like to have at least something small but stable and focus on Che 7 tests.
We decided to keep test for starting workspaces and test for build and run vert.x hello world project.
As the next step I'll create two more jobs on jenkins to run these test against clusters 1b and 1a. 

https://github.com/redhat-developer/che-functional-tests/issues/430


functional-tests/src/test/resources/suites/prcheck.xml
```diff
@@ -19,11 +19,8 @@
 
     <test name="all">
         <classes>
-        <!-- temporary changes to check if workaround works -->
-<!--             <class name="com.redhat.che.functional.tests.WorkspaceTest"/> -->
+	        <class name="com.redhat.che.functional.tests.WorkspaceTest"/>
```
amisevsk: Looks like there's a stray `<tab>` on this line messing with the indentation

Issue #938, opened 2018-10-01
Investigate strange intermittent test fail building wsagent


ScrewTSW:
**Issue problem:** 
Master build tests fail with a strange error in one of unit tests:
```
[ERROR] Tests run: 31, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 22.925 s <<< FAILURE! - in TestSuite
[ERROR] sendWorkspaceStoppedOnDestroy(com.redhat.che.plugin.analytics.wsagent.AnalyticsManagerTest)  Time elapsed: 0.033 s  <<< FAILURE!
java.lang.AssertionError: null: lists don't have the same size expected [2] but found [3]
    at com.redhat.che.plugin.analytics.wsagent.AnalyticsManagerTest.sendWorkspaceStoppedOnDestroy(AnalyticsManagerTest.java:513)
```
https://ci.centos.org/view/Devtools/job/devtools-rh-che-build-che-credentials-master/119/console

```
[ERROR] Tests run: 33, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 17.653 s <<< FAILURE! - in TestSuite
[ERROR] sendWorkspaceUsedOnActivityAfterInactivity(com.redhat.che.plugin.analytics.wsagent.AnalyticsManagerTest)  Time elapsed: 0.028 s  <<< FAILURE!
java.lang.AssertionError: Lists differ at element [0]: Keep Workspace Inactive in Che != Use Workspace in Che expected [Keep Workspace Inactive in Che] but found [Use Workspace in Che]
	at com.redhat.che.plugin.analytics.wsagent.AnalyticsManagerTest.sendWorkspaceUsedOnActivityAfterInactivity(AnalyticsManagerTest.java:467)
```
https://ci.centos.org/view/Devtools/job/devtools-rh-che-build-che-credentials-master/279/console

version: 6.11.1

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**

Intermittent issue with build.

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [ ] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

ibuziuk:
More logs from https://ci.centos.org/view/Devtools/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/779/console

```
[...]

PR #1268, opened 2019-02-27
Fix flaky test in AnalyticsManager

amisevsk:
## Related issue 

https://github.com/redhat-developer/rh-che/issues/938

## What does this PR do?
Fixes a flaky test in `AnalyticsManager`. 

Sometimes the `sendWorkspaceStoppedOnDestroy` test fails due to
```
java.lang.AssertionError: null: lists don't have the same size expected [2] but found [3]
	at com.redhat.che.plugin.analytics.wsagent.AnalyticsManagerTest.sendWorkspaceStoppedOnDestroy(AnalyticsManagerTest.java:517)
``` 

Instead we should check for the stop workspace event directly and verify that it is as expected.

I'm not sure why the test can fail (it only occasionally happens -- only reproduced the first time I ran the tests in Eclipse) but I'm guessing it's due to an inactivity event getting in the middle. FWIW I can cause the test to fail by stopping execution in debug before `destroy()` is called.

Local build has worked for me, but the error was not consistent.

### DIFF
--- a/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
+++ b/plugins/che-plugin-analytics/che-plugin-analytics-wsagent/src/main/java/com/redhat/che/plugin/analytics/wsagent/AnalyticsManager.java
@@ -359,7 +359,6 @@ public class AnalyticsManager {
     } catch (ExecutionException e) {
       LOG.warn("", e);
     }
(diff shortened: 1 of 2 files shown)


amisevsk:
Looks like PR check failed on deploying postgres

amisevsk:
[test]

Katka92:
I don't think that tests failed because of changes introduced in this PR. IMHO this PR can be merged.

PR #1273, opened 2019-03-01
Change workaround for importing project

Katka92:
### What does this PR do?
It seems current workaround does not work 100% correctly. Rewriting the workaround to hard refresh all the page.


Katka92:
[test]

PR #1274, opened 2019-03-04
Change docker image for periodic job

Katka92:
Use fresh docker image for periodic tests which is build after merge to master. The last image used was image for PR check job which was build in a job that is failing and therefore periodic tests were 
obsolete.


Katka92:
@rhopp @tdancs could some of you check if this is the correct image? Thank you.

Katka92:
The image is correct. The image is build by ```devtools-rh-che-prcheck-build-dep``` so the job needs to be fixed. Closing.

PR #1277, opened 2019-03-05
Fix job building images

Katka92:
### What does this PR do?
Fix job ```devtools-rh-che-prcheck-build-dep``` as this job is building images for PR check and for periodic tests. The image used in periodic tests is one month old so periodic tests has no value now. Fix of building PR check image is not priority.


Katka92:
Changing order helped - the image for functional tests was deployed. As far as PR check image is not priority now, skipping the fix of this build.

PR #1278, opened 2019-03-05
Fix last failure of tests

Katka92:
### What does this PR do?
Fix last failure of tests [1]. Add method for checking workspace again after refresh of the page. This should wait until project loads correctly.

[1] https://ci.centos.org/job/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/1217/console

localy against prod-preview


functional-tests/src/test/java/com/redhat/che/functional/tests/BuildAndRunProjectTest.java
```diff
@@ -36,6 +36,7 @@ public void checkWorkspace() throws Exception {
     // add hard refresh of the page to fix issue with importing project:
     // https://github.com/openshiftio/openshift.io/issues/4695
     seleniumWebDriver.get(seleniumWebDriver.getCurrentUrl());
+    checkWorkspace(workspace);
```
ibuziuk: IMO, naming is not clear what exactly `checkWorkspace` does ?

rhopp:
I don't see how this could help. The `checkWorkspace` method is called before reloading the page and it succeeds, then the page is reloaded. At this point even by calling `checkWorkspace` we cannot know whether the workspace was started properly (a.k.a. project is imported).

I guess we should implement active waiting loop after the initial `checkWorkspace` method call, which would check whether project is imported. If not, reload page and check again (until it hits a timeout - let's say 30sec-1minute). @Katka92

Katka92:
@rhopp I think this can help because checkWorkspace method waits (besides other checks) that all popups are gone. The current failures was due to these scenario:

1) workspace started but project not imported
2) checkWorkspace passes, workspace is ready to be used
3) refresh (project is probably not here)
4) import for project start
5) immediatelly try to open file - failed

Adding next checkWorkspace after refresh addes step between 4 and 5 - waits until popups are gone and therefore it waits until project is imported. So that means when the test tries to access some file, it should be imported and ready for use.

Katka92:
@rhopp maybe I can add some check to skip this refresh if project was imported correctly.

PR #1283, opened 2019-03-07
Dislabe check for che-starter as it should be deprecated.

Katka92:
### What does this PR do?
Remove test for che-starter. It causes failures in periodic tests and it should be deprecated so it is not worth to fix that.

### DIFF
--- a/functional-tests/src/test/resources/suites/prcheck.xml
+++ b/functional-tests/src/test/resources/suites/prcheck.xml
@@ -19,7 +19,6 @@
 
     <test name="all">
         <classes>
-            <class name="com.redhat.che.functional.tests.WorkspaceTest"/>
             <class name="com.redhat.che.functional.tests.BuildAndRunProjectTest"/>
         </classes>
     </test>
--- a/functional-tests/src/test/resources/suites/simpleTestSuite.xml
+++ b/functional-tests/src/test/resources/suites/simpleTestSuite.xml
@@ -20,7 +20,6 @@
     <test name="all">
         <classes>
             <class name="org.eclipse.che.selenium.miscellaneous.DialogAboutTest"/>
-            <class name="com.redhat.che.functional.tests.WorkspaceTest"/>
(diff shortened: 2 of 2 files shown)


ibuziuk:
@Katka92 am I correct that there still will be tests for workspace creation via User Dashboard ?

Katka92:
@ibuziuk Now we have only workspace created directly via API. Test you are speaking about is included in upstream tests and we can try to add it to the suite. We used to have this test but when tests became unstable we decreased the amount of tests. As now we have only one left, we can add other tests. I'll create issue about that. Thank you for your comment.

ibuziuk:
> Now we have only workspace created directly via API.

@Katka92  Do you mean che server API? I still do not understand, once this PR is merged will we still have a workspace startup test during PR check? Could you please clarify which tests will be running on PR check once the PR is merged ?

Katka92:
@ibuziuk When this PR is merged, there would be only one test running on PR check [1]. This test tests workspace startup as one part of overal flow. It creates and starts workspace. Once it is successfully done, it tries to build and run project. Then it stops and removes workspace. 
Your question was

> @Katka92 am I correct that there still will be tests for workspace creation via User Dashboard ?

We didn't have test for workspace creation via User Dashboard. I think that it uses ws-master, so yes, it calls che-server API.

[1] https://github.com/redhat-developer/rh-che/pull/1283/files#diff-ecdc7a16b8972c0f256543359f362e46R22

ibuziuk:
@Katka92 thanks for clarification. Once suggestion - test naming should be really clear and stick to the point (e.g. `WorkspaceTest` vs. `WorkspaceStartupUsingWsMasterAPI`), so that it would be 100% clear what test does even for newbies like me ;-)

Katka92:
@ibuziuk You are absolutely right, I wanted to keep it short and simple, but it would be better to keep it more descriptive.

Katka92:
[test]

PR #1288, opened 2019-03-12
Updating rh-che version to che 7 beta 1

ibuziuk:
### What does this PR do?
Maybe we should consider versioning of the rh-che due to https://github.com/eclipse/che/issues/12872

looks like there were some versioning changes in the upstream and parent version is not exposed to the dashboard anymore 
image

Not tested


ibuziuk:
Closing. proper footer versioning is going to be implemented in https://github.com/eclipse/che/issues/12872 / https://github.com/eclipse/che/issues/12871

PR #1293, opened 2019-03-14
Move job info to main README

Katka92:
### What does this PR do?
Move information about statuses of jobs to main README.


Katka92:
@amisevsk Yeah, that's why I tough that it would be better to have these statuses in functional-tests README and to have only a link to that in main README - as it really take a lot of space. But having that as a "chapter" should be some kind of compromise. I'll try that :+1:

Katka92:
@rhopp @ibuziuk could you please review?

PR #1303, opened 2019-03-19
Create locust load tests for OOM testing

ScrewTSW:
### What does this PR do?
This PR adds a completely new performance testing framework and test suite.
It is focused on the goal of stress and load testing rh-che master deployment.

https://github.com/redhat-developer/che-functional-tests/issues/383

manual testing against automated and manual deployments on local minishift, local OCP, devtools OKD


Katka92:
@ScrewTSW could you please take a look on failures on Codacy/PR Quality Review and CodeFactor plugins?

load-tests/README.md
```diff
@@ -0,0 +1,28 @@
...
+    ```
+    yum install python-pip --assumeyes \
+    pip install locust \
+    pip install websocket-client
+    ```
+
+### Running the tests
+
+```
+cd rh-che/load-tests && \
```
Katka92: Is it needed to have ``&& \`` here?

load-tests/README.md
```diff
@@ -0,0 +1,28 @@
...
+LOCUST_RHCHE_ACTIVE_TOKEN="<active_token>" \
```
Katka92: Does these variable needs to be "at one line" with the locust start? In other words, is that ```\``` needed at the end of a line?

load-tests/README.md
```diff
@@ -0,0 +1,28 @@
+# Locust based load tests for RH-Che/CReW
+
+### Prepare environment
+
+It is recommended to do python-pip installations in a separate python virtual-env
```
Katka92: Please go through this file - all senteces, which starts with large number, should end with a dot. ;)

load-tests/real-life.py
```diff
@@ -0,0 +1,102 @@
...
+        for line in json_file:
+            # Parsing line and variables
+            json_payload_raw = line.rstrip()
+            json_rpc_envelope : dict
+            json_rpc_envelope = json.loads(json_payload_raw)
+            # If it's not a request from the workspace side
+            if 'params' in json_rpc_envelope: 
+                json_payload : dict
+                json_payload = json_rpc_envelope.get('params')
+                # print(json_rpc_envelope)
```
Katka92: Should that comment be removed?

load-tests/real-life.py
```diff
@@ -0,0 +1,102 @@
...
+                # If the method is installer/log
+                if json_rpc_envelope.get('method') == "installer/log":
+                    # Process and send payload with a delay
+                    if LAST_TIMESTAMP != None:
+                        # print("Waiting for next message to send")
```
Katka92: Should that be uncommented/removed?

load-tests/real-life.py
```diff
@@ -0,0 +1,102 @@
...
+        self.client.get(self.base_uri)
+
+    def send_after_init_websocket_messages(self):
+        print("Sending AFTER INIT WEBSOCKET messages [" + self.uuid + "].")
+        self.locust.client.connect(self.uuid, self.websocket_uri, "websocket_after_init")
+        with open("websocket_after_init.json") as raw_json:
+            self.parse_json_and_send_websocket(raw_json, "websocket_after_init")
+        self.locust.client.disconnect()
+    
+    def send_after_init_wsmaster_messages(self):
```
Katka92: Does it make sense to have a method with just one line with print?

load-tests/real-life.py
```diff
@@ -0,0 +1,102 @@
...
+        print("Sending AFTER INIT WSMASTER messages [" + self.uuid + "].")
+        pass
+
+    @task
+    def connect_and_send_websocket_messages(self):
+        self.send_init_websocket_messages()
+        # self.load_dashboard()
```
Katka92: What this comment mean? Should it be removed?

load-tests/README.md
ScrewTSW: Theoretically the `&&` could be at the start of the next line, but I made it so that you can just select the command block, paste it into the terminal and it should execute without an error.

load-tests/README.md
ScrewTSW: That's for user readability.
The end result is `cd rh-che/load-tests && LOCUST_RHCHE_ACTIVE_TOKEN=<> LOCUST_RHCHE_PROTOCOL=<> LOCUST_RHCHE_BASE_URI=<> locust -f simple.py` which is really long and confusing.
This way the user can easily see the value settings one per line.

load-tests/real-life.py
ScrewTSW: This was originally meant to be sending the `wsmaster` messages, but there is nothing worth sending to that websocket endpoint.

load-tests/real-life.py
ScrewTSW: I originally also wanted to have a normal HTTP call in the test, like calling GET on dashboard, but it would just overcomplicate the code. Deleted.

load-tests/README.md
Katka92: Ah, I see. I though it is less readable, but if it is more useful, than leave it as it is.

Katka92: But is a difference between what you wrote this approach? 
```
cd rh-che/load-tests
LOCUST_RHCHE_ACTIVE_TOKEN=<> 
LOCUST_RHCHE_PROTOCOL=<> 
LOCUST_RHCHE_BASE_URI=<> 
locust -f simple.py
```

ScrewTSW: Yes. I had issues with `locust -f ...` not seeing all the locally defined variables that way without doing an `export` on them.

Katka92: Ah, ok. So leave it as it is.

PR #1313, opened 2019-03-20
Add Che 7 test

Katka92:
### What does this PR do?
Copy-paste test from upstream and make it compatible. As far as creation of Che 7 workspace is not stable and we don't have flaky test suite yet, this test will not be added to any suite now.

Locally against production. If workspace starts correctly, the test was green.


functional-tests/pom.xml
```diff
@@ -49,6 +49,11 @@
                 <artifactId>che-core-api-dto</artifactId>
                 <version>${che.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.eclipse.che.core</groupId>
+                <artifactId>che-core-commons-lang</artifactId>
+                <version>7.0.0-beta-1.0</version>
```
ibuziuk: ${che.version} should be used

Katka92: Thank you for your notice. Fixed.

Issue #1261, opened 2019-02-26
Update RhChe to 7 beta 2 and adapt configuration 


garagatyi:
**Description:**
PR https://github.com/eclipse/che/pull/12673 should be merged in 7 beta 2. It changes service executor configuration which we use to avoid OOM of Che master and successful start of workspaces. Our configuration needs to be adapted.
What to take into account:
- there are 2 endpoints which are processed by 2 different executor services - major and minor
- Bootstrapper logs **should** (not sure it is checked) go to the minor endpoint. These messages could be safely discarded
- Bootstrapper statuses **should** go to major endpoint. These messages **should not** be discarded.
- Messages queue configuration is added
- 2 executor services consume more RAM than 1, so we need to ensure services are properly configured to avoid OOM

ibuziuk:
First step for this issue https://github.com/redhat-developer/rh-che/pull/1294/

ibuziuk:
Looks like there is a problem with che 7 beta 2 related to exec execution - https://github.com/redhat-developer/rh-che/pull/1294

ibuziuk:
PR for adding `CHE_WEBSOCKET_ENDPOINT__MINOR` - https://github.com/redhat-developer/rh-che/pull/1326
Once merged that should hopefully make https://github.com/redhat-developer/rh-che/pull/1294 to pass
Second step would be updating cms on dsaas / dsaas-stg

ibuziuk:
PR for correctly processing `CHE_WEBSOCKET_ENDPOINT__MINOR` on PR check side https://github.com/redhat-developer/rh-che/pull/1328

ibuziuk:
che.openshift.io has been updated to 7 beta 2. Pending actions:
- [ ] @ScrewTSW is going to provide clear documentation related to JSONRPC queue  settings and stress tests related to OOM / DOS / DDOS
- [ ] @ScrewTSW will update https://github.com/redhat-developer/rh-che/pull/1323 and provide optimal settings for che.openshift.io

ibuziuk:
https://gitlab.cee.redhat.com/service/app-interface/merge_requests/274 is merged so closing this issue
partial docs have been provided in the `Change rhche JSONRPC queue settings` [1] description.
@ScrewTSW @rhopp could you please clarify what is the plan with the docs and https://github.com/redhat-developer/rh-che/issues/1337 in the next sprint?


[1] https://github.com/redhat-developer/rh-che/pull/1323

PR #1323, opened 2019-03-21
Change rhche JSONRPC queue settings

ScrewTSW:
### What does this PR do?
This pull request fixes issues with che-master crashing because of OOM exception.
It adjusts the values so that they correctly fit the amount of memory resources available to the pod.

Websocket major values are set based on the ratio I was able to figure out with the load testing.
Minor websocket is kept on low settings, but is still allocated more resources than in the upstream settings.

~~These settings still influence the original setting (pre-che7-beta-2) where both websockets are on the same endpoint.~~
EDIT: This has become irrelevant, since beta-3 is already in testing. moving on to beta-2

ad.: squashed merge with fix of load-tests README typo.

https://github.com/redhat-developer/che-functional-tests/issues/409
ad: https://github.com/redhat-developer/rh-che/issues/1261

Manual deployments and adjustments of the config values. Strestesting it with our locust suite. Observing behavior under long-term sustained load.

#### Values
* CHE_CORE_JSONRPC_PROCESSOR__MAX__POOL__SIZE : This value determines how many message processor threads we can have running at the same time
* CHE_CORE_JSONRPC_PROCESSOR__CORE__POOL__SIZE : This value determines how many threads will be spawned and accessible right after che-master pod startup
* CHE_CORE_JSONRPC_PROCESSOR__QUEUE__CAPACITY : This value determines how many messages will be stored in che-master pod memory for further consumption, it appears to be a FIFO queue

#### Core issue
With the way that the threadpool behaves, no more processing threads will be spawned until the message queue is full.
The default `CHE_CORE_JSONRPC_PROCESSOR__MAX__POOL__SIZE` is set to `50` which has virtually no effect on the speed of message consumption.
The default value for `CHE_CORE_JSONRPC_PROCESSOR__CORE__POOL__SIZE` is used which is `3`
With the amount of runners so low, even an insignificant but constant load will cause the server's memory to fill up over time and crash.
The message queue will never be full before the server runs out of memory, because the default value for `CHE_CORE_JSONRPC_PROCESSOR__QUEUE__CAPACITY` is `10_000_000`

#### Past attempts to fix
* The `...MAX__POOL__SIZE` variable was increased to `2000` which had no effect because of the issues described above
* The max available memory for the pod was increased to `4Gi` as a temporary fix, which seemed to be enough to prevent the che-master from crashing daily during higher popularity of our service

#### Graphs, yay!
Default upstream settings under load.
Screenshot_20190328_144043
Settings:
* container memory max: 1Gi
* max_pool_size: 50
* core_pool_size: 3
* queue_size: 10_000_000

The server was only able to store `140_000` messages into the queue, only `3` processing threads were running at a rate of `~2000` messages being processed per minute. The server crashed once the memory consumption reached `850Mi`

#### Experimentation with upstream
The upsream server was still vulnerable to the OOM with it's default settings.
The server crashed at around `1 mil` messages stored and the memory consumption only reached `2.5Gi` of all the `4Gi` available before the pod was restarted.

#### Conclusion
Each `100k` messages approximate to `350Mi` of memory usage
With the message queue size of `500k`, the memory consumption reaches `1.5Gi` which is still within the specs
Keeping the memory limit above `3Gi` should be sufficient, but I recommend to keep it at `4Gi` just to be safe
###### MAJOR
* `...CORE_POOL_SIZE` was increased to `30`, to be able to cope with larger influx of messages, processing `2.5k` messages per minute under heavy load to easier cope with influx of messages and keep the response time (message arrives -> message is processed) below 10 seconds.
* `...MAX_POOL_SIZE` was increased to `100` to efficiently clean up the queue if the messages fill it up - experimentally proven that it's enough to get rid of the messages in queue under the load I was able to create during my testing
* `...QUEUE_SIZE` was decreased to `500k` to decrease the memory consumption
###### MINOR
* `...CORE_POOL_SIZE` was increased to `5`, messages will still be dropped but they will be more efficiently processed as most of the load relies on the minor websocket now
* `...MAX_POOL_SIZE` was increased to `100` to attempt to process as many bootstrap logs as possible
* `...QUEUE_SIZE` was increased to `1000`, because at default `100` the messages were processed faster than they were able to arrive into the queue. The queue appeared to be empty and messages were being dropped


openshift/rh-che.app.yaml
```diff
@@ -120,6 +120,16 @@ objects:
             value: "openshift"
           - name: SENTRY_RELEASE
             value: ${IMAGE_TAG}
+          - name: CHE_CORE_JSONRPC_PROCESSOR__MAX__POOL__SIZE
```
garagatyi: We do not add env vars that are in config map here anymore. We include all the entries from the config map as env vars, so, please, just rename config maps entries to be appropriate env vars for Che

openshift/rh-che.config.yaml
```diff
@@ -71,3 +70,5 @@ data:
   CHE_WORKSPACE_STORAGE: "/home/user/che/workspaces"
   CHE_WORKSPACE_PLUGIN__BROKER_VSCODE_IMAGE: "eclipse/che-vscode-extension-broker:v0.13.0"
   CHE_FABRIC8_CDN_PREFETCH_EDITOR: 'NULL'
+  che.core.jsonrpc.processor.major.max.pool.size: 120
```
garagatyi: Would be great to have some clarity on why this value is used. With default minor endpoint pool size equal to 100 we have 220 threads overall for 2 endpoints. I'm not sure how to evaluate whether our RAM size for Che is enough for 220 threads or not. Do you have any hints on how to evaluate that?

ScrewTSW:
[test]

ibuziuk:
@ScrewTSW why do you keep testing PR if it is requires changes :-) ?

openshift/rh-che.config.yaml
ScrewTSW: This is major endpoint setting explicitly. The minor will be set to much lower value once I complete the testing on beta-2. I have set that value that high, because the threads don't consume that much ram and where ineffective at consuming the messages in queue when set to lower values.

openshift/rh-che.app.yaml
ScrewTSW: :+1:

ScrewTSW: fixed.

ScrewTSW:
> @ScrewTSW why do you keep testing PR if it requires changes :-) ?

Don't worry. I wasn't continuing testing it with the issues. I saw the console output and input from Alex. I moved on to testing the values for beta-2 version.

skabashnyuk:
@ScrewTSW can you explain what is the plan with this PR?

ScrewTSW:
> @ScrewTSW can you explain what is the plan with this PR?

Hello Sergii.
The purpose of this PR is to test values I have interpolated from my findings so far and to see if I can finetune them to the new mechanics of the major and minor websockets.

pom.xml
```diff
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>che-parent</artifactId>
         <groupId>org.eclipse.che</groupId>
-        <version>7.0.0-beta-1.0</version>
+        <version>7.0.0-beta-2.0</version>
```
ibuziuk: Please, rebase on the master which is already updated to 7.0.0-beta-2.0

openshift/rh-che.config.yaml
```diff
@@ -72,3 +71,9 @@ data:
   CHE_WORKSPACE_STORAGE: "/home/user/che/workspaces"
   CHE_WORKSPACE_PLUGIN__BROKER_VSCODE_IMAGE: "eclipse/che-vscode-extension-broker:v0.13.0"
   CHE_FABRIC8_CDN_PREFETCH_EDITOR: 'NULL'
+  CHE_CORE_JSONRPC_PROCESSOR__MAX__POOL__SIZE: '60'
+  CHE_CORE_JSONRPC_PROCESSOR__CORE__POOL__SIZE: '10'
+  CHE_CORE_JSONRPC_PROCESSOR__QUEUE_CAPACITY: '30000'
+  CHE_CORE_JSONRPC__MINOR__PROCESSOR__MAX__POOL__SIZE: '60'
+  CHE_CORE_JSONRPC__MINOR__PROCESSOR__CORE__POOL__SIZE: '5'
+  CHE_CORE_JSONRPC__MINOR__PROCESSOR__QUEUE__CAPACITY: '10000'
```
ibuziuk: is this config optimal for 4GB RAM for che server assuming that  `-XX:MaxRAMFraction=2 ` is set for che server java opts

ibuziuk:
NOTE that all cm values should be updated on dsaas / dsaas-stg accordingly - https://gitlab.cee.redhat.com/service/app-interface#manage-openshift-resources-via-app-interface-openshiftnamespace-1yml

skabashnyuk:
@ScrewTSW can you explain what is the goal of your study? What and why you want to adjust?

ibuziuk:
@ScrewTSW I believe we should merge this PR only once there would be clear documentation about what problem we are solving and why we need to adjust `CHE_CORE_JSONRPC_*`  in comparison with upstream. As I understand it is NOT possible to reproduce OOM against che 7 beta 2 which is currently deployed on prod, so there should be clarity about what issue are we solving by adjusting CHE_CORE_JSONRPC_*

Katka92:
[test]

ScrewTSW:
> @ScrewTSW can you explain what is the goal of your study? What and why you want to adjust?

Hello.

Our original rh-che settings for the queue are
```
CHE_CORE_JSONRPC_PROCESSOR__MAX__POOL__SIZE: "2000"
```
, and don't include the queue setting which is left on a default value of 1M.

These values were causing us to run out of memory very quickly when the server was under load.
My study is trying to set all the values so that they fit our available 4GB of memory without causing an OOM.

pom.xml
ScrewTSW: done.

openshift/rh-che.config.yaml
ScrewTSW: The latest values have been agreed on by Sergii Kabashniuk.
They should be correct for the server from version 7-beta-2 and forwards.

ScrewTSW:
I was able to reproduce the crash on the OSIO configmap settings. Given enough load would strike the server, it could crash (it's enough to overwhelm 3 threads currently. The queue will keep growing and the threads will never increase to the max setting. From my observation, new threads are only being spawned when the queue is full, which will never happen on OSIO as the server crashes and restarts when there are about 1 milion messages stored for execution while the queue size is 10 million).

ScrewTSW:
Opened a PR in `app-interface` https://gitlab.cee.redhat.com/service/app-interface/merge_requests/274

ibuziuk:
I have created a separate issue for verifying if the default setting are good / not good for che.openshift.io https://github.com/redhat-developer/rh-che/issues/1337

aditya-konarde:
@ScrewTSW Bonus points for the graphs on the issue :D

PR #1294, opened 2019-03-15
Updating rh-che to the 7.0.0-beta-2.0 upstream version

ibuziuk:
### What does this PR do?
Updating rh-che to the 7.0.0-beta-2.0 upstream version

### What issues does this PR fix or reference?
https://github.com/redhat-developer/rh-che/issues/1261

### How have you tested this PR?
supposed to be tested via CI

PR #1326, opened 2019-03-22
Adding 'CHE_WEBSOCKET_ENDPOINT__MINOR' env var that was introduced in  7 beta 2 version

ibuziuk:
### What does this PR do?
Adding 'CHE_WEBSOCKET_ENDPOINT__MINOR' env var that was introduced in  7 beta 2 version

Related to https://github.com/redhat-developer/rh-che/issues/1261
upstream issue https://github.com/eclipse/che/pull/12673

supposed to be tested by PR check Ci


ibuziuk:
Once merged that should hopefully make #1294 to pass
Second step would be updating cms on dsaas / dsaas-stg

PR #1327, opened 2019-03-25
Add info about images

Katka92:
### What does this PR do?
There are several images used for different type of tests. This PR adds information about which images are used, what these images contain and when are they built. 
The part of README about running e2e tests is removed as it is not valid.


Katka92:
@ScrewTSW please confirm that there are all images used in tests.

PR #1328, opened 2019-03-25
Adding support of `CHE_WEBSOCKET_ENDPOINT__MINOR` endpoint for deployment against dev cluster

ibuziuk:
### What does this PR do?
Adding support of `CHE_WEBSOCKET_ENDPOINT__MINOR` endpoint for deployment against dev cluster

- https://github.com/redhat-developer/rh-che/pull/1326
- should potentially fix https://github.com/redhat-developer/rh-che/pull/1294#issuecomment-475910313

Supposed to be tested by PR check CI


ibuziuk:
tested manually against free-stg via http://rhche-prcheck-1328.devtools-dev.ext.devshift.net startup works just fine. Unfortunately, it is not clear why CI is so unstable

PR #1334, opened 2019-03-27
Remove VS Code broker image config

garagatyi:
### What does this PR do?
Remove VS Code broker image config because it is outdated
and current version of upstream provides newer version of this broker.
See https://github.com/eclipse/che/blob/7.0.0-beta-2.0/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L548


ibuziuk:
@garagatyi do we have dedicated PRs for cm updates?

garagatyi:
@ibuziuk I was going to create PRs for that after merge of this PR

PR #1335, opened 2019-03-27
Adding websocket env var option for docker run parameters

ScrewTSW:
### What does this PR do?
Adds a `-e "WEBSOCKET="` variable into the docker run parameters

-

Manually built and tested the image.


Katka92:
I'm not much familiar with the code but seems good.

PR #1340, opened 2019-04-01
Add Che 7 test to PR check suite.

Katka92:
### What does this PR do?
Add test for Che 7 workspace start up to the PR check suite. 
This test was run as part of flaky test suite and succeeded in last 16 runs.


ibuziuk:
[test]

ibuziuk:
[test]

Katka92:
Test started to fail because of UI change. Need to fix that before merging.
Issue: developer/che-functional-tests/issues/469

ibuziuk:
@Katka92 could we potentially add test for content assist in che 7 or this functionality is not yet availble in the test suite?

Katka92:
@ibuziuk I've slightly changed the behaviour of this test, but still it only creates and starts Che 7 workspace, wait until it loads and open Dialog About. This is basically the same functionality that we have for Che 6 Dialog About tests. 
That means that in actual selenium test we don't test import of project, opening files etc. There are no upstream selenium tests doing that, so we have two options here - we can write our own test, or we can adopt cypress tests that should implement happy path. I'm working on cypress tests from yesterday, the workflow here is: start workspace, wait until started, open files, wait for java language server initialization, test content assist, stop and remove workspace. 
So I would not add test to selenium, but focus on adopting cypress tests and once we find out that we miss something in upstream cypress tests, we can extend tests on our side.

Katka92:
[test]

ibuziuk:
@Katka92 thanks for clarifications. 

> So I would not add test to selenium, but focus on adopting cypress tests and once we find out that we miss something in upstream cypress tests, we can extend tests on our side.

yeah, makes sense. cc: @rhopp

rhopp:
I'm aware of this work. But I'm afraid we won't be able to consume upstream completely without customization. These tests should be whole end to end scenario - and to do that, there are some parts which needs to be customized (login for example)

Katka92:
[test]

Katka92:
[test]

Katka92:
I would like to test it once more to see, if issues are really gone. If that will be green, I'll merge that.

PR #1348, opened 2019-04-08
Add print USERNAME for Jenkins.

Katka92:
### What does this PR do?
Print username in format ```3 letters + space + the rest of letters```. If USERNAME is saved in Jenkins credentials it is not shown in a job (shown as set of ``` * ```) and is hard to track which user is used. 

Example:
for user ```osiotest2``` there will be shown ```User name printed in format: 3 first letters, space, the rest of letters.    osio test2```.


ibuziuk:
@rhopp any security concerns?

functional-tests/devscripts/run_tests.sh
```diff
@@ -54,6 +54,10 @@ else
 	TEST_SUITE="simpleTestSuite.xml"
 fi
 
+#This format allows us to see username even if it is placed in Jenkins credential store. 
+USERNAME_TO_PRINT="${USERNAME:0:3} ${USERNAME:3:${#USERNAME}}"
+echo "User name printed in format: 3 first letters, space, the rest of letters.    $USERNAME_TO_PRINT"
```
amisevsk: Shouldn't this be 4 letters?

Katka92: I don't know why but if you want, I'm ok with 4 letters too. :) Should I change it?

amisevsk: Ah no, I was just confused, never mind.

`User name printed in format: 3 first letters, space, the rest of letters. osio test2.` would be `osi otest2` in this case, right?

Katka92: Heh, yes, this is my typo. Sorry for confusion.

rhopp:
By printing only username I don't think we are under any security risk...

PR #1367, opened 2019-04-16
Remove dependency duplicate

garagatyi:
### What does this PR do?
Remove dependency duplicate


ibuziuk:
funny, I thought maven is smart enough to detect this at build time

garagatyi:
@ibuziuk maven does detect this, but log only warning that we weren't noticing successfully

PR #1386, opened 2019-04-25
Fix gathering logs

Katka92:
### What does this PR do?
The issue with gathering logs by ```oc``` is that the current approach does not allow us to follow redirects when downloading openshift client. If we would like to change that approach we have to change upstream. We decided to use a workaround - download ```oc``` directly from bash before running scripts. 
Once logs are not needed, it can be deleted. 

localy


Katka92:
@tdancs Because test will check the ```/tmp``` location. If it will not find the ```oc``` here, it will try to install it and fail. So we will install it before test is running, test will only check it and continue.

PR #1391, opened 2019-04-29
Fix log gathering

Katka92:
### What does this PR do?
Move the code to docker image entrypoint. As the tests are run as a docker image, the ```oc``` installed in ```prepare-env.sh``` is useless and this should be done in docker image. 
@tdancs could you please verify, that the docker image is automatically refreshed so no extra action for the updated docker image is needed?


dockerfiles/functional-tests/docker-entrypoint.sh
```diff
@@ -56,6 +56,11 @@ if [ -d "/root/che" ]; then
   cd /root/rh-che/
 fi
 
+#next lines added to be able to gather logs for issue https://github.com/redhat-developer/che-functional-tests/issues/476
```
rhopp: I'm thinking whether it would be better to download latest version, or stick with some specific one...
Anyway I think downloading at least 3.11 shouldn't hurt  ;-)

And also... Wouldn't it be better to move this code to dockerfile? That way we wouldn't need to download oc every time we are running this container.

Katka92: Version updated, code moved to dockerfile.

Katka92:
[test]

Issue #1320, opened 2019-03-20
Fix issue with Getting Started in the email registration template


ibuziuk:
Fix issue with Getting Started in the email registration template https://user-images.githubusercontent.com/1461122/53915886-1df34b80-4061-11e9-80e3-128495545f01.png

ibuziuk:
I believe this issue would be fixed during major branding update related to openshift.io -> che.openshift.io transition

ibuziuk:
@Katka92 @ScrewTSW I know you checked the registration flow just after the registration. Could you please shed some light on how email template looks now for a newly provisioned users? Also would be great to provide info about registration flow when user is provisioned for the first time via factory from https://www.eclipse.org/che/getting-started/cloud/ e.g. Spring Boot - https://che.openshift.io/f?id=factorypedk0ijio2y27tzf 
what is the registration UX in this case

Katka92:
@ibuziuk I've created an account for a red hat user yesterday on production and email looks like that:
![Screenshot from 2019-04-24 08-04-07](https://user-images.githubusercontent.com/29278800/56635986-b53f4d00-6667-11e9-9222-189447d2dc4b.png)

So today I created new user using the link to Spring Boot factory you posted. It redirects me to the login page, so I click ``` Create one now ```. I filled information and created new account. I got an email:

![Screenshot from 2019-04-24 08-15-56](https://user-images.githubusercontent.com/29278800/56640133-84b0e080-6672-11e9-83a1-fb0f02937b9e.png)

Which sends me to verification:
![Screenshot from 2019-04-24 08-16-11](https://user-images.githubusercontent.com/29278800/56640154-94c8c000-6672-11e9-85e7-4fbff85a3743.png)

Then informs me that account was verified:
![Screenshot from 2019-04-24 08-16-40](https://user-images.githubusercontent.com/29278800/56640175-a14d1880-6672-11e9-96b2-642fb71183f8.png)

I was asked to fill a captcha, but the css for this page seems broken:
![Screenshot from 2019-04-24 08-16-58](https://user-images.githubusercontent.com/29278800/56640226-c3df3180-6672-11e9-9a6c-b44bda5de514.png)

After filling captha, I continued. Then I was waiting for provisioning of the user.
![Screenshot from 2019-04-24 08-17-42](https://user-images.githubusercontent.com/29278800/56640295-f1c47600-6672-11e9-8b42-14c59a493439.png)


[...]

PR #1404, opened 2019-05-09
Fix end 2 end flow

davidfestal:
### What does this PR do?

This PR fixes the End2End flow that was broken after the Unified RHD login rollout.

This PR fixes issue described in comment https://github.com/redhat-developer/rh-che/issues/1320#issuecomment-486262528 

Yes, on Dev cluster, and also deployed a temporary fix on production (that didn't need to create a new RhChe release, and is explained here: https://gitlab.cee.redhat.com/service/app-interface/merge_requests/527)


Katka92:
[test]

PR #1406, opened 2019-05-13
Update route tests

Katka92:
### What does this PR do?
Change logic for checking route availability - now uses times in seconds, not tries.
There are three possible failures now - SOFT says that the route was started after soft timeout (now set to 30 seconds). HARD says that the route was not started in hard timeout (now set to 150 seconds). FLAPPING says that there were some another response then 200 from a moment when route was marked as available (set to 20 tries). 

This PR is fixing README too.


ibuziuk:
so will test be red or I will need to find info in the logs info about soft timeout?

Katka92:
@ibuziuk The test will be red in all 3 cases. At the bottom of the consol output there will be the reason why the test failed.

PR #1390, opened 2019-04-28
Adapt to new plugin notation

garagatyi:
### What does this PR do?
Do not merge until PR https://github.com/eclipse/che/pull/13204 is in the version of upstream RhChe uses.
Note: this PR doesn't adapt resources prefetcher by @davidfestal

PR #1407, opened 2019-05-13
Adapt dev-guide to new plugin notation

amisevsk:
### What does this PR do?
This PR is the dev-guide changes from https://github.com/redhat-developer/rh-che/pull/1390. The remaining adaptations have already been made elsewhere.


amisevsk:
Originally had wrong base for the PR, should be better now.

PR #1411, opened 2019-05-15
Upgrade route tests

Katka92:
### What does this PR do?
Add reporting to zabbix, enhance messages shown at the end for tests.


Katka92:
[test]

Katka92:
[test]

functional-tests/devscripts/route_tests.sh
```diff
@@ -39,8 +44,30 @@ function checkPrereq {
...
+ZABBIX_SERVER="zabbix.devshift.net"
+ZABBIX_PORT="9443"
+case "$OC_CLUSTER_URL" in
+    *"2a"*) ZABBIX_HOST="qa-starter-us-east-2a"
+            ;;
+    *"1a"*) ZABBIX_HOST="qa-starter-us-east-1a"
+	        ;;
+    *"1b"*) ZABBIX_HOST="qa-starter-us-east-1b"
+            ;;
+    *) ZABBIX_HOST="qa-starter-us-east-2"
```
ppitonak: IMHO this could be dangerous. I suggest to list `2` explicitly and in the case of `*` would set the host to something like `unknown`, `undefined`, `null` or something similar.

PR #1427, opened 2019-05-27
Fixing CDN pre-fetching of the 'eclipse/che-theia/next' editor

ibuziuk:
### What does this PR do?
Fixing CDN pre-fetching of the 'eclipse/che-theia/next' editor - on dsaas / dsaas-stg `CHE_FABRIC8_CDN_PREFETCH_EDITOR` is set to obsolete `org.eclipse.che.editor.theia:1.0.0`
Related Merge Request - https://gitlab.cee.redhat.com/service/app-interface/merge_requests/650 (merged)

Currently, CDN pre-fetching is broken on prod and prod-preview:

> Unable to find Theia CDN resources to cache 
> Object { data: "Plugin 'org.eclipse.che.editor.theia:1.0.0' has incorrect format. Should be: 'registryURL/publisher/name/version' or 'publisher/name/version'"

Tested locally via minishift-addon


ibuziuk:
@davidfestal for some reason when testing against dev cluster call to GET `api/cdn-support/paths` results in an empty array. Do you happen to know if this might be expected?

plugins/fabric8-cdn-support/src/test/java/com/redhat/che/cdn/HttpYamlDownloaderTest.java
```diff
@@ -0,0 +1,38 @@
...
+  private final String EDITOR_PLUGIN_URL =
+      "https://che-plugin-registry.openshift.io/v2/plugins/eclipse/che-theia/next/";
+  private final String EDITOR_PLUGIN_VERSION = "next";
+  private final String EDITOR_PLUGIN_REPOSITORY = "https://github.com/eclipse/che-theia";
+  private final String EDITOR_PLUGIN_CATEGORY = "Editor";
+
+  @Test
+  public void getYamlResponseAndParse() throws URISyntaxException, IOException {
+    HttpYamlDownloader httpYamlDownloader = new HttpYamlDownloader();
+    PluginMeta pluginMeta = httpYamlDownloader.getYamlResponseAndParse(new URI(EDITOR_PLUGIN_URL));
```
amisevsk: I don't think coding network requests into test cases is a good idea (and is the reason I didn't include any tests to begin with). This could potentially be very fragile.

plugins/fabric8-cdn-support/src/main/java/com/redhat/che/cdn/CdnSupportService.java
```diff
@@ -265,8 +265,8 @@ private PluginMeta getEditorMeta() throws InfrastructureException {
         pluginFQN.getRegistry() == null
             ? pluginRegistry.clone()
             : UriBuilder.fromUri(pluginFQN.getRegistry());
-
-    URI editorURI = uriBuilder.path(pluginFQN.getId()).build();
+    // slash after 'id' is required to avoid '301 permanently moved' http error
+    URI editorURI = uriBuilder.path(pluginFQN.getId() + "/").build();
```
amisevsk: The technically correct URL here would be 
```java
uriBuilder.path(pluginFQN.getId()).path("meta.yaml").build()
```
We're getting a `301` because we're querying a directory, not a specific file. I believe it's only returning the meta.yaml on this path because the meta.yaml is typically the only file in that directory (which is not required -- i.e. this could stop working if someone adds a `README.md` to a plugin's directory).

amisevsk:
(Apologies the initial implementation here was so busted, that's my fault)

plugins/fabric8-cdn-support/src/test/java/com/redhat/che/cdn/HttpYamlDownloaderTest.java
ibuziuk: why do you think having integration tests is a bad idea? this could be broken every now and than due to changes in the registry but we would catch it immediately and fail fast IMO

plugins/fabric8-cdn-support/src/main/java/com/redhat/che/cdn/CdnSupportService.java
ibuziuk: good, point - we probably need to fall back on `meta.yaml`

ibuziuk:
after a few more tests fix seems to work both locally and against dev cluster:

image

image

plugins/fabric8-cdn-support/src/main/java/com/redhat/che/cdn/CdnSupportService.java
ibuziuk: fixed

ibuziuk:
Merge Request for updating prod cm - https://gitlab.cee.redhat.com/service/app-interface/merge_requests/658

plugins/fabric8-cdn-support/src/test/java/com/redhat/che/cdn/HttpYamlDownloaderTest.java
amisevsk: I have no problem with the test existing but don't really like it being a part of the main build process; it makes compilation depend on an internet connection and, if there are enough such tests, would slow builds down significantly.

plugins/fabric8-cdn-support/src/main/java/com/redhat/che/cdn/CdnSupportService.java
```diff
@@ -265,8 +266,7 @@ private PluginMeta getEditorMeta() throws InfrastructureException {
         pluginFQN.getRegistry() == null
             ? pluginRegistry.clone()
             : UriBuilder.fromUri(pluginFQN.getRegistry());
-
-    URI editorURI = uriBuilder.path(pluginFQN.getId()).build();
+    URI editorURI = uriBuilder.path(pluginFQN.getId()).path(META_YAML).build();
```
amisevsk: Minor quibble: since we're building the full URL here, we may as well include the plugins part of the path. This isn't too important though.
```suggestion
    URI editorURI = uriBuilder..path("plugins").path(pluginFQN.getId()).path(META_YAML).build();
```

ibuziuk: @amisevsk good point, but I will probably include this into upcoming PR for beta 5 update

PR #1432, opened 2019-05-29
Route test - change message

Katka92:
### What does this PR do?
Change the report at the end of route test to make more sense.


Katka92:
@amisevsk Thank you I didn't notice. I'll fix that :)

PR #1441, opened 2019-06-03
Increasing `CHE_LIMITS_WORKSPACE_ENV_RAM` to 8GB

ibuziuk:
### What does this PR do?
Increasing `CHE_LIMITS_WORKSPACE_ENV_RAM` to 8GB (preventing UD validation to fail for users which have quota more than 3Gb RAM)

N/A

dev cluster


ibuziuk:
behaviour with new RAM limits:
- UD validation will fail only with RAM > 8 GB

image

- Che 6 Workspace with RAM setting more > 3 GB and < 8GB will fail on workspace creation step
image

- Che 7 workspace creation with RAM setting more > 3 GB and < 8GB will fail after brokering process:

image
image

With this change we will not be able to fail fast, but on the other hand `CHE_LIMIT_WORKSPACE_ENV_RAM` validates only main container and does not take into account RAM required for side-car containers and jwt-proxy.
So, Currently from UD perspective, it is possible to create `java-maven` Che 7 workspace with 2 GB of RAM, but this will eventually fail with the very same quota issue.

@l0rd do we have an issue for improving RAM calculation for Che 7 workspace during UD validation? For me it sounds like there are 2 issues:
- we need to calculate precise RAM required for a given Che 7 workspace
- we need to match this RAM with quota in the namespace

or you think we should not have any UD validation at all and fall back on k8s ?

l0rd:
@ibuziuk this is the issue for RAM calculation: https://github.com/eclipse/che/issues/11424

ibuziuk:
MR with cm update - https://gitlab.cee.redhat.com/service/app-interface/merge_requests/692/

PR #1472, opened 2019-07-02
Enabled selenium tests to use old stack-based approach

ScrewTSW:
### What does this PR do?
Allows us to run tests both on older versions of che and latest Che7 RC2+
Old Che releases were using the stack based approach, Che7 RC2 introduced devfiles which required some changes.

These changes are necessary in order to keep our periodic tests and PR-check on production running.

https://github.com/redhat-developer/che-functional-tests/issues/539

manual testing, pr-check

### DIFF
--- a/functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
+++ b/functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
@@ -12,6 +12,7 @@
 package com.redhat.che.functional.tests;
 
 import com.google.inject.Inject;
+import com.google.inject.name.Named;
 import java.util.concurrent.ExecutionException;
 import org.eclipse.che.selenium.core.workspace.TestWorkspace;
 import org.eclipse.che.selenium.pageobject.CodenvyEditor;
@@ -26,10 +27,17 @@ public abstract class RhCheAbstractTestClass {
 
   private static final Logger LOG = LoggerFactory.getLogger(RhCheAbstractTestClass.class);
 
+  public static final String CHE_PROD_PREVIEW_URL = "che.prod-preview.openshift.io";
+  public static final String CHE_PROD_URL = "che.openshift.io";
+
   @Inject private Ide ide;
   @Inject private NotificationsPopupPanel notificationsPopupPanel;
   @Inject private ProjectExplorer projectExplorer;
   @Inject private CodenvyEditor editor;
+
+  @Inject
+  @Named("che.host")
+  String cheHost;
 
   public void checkWorkspace(TestWorkspace workspace) throws Exception {
     try {
@@ -73,5 +81,13 @@ public abstract class RhCheAbstractTestClass {
 
   public void closeBrowser() {
     ide.close();
+  }
+
+  public boolean isProd() {
+    return CHE_PROD_URL.equals(cheHost);
+  }
+
+  public boolean isPreview() {
+    return CHE_PROD_PREVIEW_URL.equals(cheHost);
   }
 }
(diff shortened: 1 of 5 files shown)


functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
```diff
@@ -74,4 +82,12 @@ public void closeFilesAndProject() {
   public void closeBrowser() {
     ide.close();
   }
+
+  public boolean isProd() {
+    return cheHost.equals(CHE_PROD_URL);
```
ibuziuk: I would change it to CHE_PROD_URL.equals(cheHost) to avoid NPE

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
```diff
@@ -74,4 +82,12 @@ public void closeFilesAndProject() {
...
+  }
+
+  public boolean isPreview() {
+    return cheHost.equals(CHE_PROD_PREVIEW_URL);
```
ibuziuk: I would change it to `CHE_PROD_PREVIEW_URL.equals(cheHost)` to avoid NPE

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
ScrewTSW: both variables must always be set. if they are not the test would never run and fail.

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
ScrewTSW: Same as my comment above

ScrewTSW:
Fixed issue in logic (inverted if statement condition)

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
ScrewTSW: refactored.

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
ScrewTSW: refactored.

functional-tests/src/test/java/com/redhat/che/functional/tests/RhCheAbstractTestClass.java
```diff
@@ -26,11 +27,18 @@
...
+  public static final String CHE_PROD_URL = "che.openshift.io";
+
   @Inject private Ide ide;
   @Inject private NotificationsPopupPanel notificationsPopupPanel;
   @Inject private ProjectExplorer projectExplorer;
   @Inject private CodenvyEditor editor;
 
+  @Inject
+  @Named("che.host")
+  String cheHost;
```
ibuziuk: @ScrewTSW could we avoid injecting env var in the multiple places? Could we have some class that would be in charge of checking the infrastructure with 2 methods "isProd" / "isPreview" that would be injected. Currently we have multiple duplications of methods in different classes e.g. :

  public boolean isProd() {
    return CHE_PROD_URL.equals(cheHost);
  }

ScrewTSW: I could maybe create a helper class, but in this case it has been discussed with @Katka92 and @rhopp 
This class cannot extend the Abstract test class because this one is che7 based selenium test, which uses completely different page fragments and using the AbstractTestClass would break it (it would try to `@Inject` che6 based fragments and cause test failure) hence the repeated logic.

The question is, is it really worth investing extra time and effort into something that will be deleted in 2-3 weeks?

ibuziuk: I find it really easy to implement e.g. https://github.com/redhat-developer/che-starter/blob/master/src/main/java/io/fabric8/che/starter/util/CheServerUrlProvider.java#L21 but up to you if this code will be obsolete and deleted in a couple of weeks

Issue #1442, opened 2019-06-03
remove a workaround with jar skipping  in `assembly-wsmaster-war`


ibuziuk:
Once is merged https://github.com/eclipse/che/pull/13464 and a new version is released we need to remove a  workaround with jar skipping on our end 36897e4#diff-d79082aaff35c41c21db3933361c39efR148

Issue #1418, opened 2019-05-21
Need to update prod to the latest 7.0.0-beta-5.0 upstream


ibuziuk:
Related PR - https://github.com/redhat-developer/rh-che/pull/1419

ibuziuk:
Related PR - https://github.com/redhat-developer/rh-che/pull/1419
PR for promotion to prod - https://github.com/openshiftio/saas-openshiftio/pull/1292

ibuziuk:
Prod has been updated to 7 beta 5. Closing

PR #1480, opened 2019-07-08
rh-che #1418: Removing workaround with jar skipping in

ibuziuk:
### What does this PR do?
rh-che #1418: Removing workaround with jar skipping in

https://github.com/redhat-developer/rh-che/issues/1442

PR check CI


ibuziuk:
@davidfestal we had a problem with build and had to add this workaround. It was then reworked on the upstream side as part of the https://github.com/eclipse/che/pull/13464

PR #1487, opened 2019-07-18
[E2E] Add logs to investigate project import failure

Katka92:
### What does this PR do?
Echo response from /api/workspace/<namespace>:<workspace-name> when test fails on import project. Project is not imported and we want to know if it is in workspace specification - so it should be visible imported and visible in UI, or if it is not in workspace definition, than maybe project is not added by test and test should be fixed. 

https://github.com/redhat-developer/che-functional-tests/issues/547

Locally against prod-preview


e2e-saas/tests/HappyPath.spec.ts
```diff
@@ -54,7 +57,19 @@ suite('RhChe E2E', async () => {
...
 
         test('Wait project imported', async () => {
-            await projectTree.waitProjectImported(sampleName, 'src');
+            try {
+                await projectTree.waitProjectImported(sampleName, 'src');
+            } catch (err) {
+                if (!(err instanceof error.TimeoutError)) {
+                    throw err;
+                }
+                var rest: rm.RestClient = new rm.RestClient('rest-samples');
```
rhopp: What is this "rest-samples"? Seems bit weird...

e2e-saas/tests/HappyPath.spec.ts
```diff
@@ -10,6 +10,9 @@
 import 'reflect-metadata';
 import { TYPES, CLASSES, TestConstants, ILoginPage, Dashboard, Editor, Ide, NameGenerator , NewWorkspace, ProjectTree } from 'e2e';
 import { rhCheContainer,  } from '../inversify.config';
+import { error } from 'selenium-webdriver';
+import * as rm from 'typed-rest-client/RestClient';
```
rhopp: `rm` seems weird as a name for rest client variable/naming :-D

Katka92:
Old tests fails on `Timeout from renderer` (test issue) and typescript tests passed. Merging.

PR #1501, opened 2019-07-24
[E2E] Set precise version of tests.

Katka92:
### What does this PR do?
Set precise version of upstream tests that RH-Che tests are depending on. Test should be broken without setting this version when rc-4 is released.


Katka92:
Failed on old test intermittent issue. New e2e tests passed - merging.

PR #1507, opened 2019-07-25
[E2E] Add versioning

Katka92:
### What does this PR do?
It can happen that we will have different versions of RH-Che on production, prod-preview or PR check (e.g. when testing new version) so we need some mechanism how to differentiate between them and how to use correct version.

### DIFF
--- a/.ci/cico_build_func_dep.sh
+++ b/.ci/cico_build_func_dep.sh
@@ -57,18 +57,20 @@ date '+DEP-TIMESTAMP: %d.%m.%Y - %H:%M:%S %Z'
 
 DOCKERFILE="e2e-saas/"
 DOCKER_IMAGE="rh-che-e2e-tests"
+TAG=$(getVersionFromPom)
 DOCKER_IMAGE_URL="${REGISTRY}/openshiftio/${NAMESPACE}-${DOCKER_IMAGE}"
 
 echo "Building docker image for functional tests (used e.g. in periodic tests)."
-docker build -t ${DOCKER_IMAGE_URL}:${TAG} ${DOCKER_PATH}${DOCKERFILE}
+docker build -t ${DOCKER_IMAGE_URL}:${TAG} --build-tag TAG=$TAG ${DOCKER_PATH}${DOCKERFILE}
 if [ $? -ne 0 ]; then
   echo 'Docker Build Failed'
   exit 2
 fi
 
-echo "Build was successful, pushing image ${DOCKER_IMAGE_URL}:${SHORT_HASH}"
+echo "Build was successful, pushing image ${DOCKER_IMAGE_URL} with tags ${TAG}, ${SHORT_HAS} and latest"
 date '+DEP-TIMESTAMP: %d.%m.%Y - %H:%M:%S %Z'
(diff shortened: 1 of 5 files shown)


.ci/cico_build_func_dep.sh
```diff
@@ -57,18 +57,20 @@ date '+DEP-TIMESTAMP: %d.%m.%Y - %H:%M:%S %Z'
 
 DOCKERFILE="e2e-saas/"
 DOCKER_IMAGE="rh-che-e2e-tests"
+TAG=$(getVersionFromPom)
 DOCKER_IMAGE_URL="${REGISTRY}/openshiftio/${NAMESPACE}-${DOCKER_IMAGE}"
 
 echo "Building docker image for functional tests (used e.g. in periodic tests)."
-docker build -t ${DOCKER_IMAGE_URL}:${TAG} ${DOCKER_PATH}${DOCKERFILE}
+docker build -t ${DOCKER_IMAGE_URL}:${TAG} --build-tag TAG=$TAG ${DOCKER_PATH}${DOCKERFILE}
```
rhopp: Shouldn't this be `--build-arg`?

.ci/functional_tests_utils.sh
```diff
@@ -102,3 +105,75 @@ function archiveArtifacts() {
...
 }
+
+function getVersionFromPom() {
+  version=$(mvn org.apache.maven.plugins:maven-help-plugin:evaluate -q -Dexpression=project.parent.version -DforceStdout)
+  echo $version
+}
+
+function getActiveToken() {
+  rm -rf cookie-file loginfile.html
+  if [[ "$USERNAME" == *"preview"* ]]; then
```
rhopp: What about just setting `$preview` value in this if, then both branches of this if, could be squashed into one:
```
    response=$(curl -s -g -X GET --header 'Accept: application/json' "https://api.${preview}openshift.io/api/users?filter[username]=$USERNAME")
    data=$(echo "$response" | jq .data)
    if [ "$data" == "[]" ]; then
      exit 1
    fi     
```

.ci/functional_tests_utils.sh
```diff
@@ -102,3 +105,75 @@ function archiveArtifacts() {
...
+    fi
+    found=$(echo "$url" | grep "token_json")
+    if [[ ${#found} -gt 0 ]]; then
+      #some redirects were done and token was found as a part of url
+      break
+    fi
+  done
+  set -e
+
+  #substract active token
```
rhopp: exctract? :-D

functional-tests/devscripts/run_tests.sh
```diff
@@ -91,75 +91,106 @@ echo "User name printed in format: 3 first letters, space, the rest of letters.
...
+    echo "[ERROR]: Could not find version in pom.xml."
+    exit 1
+  fi
+  rhche_image="quay.io/openshiftio/rhchestage-rh-che-e2e-tests:${version}"
+
+  docker pull $rhche_image
+  RESULT=$?
+
+  #reuse image if exists or build new image for test
+  if [[ $RESULT != 0 ]]; then
```
rhopp: I'm not sure... wouldn't it be nicer to just use the command in if?
```
if docker pull "$rhche_image"; then
```
? 
Your call ;-)

Issue #1536, opened 2019-08-01
Update export of classes in upstream tests


Katka92:
To be able to reuse/override all the files from upstream tests, these classes needs to be exported. Currently we have workaround for that - we are replacing this upstream file with our own one. Fix that so we can remove this workaround with the next released version.
File that needs to be updated https://github.com/eclipse/che/blob/master/e2e/index.ts

Katka92:
All PR's merged. Closing.

PR #1537, opened 2019-08-01
[E2E] Add workaround for exporting classes from upstream

Katka92:
### What does this PR do?
Add workaround for exposing classes from upstream. Will create appropriate PR in upstream so we will not need this workaround in next released version (issue for that: https://github.com/redhat-developer/rh-che/issues/1536).


ibuziuk:
@Katka92 please link the upstream PR to the issue

PR #1562, opened 2019-08-15
[Periodic tests] Add yum clean to fix problems with epel.

Katka92:
### What does this PR do?
All tests are failing on:

According to centos wiki, yum clean should help. https://wiki.centos.org/yum-errors


Katka92:
None of workarounds seems to work. Issue was opened, we would need to wait for a fix https://bugs.centos.org/view.php?id=16337.

PR #1564, opened 2019-08-15
Trying to fix epel in tests

Katka92:
### What does this PR do?
Trying to fix epel problem https://bugs.centos.org/view.php?id=16337


Katka92:
[test]

Katka92:
Problem with epel is fixed. Tests are failing due to LS problems - will solve that in another PR.

Issue #1575, opened 2019-08-28
[E2E] Fix tests for 7.1.0


Katka92:
Fix tests in a PR that is updating prod-preview to 7.1.0 https://github.com/redhat-developer/rh-che/pull/1573

Currently the tests are failing on
``` 
Unable to find image 'e2e_tests:latest' locally
docker: Error response from daemon: pull access denied for e2e_tests, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```

Katka92:
Docker issues seems to be fixed, incompatible code seems to be fixed. Need to verify before merge, but now dev cluster is under stress and tests are failing on all PRs due to that.

Katka92:
Tests for 7.1.0 were fixed. There are some follow up issues created. Closing.

PR #1608, opened 2019-09-09
[E2E] Fix LS test.

Katka92:
### What does this PR do?
Old LS test was failing on 7.1.0 version. Fixing the test using the same approach as is used in HappyPath.ts in upstream.

Part of https://github.com/redhat-developer/rh-che/issues/1575

Locally against prod-preview.


Katka92:
The LS test has passed. PR is ready for review.

Screenshot from 2019-09-09 11-52-30

Issue #1645, opened 2019-10-08
Adjust CHE_WORKSPACE_POOL_EXACT__SIZE in che-server configuration


skabashnyuk:
At this moment default values for WorkspaceSharedPool are too low. Its configured in "fixed" mode without cores_multiplier. I think it's too low. On the other hand changing this configuration in "cached" mode is unsafe because of the unpredictable possibility of thread pool grow.   I suggest setting ```CHE_WORKSPACE_POOL_EXACT__SIZE``` to some reasonable value of parallelism. Let's say ```30```

skabashnyuk:
CC @amisevsk @l0rd @ibuziuk

PR #1646, opened 2019-10-08
Adjust CHE_WORKSPACE_POOL_EXACT__SIZE to better handle workspace startup load

skabashnyuk:
### What does this PR do?
Adjust CHE_WORKSPACE_POOL_EXACT__SIZE to better handle workspace startup load

https://github.com/redhat-developer/rh-che/issues/1645

Not tested. That is just a theoretical assumption. 
to be able to guarantee results we are missing data from this issue
Monitoring & Tracing capability for multiple Thread pools of Che Server https://github.com/eclipse/che/issues/14601


amisevsk:
[test]

PR #1629, opened 2019-09-24
WIP: Fix compatibility tests

ScrewTSW:
### What does this PR do?
This PR re-enables compatibility tests, changing the logic around a little

### What issues does this PR fix or reference?
https://github.com/redhat-developer/che-functional-tests/issues/448

### How have you tested this PR?
ci.centos.org script runs

Issue #936, opened 2018-09-28
Create a CI job that would check RhChe with dependency to nightly upstream


garagatyi:
**Issue problem:** 
RhChe depends on tags of upstream Che. We usually update the tag after each new upstream release. But when we try to move to a new tag we quite often see incompatibilities that weren't addressed in RhChe after changes in upstream. Often this happens after upstream Che is released and fixes can be applied only in a new bugfix release. We also find out new problems too late to be able rapidly deliver new features on prod. 
We can improve this situation by creating a CI job that creates PR with current upstream nightly and runs it each day or two to check whether RhChe compiles against this version of the upstream and whether functional tests show no regression.

I'm not sure how our CI checks PRs right now but in case it merges PR branch to master on each run we can create a CI job that would just trigger regular PR check on a specific PR that we would maintain just for that purpose.
Another improvement we can do is to have a CI job for creating a PR:
- greps current RhChe version and removes branch uptream-check-<RhChe version>-SNAPSHOT, e.g. `uptream-check-6.11.0-SNAPSHOT`
- curl che pom from github and gets current nightly
- check whether there is a branch in RhChe with a name: uptream-check-<Che version>, e.g. `uptream-check-6.12.0-SNAPSHOT`
- if there is no branch - create it
- change upstream dependency version
- if the local repo is dirty (job runs first time for this particular upstream version) commit/push it to the repo
- create PR if it is not created yet
- comment `[test]` in the PR (credentials should be whitelisted on regular PR CI)

This CI would maintain fresh PR for each upstream version so we can have a discussion regarding errors for a particular Che version in a PR and we would have a PR for each version.

**Red Hat Che version:** 

version: (help/about menu)

 - [ ] I can reproduce it on latest official image

**Reproduction Steps:**

Describe how to reproduce the problem

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [ ] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)
   - [ ] k8s (include output of `kubectl version`)

l0rd:
@garagatyi such a job already exist https://ci.codenvycorp.com/view/All/job/rh-che-ci-master/

garagatyi:
@l0rd AFAIK it just checks that RhChe compiles, but not that it passes RhChe specific deployment and functional tests. Whereas it is not critical but it would make us almost 100% sure that we can gracefully migrate to the new version without any issues. 
[...]

PR #1647, opened 2019-10-08
Fix compatibility tests

ScrewTSW:
### What does this PR do?
This pull request fixes compatibility tests logic and allows them to run once again.

https://github.com/redhat-developer/che-functional-tests/issues/448
https://github.com/redhat-developer/rh-che/issues/936

manual testing on modified prcheck script
https://github.com/redhat-developer/rh-che/pull/1629

### DIFF
--- a/.ci/cico_rhche_compatibility_test.sh
+++ b/.ci/cico_rhche_compatibility_test.sh
@@ -5,15 +5,56 @@
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
 
-set -e
+set -e -m
 
+function checkPullRequest() {
+  che_version=$1
+  job_name=$2
+  branch=$3
+
+  RELATED_PR_TITLE="Update to $(echo $che_version | cut -d'-' -f 1)"
+  PULL_REQUESTS=$(curl -s https://api.github.com/repos/redhat-developer/rh-che/pulls?state=open | jq '.[].title')
+  PR_EXISTS="false"
+  while read -r pr_title
+  do
+    if [[ "$pr_title" == "\"$RELATED_PR_TITLE\"" ]]; then
+      PR_EXISTS="true"
+      break
+    fi
+  done <<< "$PULL_REQUESTS"
+
+  if [[ "$PR_EXISTS" == "false" ]]; then
+    echo "false"
+  else
+    echo "true"
+  fi
+}
+
+function createPullRequest() {
+  PR_BODY="Tracking changes for fixing compatibility with upstream $CHE_VERSION. This PR was created automatically by Jenkins from job $JOB_NAME"
+  PR_HEAD="$BRANCH"
+  PR_BASE="master"
+  RELATED_PR_TITLE="Update to $(echo $CHE_VERSION | cut -d'-' -f 1)"
+  echo "Pull request for tracking changes of version $CHE_VERSION was not found - creating new one."
+  git push origin "$BRANCH" -f
+  curl -X POST -s -L -H "Content-Type: application/json" \
+       -H "Authorization: token $(echo ${FABRIC8_HUB_TOKEN}|base64 --decode)" \
+       --data "{\"title\":\"$RELATED_PR_TITLE\", \"head\":\"$PR_HEAD\", \"base\":\"$PR_BASE\", \"body\":\"$PR_BODY\"}" \
+       https://api.github.com/repos/redhat-developer/rh-che/pulls
+  if [[ "$?" != "0" ]]; then
+    echo $? > compatibility_test
+  fi
+}
+
+function runCompatibilityTest() {
 export USE_CHE_LATEST_SNAPSHOT="true"
 export BASEDIR=$(pwd)
 export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
 CHE_VERSION=$(curl -s https://raw.githubusercontent.com/eclipse/che/master/pom.xml | grep "^    <version>.*</version>$" | awk -F'[><]' '{print $3}')
 if [[ -z $CHE_VERSION ]]; then
-	echo "FAILED to get che version. Finishing script."
-	exit 1
+  echo "FAILED to get che version. Finishing script."
+  echo "1" > compatibility_test
+  exit 1
 fi
 
 echo "********** Prepare environment for running compatibility test with upstream version of che: $CHE_VERSION **********"
@@ -27,7 +68,7 @@ eval "$(./env-toolkit load -f jenkins-env.json -r \
         ^ghprbPullId$ \
         ^RH_CHE \
         ^FABRIC8_HUB_TOKEN)"
-        
+
 source ./config
 source ./.ci/functional_tests_utils.sh
 
@@ -40,19 +81,41 @@ installDependencies
 BRANCH="upstream-check-$CHE_VERSION"
 echo "Checking if branch $BRANCH exists."
 set +e
-git checkout "$BRANCH"
+git checkout "$BRANCH" 2>&1 >/dev/null
 return_code=$?
 set -e
-if [ $return_code -eq 0 ]; then 
-  echo "Branch $BRANCH found - rebasing."
-  git rebase origin/master "$BRANCH"
-else 
+if [ $return_code -eq 0 ]; then
+  #If branch exists
+  if [[ "$(checkPullRequest $CHE_VERSION $JOB_NAME $BRANCH)" == "false" ]]; then
+    #If PR doesn't exist
+    createPullRequest
+  fi
+else
   echo "Branch $BRANCH not found - creating new one."
+  git checkout master
   git checkout -b "$BRANCH"
-	
-  #change version of used che
+  git config -l
+  echo "Setting new branch origin"
+  git push -v --set-upstream origin $BRANCH
   echo ">>> change upstream version to: $CHE_VERSION"
   scl enable rh-maven33 "mvn versions:update-parent  versions:commit -DallowSnapshots=true -DparentVersion=[${CHE_VERSION}] -U"
+fi
+
+git rebase origin/master "$BRANCH"
+if [[ "$?" != "0" ]]; then
+    echo $? > compatibility_test
+fi
+
+if ( git diff --exit-code ); then
+  echo "Nothing to commit, continue."
+else
+  echo "Changes found. Commit and push them before creating PR."
+  git add -u
+  git commit -m "Changing version of parent che to $CHE_VERSION" || echo "No changes found to commit."
+  git push origin "$BRANCH" -f
+  if [[ "$?" != "0" ]]; then
+    echo $? > compatibility_test
+  fi
 fi
 
 echo "Setting image tags for pushing to quay."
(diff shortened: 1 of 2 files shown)


.ci/cico_rhche_compatibility_test.sh
```diff
@@ -36,25 +79,53 @@ checkAllCreds
...
-  git rebase origin/master "$BRANCH"
-else 
+if [ ! $return_code -eq 0 ]; then
   echo "Branch $BRANCH not found - creating new one."
+  git checkout master
   git checkout -b "$BRANCH"
-	
-  #change version of used che
+  echo "Setting new branch origin"
+  git push --set-upstream origin $BRANCH
```
Katka92: Can you explain the meaning of this line - what it was before and why to change that?

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -36,25 +79,53 @@ checkAllCreds
...
 
+if ( git diff --exit-code ); then
+  echo "Nothing to commit, continue."
+else
+  echo "Changes found. Commit and push them before creating PR."
+  git add -u
+  git commit -m "Changing version of parent che to $CHE_VERSION" || echo "No changes found to commit."
+  git push origin "$BRANCH" -f
+  if [[ "$?" != "0" ]]; then
+    echo $? > compatibility_test
```
Katka92: Do we want to continue when branch is not pushed?

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -36,25 +79,53 @@ checkAllCreds
...
+  fi
+fi
+
+git rebase origin/master "$BRANCH"
+if [[ "$?" != "0" ]]; then
+  echo $? > compatibility_test
+else
+  if ( ! git diff --exit-code ); then
```
Katka92: I think this `if` is not needed here, we can just check if PR exists.

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -127,7 +157,34 @@ if [ $RETURN_CODE != 0 ]; then
...
+
+run_tests_timeout_seconds=${RUN_TEST_TIMEOUT:-900}
+SECONDS=0
+touch compatibility_status
+echo "Starting compatibility test"
+runCompatibilityTest &
+COMPATIBILITY_PID=$!
+COMPATIBILITY_STATUS=$(cat compatibility_status)
+echo "Compatibility test running in background thread $COMPATIBILITY_PID $COMPATIBILITY_STATUS"
+while true; do
```
Katka92: 1) Why is this sort of timeout better then timeout set for a job?
2) It seems that SECONDS are not incremented.

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -36,25 +79,53 @@ checkAllCreds
...
 
+echo "Configuring remote:"
+git remote set-url origin https://${GITHUB_AUTH_STRING}@github.com/redhat-developer/rh-che.git
+
 #Check branch for tracking changes/Create new branch
 BRANCH="upstream-check-$CHE_VERSION"
 echo "Checking if branch $BRANCH exists."
 set +e
-git checkout "$BRANCH"
+git checkout "$BRANCH" 2>&1 >/dev/null
```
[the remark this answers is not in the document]
ScrewTSW: so that we don't pollute logs with unnecessary messages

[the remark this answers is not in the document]
ScrewTSW: we manually check status after that

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,15 +5,55 @@
...
+  if [[ "$PR_EXISTS" == "false" ]]; then
+    echo "false"
+  else
+    echo "true"
+  fi
+}
+
+function createPullRequest() {
+  PR_BODY="Tracking changes for fixing compatibility with upstream $CHE_VERSION. This PR was created automatically by Jenkins from job $JOB_NAME"
+  PR_HEAD="$BRANCH"
```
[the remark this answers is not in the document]
ScrewTSW: fixed

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,15 +5,55 @@
...
+  PR_BASE="master"
+  RELATED_PR_TITLE="Update to $(echo $CHE_VERSION | cut -d'-' -f 1)"
```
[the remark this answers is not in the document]
ScrewTSW: cleaned up

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,15 +5,55 @@
...
+  echo "Pull request for tracking changes of version $CHE_VERSION was not found - creating new one."
+  git push origin "$BRANCH" -f
+  curl -X POST -s -L -H "Content-Type: application/json" -u ${GITHUB_AUTH_STRING} \
+       --data "{\"title\":\"$RELATED_PR_TITLE\", \"head\":\"$PR_HEAD\", \"base\":\"$PR_BASE\", \"body\":\"$PR_BODY\"}" \
+       https://api.github.com/repos/redhat-developer/rh-che/pulls
+  if [[ "$?" != "0" ]]; then
+    echo $? > compatibility_test
```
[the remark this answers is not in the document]
ScrewTSW: if the `compatibility_test` file contains any text, it automatically stops the script execution as the timer checks the file periodically every second for a status code.
If it contains anything it attempts to kill the background process which actually runs the compatibility test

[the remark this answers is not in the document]
ScrewTSW: I can still add `exit $return_code` just to make sure the script doesn't continue

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -36,25 +79,53 @@ checkAllCreds
...
   echo ">>> change upstream version to: $CHE_VERSION"
   scl enable rh-maven33 "mvn versions:update-parent  versions:commit -DallowSnapshots=true -DparentVersion=[${CHE_VERSION}] -U"
 fi
 
+if ( git diff --exit-code ); then
+  echo "Nothing to commit, continue."
+else
+  echo "Changes found. Commit and push them before creating PR."
+  git add -u
+  git commit -m "Changing version of parent che to $CHE_VERSION" || echo "No changes found to commit."
```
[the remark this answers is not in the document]
ScrewTSW: cleaned up

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: At that point, the branch doesn't have an upstream tracking branch. This command creates it in origin, so that the PR can be created

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: same as https://github.com/redhat-developer/rh-che/pull/1647#discussion_r332447246

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: cleaned up

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: this is a quality of life improvement. before, when the job got stuck on anything, it was running there until it timed out on jenkins and I believe the timeout was 1 hour.

ScrewTSW: the seconds are indeed incremented, you can see it in the logs as `Tick Tock [time]` message

.ci/cico_rhche_compatibility_test.sh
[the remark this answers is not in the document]
Katka92: I think that this is not clear for somebody who is not familiar with the logic. And one second is still a lot of time when the code can continue, so I would prefer having the exit here (and on all other places where the logic is the same ;) ).

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: it is an internal bash variable that's automatically being incremented during a script run

.ci/cico_rhche_compatibility_test.sh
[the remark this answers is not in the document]
ScrewTSW: fixed

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,181 @@
...
+  return_code=$?
+  if [ ! $return_code -eq 0 ]; then
+    echo $return_code > compatibility_test
+    exit $return_code
+  fi
+}
+
+function runCompatibilityTest() {
+  export USE_CHE_LATEST_SNAPSHOT="true"
+  export BASEDIR=$(pwd)
```
Katka92: Both `USE_CHE_LATEST_SNAPSHOT` and `BASEDIR` seems to be not used.

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,181 @@
...
+  export DEV_CLUSTER_URL=https://devtools-dev.ext.devshift.net:8443/
+  export CHE_VERSION=$(curl -s https://raw.githubusercontent.com/eclipse/che/master/pom.xml | grep "^    <version>.*</version>$" | awk -F'[><]' '{print $3}')
```
amisevsk: Another option here, if `yq` is installed in the CI system, is to use `xq`, which I believe comes with `yq`:

```
curl -s [url] | xq -r '.project.version'
```

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,181 @@
...
+  if [[ -z $CHE_VERSION ]]; then
+    echo "FAILED to get che version. Finishing script."
+    echo "1" > compatibility_test
+    exit 1
+  fi
+  export RELATED_PR_TITLE="Update to $(echo $CHE_VERSION | cut -d'-' -f 1)"
```
amisevsk: You can also use `${CHE_VERSION%-SNAPSHOT}` to get Che version without the `-SNAPSHOT`

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,181 @@
...
+  echo "Configuring remote:"
+  git remote set-url origin https://${GITHUB_AUTH_STRING}@github.com/redhat-developer/rh-che.git
+
+  #Check branch for tracking changes/Create new branch
+  echo "Checking if branch $BRANCH exists."
+  set +e
+  git checkout "$BRANCH" 2>&1 >/dev/null
+  return_code=$?
+  set -e
+  if [ ! $return_code -eq 0 ]; then
```
amisevsk: What about `if ! git checkout $BRANCH 2>&1 >/dev/null; then`

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,181 @@
...
+    if [[ "$pr_title" == "\"$RELATED_PR_TITLE\"" ]]; then
+      PR_EXISTS="true"
+      break
+    fi
+  done <<< "$PULL_REQUESTS"
+
+  if [[ "$PR_EXISTS" == "false" ]]; then
+    echo "false"
+  else
+    echo "true"
```
amisevsk: This function could be reworked to return an exit code:
```bash
if [ "$PR_EXISTS" == "true"; then
  return 0
else
  return 1
fi
```
It can then be used directly in a condition check:
```bash
if ! checkPullRequest; then
  # PR doesn't exist
fi
```

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: git can return varying return codes based on the error that happened. I would like to have those error codes printed as the exit code of the script.

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: Good catch :) didn't know about this capability of bash.

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: This is also news to me :D but thank you for the suggestion, will try it out

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: good catch :) I will do a complete refactor of the script at some later time anyways.

.ci/cico_rhche_compatibility_test.sh
ScrewTSW: BASEDIR is being set in the cico_rhche_prcheck.sh, but it's not used in the build scripts. Removed

ScrewTSW:
[test]

.ci/cico_rhche_compatibility_test.sh
```diff
@@ -5,129 +5,184 @@
...
+    url="${url}/comments"
+    job_url="https://ci.centos.org/job/devtools-rh-che-rh-che-compatibility-test-dev.rdu2c.fabric8.io/${BUILD_NUMBER}/console"
+    message="Periodic compatibility check failed. See more details here: ${job_url}"
+    curl -X POST -s -L -u "${GITHUB_AUTH_STRING}" "${url}" -d "{\"body\": \"${message}\"}"
+  fi
+
+  echo $return_code > compatibility_status
+}
+
+<<<<<<< HEAD
```
Katka92: Merge conflict, needs to be fixed...

ScrewTSW:
[test]

Katka92:
[test]

PR #1654, opened 2019-10-10
Reducde PR scripts unnecessary logging

ScrewTSW:
### What does this PR do?
Reducing debug levels for yum dependency install and maven artifact downloads

none

manual testing with prcheck

### DIFF
--- a/.ci/cico_do_build_che.sh
+++ b/.ci/cico_do_build_che.sh
@@ -20,12 +20,12 @@ mvnche() {
     then
       # gulp-cli is needed to build the dashboard
       scl enable rh-nodejs8 "npm install --global gulp-cli"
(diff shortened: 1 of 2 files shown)


ScrewTSW:
> Would it also be possible to reduce the logging for docker pulls/pushes? it looks like it's doing full progress bars still.

I was indeed looking into that, but the options that I found were very unsatisfactory.

`less` works beautifully until the text gets bigger than one page
`tail -f` has no output for docker until the command exits.

If you have any idea that would be awesome. I'm merging this PR and I can open another when we find a solution.

amisevsk:
@ScrewTSW The general tip I've seen elsewhere is to pipe to `cat`, so that docker doesn't think it's outputting to a tty, maybe that would work?

PR #1662, opened 2019-10-18
Reduce logging from rsync when transferring images in CI

amisevsk:
### What does this PR do?
Removes progress bars from rsync step in CI builds, since some failures can fill 1/3rd of the log with progress that's formatted for a tty.


amisevsk:
The two commits here roughly half log size from successful builds:

- Successful build w/o PR (\#1900): 7707 lines, 360KB
- Successful build with PR (\#1910): 3603 lines, 252KB

PR #1663, opened 2019-10-22
Fixing versioning.

Katka92:
### What does this PR do?
Fix logic of versioning. The test fails if docker image with a current tag does not exist.


ibuziuk:
[test]

Issue #1589, opened 2019-09-03
Replace workspace config with devfile in workspace-startup tests


ScrewTSW:

ibuziuk:
@ScrewTSW ca we close this one?

ibuziuk:
ok, I see we need to merge - https://github.com/redhat-developer/rh-che/pull/1679

ScrewTSW:
> ok, I see we need to merge - #1679

yes.

ibuziuk:
https://github.com/redhat-developer/rh-che/pull/1679 merged. Closing

PR #1679, opened 2019-11-04
Switch workspace startup test to devfiles

ScrewTSW:
### What does this PR do?
This devfile deprecates our outdated workspace definition jsons and replaces them with devfiles from our registries

#1589 

manual testing and deployment on codeready jenkins


.ci/workspace-startup/workspace-startup-test-ephemeral.groovy
```diff
@@ -40,9 +40,10 @@ pipeline {
                 }
                 dir ("${LOG_DIR}") {
                     silent_sh """
+                    wget https://che-devfile-registry.prod-preview.openshift.io/devfiles/nodejs/devfile.yaml -O ../${RELATIVE_PATH}/nodejs_eph.yaml
```
ibuziuk: @ScrewTSW why do you use prod-preview specific devfile?

ScrewTSW: because the tests are running against prod-preview. so it makes sense to use it's devfile registry

ScrewTSW:
[test]

.ci/workspace-startup/workspace-startup-test-ephemeral.groovy
ibuziuk: can we use getting-started devfile instead e.g.  https://github.com/redhat-developer/devfile/tree/master/getting-started/nodejs - https://www.eclipse.org/che/getting-started/cloud/

This one must work for both prod & prod-preview

ScrewTSW: It was recommended against by @rhopp and I understand where he's coming from. This way we are absolutely sure that the devfiles we use should be working with the version of Che they are running against and this process is automatic, no manual steps required.

ibuziuk: not following, we have absolutely no tests for getting-started devfiles, why not to have at least startup tests running against them?

ScrewTSW: @rhopp ^ what do you think?

Katka92: @ScrewTSW OK, so let's keep stacks from prod and prod-preview here. But please create an issue about testing stacks on getting started page so we will not forget that. @ibuziuk Are you OK with that?

ibuziuk: > So basically you are saying that devfiles from getting-started should work both on production and prod-preview so there should be no problem using them. And as a value-added, we will have at least one of them tested. Am I right? 

correct

> I wouldn't necessarily mix this together. Workspace startup tests are supposed to test "known good" stacks and collect data about startup times.

wdym, "known good"? if getting-started devfiles do not work we are screwed. Those devfiles  
even more important than UD devfiles since it is first user experience

ScrewTSW: @ibuziuk I still think that we should have a separate test suite and jenkins jobs for the getting started ones.
By "known good" I meant that the devfiles we have in registries should be tested by upstream and provided in a working form.

Katka92:
[test]

Katka92:
Tests are passing. Can we merge that? cc @ibuziuk @ScrewTSW

Issue #1664, opened 2019-10-22
Satisfy Requirements for AppSRE applications - define requests and limits, for both cpu and memory.


ibuziuk:
As part of our effort to standardize all the applications being run by the
AppSRE team, we want to enforce the following requirements:

- All containers images must come from quay.io
- All containers must define requests and limits, for both cpu and memory.
  Requests must be strictly smaller than limits.

We have identified a few containers that don't satisfy the above requirements:

http://file.rdu.redhat.com/jmelisba/saas-repos-logs-latest/saas-openshiftio.log
http://file.rdu.redhat.com/jmelisba/report-201910211441.html

For Hosted Che team we need to tackle:
- [x] che-plugin-registry.yaml:
   - [x] https://github.com/eclipse/che-plugin-registry/pull/292
   - [x] port to 7.3.x - https://github.com/eclipse/che-plugin-registry/pull/293
ERROR: CheckBestEffort:check_best_effort[kind='DeploymentConfig', name='che-plugin-registry']: Expecting requests/cpu in container 'che-plugin-registry'.
- [x] che-tenant-maintainer.yaml:
ERROR: CheckBestEffort:check_best_effort[kind='DeploymentConfig', name='che-tenant-maintainer']: Expecting requests/cpu in container 'che-tenant-maintainer'.
   - [x] https://github.com/fabric8-services/che-tenant-maintainer/pull/29
- [x] kubernetes-image-puller.yaml:
ERROR: CheckBestEffort:check_best_effort[kind='DeploymentConfig', name='kubernetes-image-puller']: Expecting requests/cpu in container 'kubernetes-image-puller'.
   - [x] https://github.com/redhat-developer/kubernetes-image-puller/pull/17
- [x] rh-che6.yaml:
ERROR: CheckBestEffort:check_best_effort[kind='DeploymentConfig', name='rhche']: Expecting requests/cpu in container 'rhche'.
  - [x] https://github.com/redhat-developer/rh-che/pull/1688
- [x] che-devfile-registry.yaml:
ERROR: CheckBestEffort:check_best_effort[kind='DeploymentConfig', name='che-devfile-registry']: Expecting requests/cpu in container 'che-devfile-registry'.
  - [x] https://github.com/eclipse/che-devfile-registry/pull/141
  - [x] port to `7.3.x` - https://github.com/eclipse/che-devfile-registry/pull/142
- [x] che-monitoring.yaml:
[...]

PR #1688, opened 2019-11-08
rhche #1664 Adding cpu requests / limits to the openshift template

ibuziuk:
### What does this PR do?
Adding cpu requests / limits to the openshift template

 #1664

PR check CI

### DIFF
--- a/openshift/rh-che.app.yaml
+++ b/openshift/rh-che.app.yaml
@@ -149,8 +149,10 @@ objects:
             timeoutSeconds: 60
           resources:
             limits:
+              cpu: 1000m
               memory: 4Gi
             requests:
+              cpu: 35m
               memory: 512Mi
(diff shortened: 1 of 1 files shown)


ibuziuk:
[test]

openshift/rh-che.app.yaml
```diff
@@ -149,8 +149,10 @@ objects:
             timeoutSeconds: 60
           resources:
             limits:
+              cpu: 1000m
```
amisevsk: Do we have data on whether this limit will be sufficient? `1000m` seems low to me as a maximum.

ibuziuk: Should be more than enough for the last week the usage was around 0.05  - also see http://file.rdu.redhat.com/jmelisba/report-201910211441.html

Issue #1042, opened 2018-11-03
It should be possible to start Che workspace when namespace of type user is missing


garagatyi:
**Issue problem:** 
On workspace provisioning step we get information about namepsace of type `user`. But if such namespace is missing user can't start any workspace.

ibuziuk:
@alexeykazakov this is the second iteration of https://github.com/redhat-developer/rh-che/issues/1296 
should we consider taking it to the next sprint?

alexeykazakov:
@ibuziuk IMO yes. But @gorkem can confirm if this is what we want (to be ready to drop usernamespaces the next sprint).

ibuziuk:
ok waiting for confirmation before adding to the pri doc for the next sprint

ibuziuk:
@alexeykazakov we are NOT planning to take this issue to the next sprint. If the issue should be tackled during the next sprint please comment

amisevsk:
FWIW it looks like a relatively simple issue to fix, so if it's pressing I wouldn't be against spending half a day working on it next sprint -- this wouldn't interfere with planning much. WDYT @ibuziuk ?

If it turns out to be more complicated than it looks, we can drop it from the sprint but if it can be fixed in a few hours I think we could include it.

amisevsk:
From what I can tell, what we're doing now is getting namespace, cluster URL, and certs from the user namespace and injecting them into the workspace (presumably so that `oc` commands in the workspace use the `user` namespace. Simply replacing the `user` namespace with `-che` would only change where `oc` commands are run. The caveat here is that whatever we do with `oc` would then have to fit in the `-che` namespace alongside the workspace, which could be a problem.

ibuziuk:
@amisevsk no, I suppose we should not work on until user workspace would be confirmed to be removed, once it is done we would simply do a cleanup including removal of `login to OSO` installer. At the moment we do not plan to add this issue to the next sprint, so no work is needed in this deirection cc: @alexeykazakov @gorkem

ibuziuk:
Going to close after the next production update

ibuziuk:
The fix is on prod. Closing

PR #1732, opened 2019-12-11
Allow Che workspace statup when namespace of type user is missing

ibuziuk:
### What does this PR do?
- Removing obsolete 'login to oc' Che 6 installer
- Return blank tenant data instead of throwing an exception if namespace of type 'user' is missing (this might happen only if there is a bug on init-tenant or 'api/user/services' side)

https://github.com/redhat-developer/rh-che/issues/1042

Unit tests / not tested in real conditions when the 'user' namespace is really absent (potentially could be tested on preview with a request of manual namespace removal)


ibuziuk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

ibuziuk:
[test]

Issue #1680, opened 2019-11-05
Improve workspace start/stop reports


Katka92:
**Description:**
The **Che QE bot** is nicely reporting the workspace start/stop times [1] every day. Due to logic for labeling tests are green/red we have the whole report marked as red almost every day even if there is no problem wit workspace start/stop times.

**Current state: **
The threshold for the test to be green/red is set by percentage (currently set to 5%). This is quite misleading nowadays because times are quite stable so even the small change is counted as a high percentage. That's why we are having all reports red even there is no problem with workspace start/stops. 

**Possible solutions:**
1) Remove color labels so everyone can watch exact times and make own picture about how good/bad the start/stop times are.
2) Adjusting the algorithm for labeling gathered times as red/green. Maybe that would be suitable for us to set seconds as a threshold, not a percentage. 

[1] 
![Screenshot from 2019-11-05 13-46-37](https://user-images.githubusercontent.com/29278800/68208991-b9630480-ffd2-11e9-9c85-f49ec4153966.png)

amisevsk:
One approach that may make more sense now that the timing has settled down is to just specify explicit cutoffs. E.g. for start, avg <35 seconds is green, 35-45 seconds is yellow, >45 seconds is red. These could be adjusted as we bring start times down (to reflect what we expect our performance to be).

amisevsk:
One approach that may make more sense now that the timing has settled down is to just specify explicit cutoffs. E.g. for start, avg <35 seconds is green, 35-45 seconds is yellow, >45 seconds is red. These could be adjusted as we bring start times down (to reflect what we expect our performance to be).

amisevsk:
One approach that may make more sense now that the timing has settled down is to just specify explicit cutoffs. E.g. for start, avg <35 seconds is green, 35-45 seconds is yellow, >45 seconds is red. These could be adjusted as we bring start times down (to reflect what we expect our performance to be).

ScrewTSW:
![Screenshot_20200109_094004](https://user-images.githubusercontent.com/16451875/72054114-aa494d00-32c8-11ea-931a-7362ede533e5.png)

Current updated layout.
Working on cleaning up the code and then I'll submit a PR

Issue #1660, opened 2019-10-16
Verify workspace-start-tests are working correctly


Katka92:
**Description:**
It seems that data gathered from `workspace-start-tests` are not consistent. The consistency problem seems to be not only between Zabbix and Grafana, but even between different Zabbix charts.

1. It seems that data in Zabbix differs a lot from data in Grafana
Zabbix (e.g. cluster 3):
![Screenshot from 2019-10-16 15-17-40](https://user-images.githubusercontent.com/29278800/66922838-257fc780-f028-11e9-807e-9ac80b35f228.png)

Grafana:
![Screenshot from 2019-10-16 15-18-24](https://user-images.githubusercontent.com/29278800/66922897-39c3c480-f028-11e9-8550-2e0a74469c9a.png)

2. The soft-fail seems to be set to 5 sec but the data in the chart are not corresponding.
![Screenshot from 2019-10-16 15-16-05](https://user-images.githubusercontent.com/29278800/66922746-f406fc00-f027-11e9-9b78-680eb7c29604.png)

ibuziuk:
@Katka92 should we close it as outdated? I believe we need to re-evaluate the tests if / when we would have a monitoring stack on osd

Katka92:
Closing as outdated.

PR #1751, opened 2020-01-09
Fixed the attachment colors, they work based on AVG startup times now

ScrewTSW:
### What does this PR do?
This PR changes the logic for the workspace-startup-reporter regarding how are the thresholds calculated, switching over from the percent change over to fixed startup time values

https://github.com/redhat-developer/rh-che/issues/1680

Manual testing, sending into private slack channel.


Katka92:
I approved as there will be the next step that will probably require rewriting some logic and refactoring. We'll start to work on this probably next week. https://github.com/redhat-developer/rh-che/issues/1660

Issue #1604, opened 2019-09-09
[Che issue] Timeout when waiting for IDE


Katka92:
Workspace is created correctly, but it last too long for IDE to show up. Current timeout is set to 2 minutes from clicking button ` Create & Open`. 

![image.png](https://images.zenhubusercontent.com/59a50d048f62dc7798c4656a/52a9dd99-20dd-4805-8702-b113bde8a8ae)

Last seen: https://ci.centos.org/job/devtools-rh-che-periodic-prod-preview-2a/1356/console

Katka92:
Seen again on prod-preview 
https://ci.centos.org/job/devtools-rh-che-periodic-prod-preview-2a/1366/console 
Screenshot:
![image](https://user-images.githubusercontent.com/29278800/64673703-28a6f700-d46f-11e9-9e7b-e8be987b3686.png)

Katka92:
Seen again on 1a and 1b.

ScrewTSW:
seen today at 1b cluster x1

ScrewTSW:
Seen today again on prod `1a` two times. and once on `1b`

ScrewTSW:
Appeared today only once at `1b` cluster.

ScrewTSW:
Seen today once on cluster `1b`

Katka92:
Happened again 4x on `1a` and 2x on `1b`.

Katka92:
Happened again on `1a` 1x.

Katka92:
Happened again on `1a` 1x.

Katka92:
Seen on `1a` 2x.

ScrewTSW:
Seen on `1a` twice and once on `prod-preview 2a` stuck at IDE load screen
![image](https://user-images.githubusercontent.com/16451875/65878474-71770f00-e38e-11e9-8f53-ee29b18aba6f.png)

ScrewTSW:
happened once on `1a`

ScrewTSW:
happened once on `1a` and twice on `2a`

Katka92:
Happened again on 1a once.

Katka92:
Seen 2x on 2, once on 2a, once 1a, once on 1b.

Katka92:
Happened 1x on 1a and 1x on 1b.

ScrewTSW:
Happened 1x on `1a`

Katka92:
[...]

PR #1754, opened 2020-01-10
Increase timeout to workaround issue with IDE not loading in given time.

Katka92:
### What does this PR do?
Increase timeout for loading a page - it should workaround issue with IDE not loading in a given time.
The timeout was set by default to 120000 and is increased to 180000.

#1604

### DIFF
--- a/functional-tests/devscripts/run_tests.sh
+++ b/functional-tests/devscripts/run_tests.sh
@@ -124,6 +124,7 @@ if [[ "$PR_CHECK_BUILD" == "true" ]]; then
     fi
   fi
   
+  #increase timeout for load page to workaround https://github.com/redhat-developer/rh-che/issues/1604
   docker run \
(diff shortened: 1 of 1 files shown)


ibuziuk:
I suspect that this will not help, according to the screenshots the workspace is running, just for some reason, the IDE is not opened, So, my bet is that this isn upstream issue and timeout increase will not help

Katka92:
@ibuziuk This issue happened e.g. here: https://ci.centos.org/job/devtools-rh-che-periodic-prod-preview-2a/2086/console and in a log, you can see that even the IDE was not loaded at the given timeout, it loaded a while after that, so the rest of a test resumed correctly. The test which is waiting for IDE is not checking a worksapce status but is waiting for specific IDE page objects, like project explorer or top menu panel. So that's why I think changing the timeout can help us.

PR #1761, opened 2020-01-15
Fixing typo, enhancing logic.

Katka92:
### What does this PR do?
Fix a typo which causes the script not exiting after a failure.
Change logic:
was
 - branch can not be rebased
 - fail

changed to
 - branch can not be rebased
 - check PR exists (create if does not)
 - comment to the PR about rebasing failure
 - fail

I'm afraid that I can not test this scenario easily. I can verify that there are no failures in the script, but the logic will be tested once it will happen again I think.


Katka92:
[test]

PR #1767, opened 2020-01-21
Fix 7.7.1b test images

Katka92:
### What does this PR do?
Hotfix for 7.7.1b. Those images were not pushed to the upstream test repo, so the downstream images are failing.

PR #1768, opened 2020-01-21
Check upstream images on quay instead of docker.

Katka92:
### What does this PR do?
Upstream test images moved to quay. Changing logic in our tests to check whether an upstream image exists on quay.io, not docker.io.


Katka92:
Closing, need to be solved as part of https://github.com/redhat-developer/rh-che/pull/1767.

Issue #1771, opened 2020-01-21
After 7.7.1 update PVC based workspace are failed for users with usernames != <namepsace> name


ibuziuk:
**Issue problem:** 
After 7.7.1 update PVC based workspace are failed for users with usernames != <namepsace> name. Ephemeral (emptyDir volume namespace work just fine)

version: (help/about menu)

 - [ ] Hosted Che 7.7.1 / 7.7.1b

**Reproduction Steps:**

- Have an account where username = email
- Start default ephemeral emptyDir workspace
- SUCCESS: workspace startup works just fine
- Stop the workspace and turn off the ephemeral toggle
- ERROR: workspace is failed to start with the following message:

> Error: Failed to run the workspace: "Failure executing: GET at: https://f8osoproxy-test-dsaas-production.09b5.dsaas.openshiftapps.com/api/v1/namespaces/ibuziuk@redhat.com-che/pods/mkdir-workspacexjhg28wbmpxn4pdj. Message: Unauthorized! Configured service account doesn't have access. Service account may have been revoked.."

**Runtime:**

runtime used:
   - [ ] minishift (include output of `minishift version`)
   - [x] OpenShift.io
   - [ ] Openshift Container Platform (include output of `oc version`)

```
logs:

org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException: Failure executing: GET at: https://f8osoproxy-test-dsaas-production.09b5.dsaas.openshiftapps.com/api/v1/namespaces/blue.granite-che/pods/mkdir-workspace1pjk8i5p0pcz8brk. Message: Unauthorized! Configured service account doesn't have access. Service account may have been revoked..
	at org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesInternalRuntime.wrapAndRethrow(KubernetesInternalRuntime.java:815)
	at org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesInternalRuntime.internalStart(KubernetesInternalRuntime.java:274)
	at org.eclipse.che.api.workspace.server.spi.InternalRuntime.start(InternalRuntime.java:141)
	at org.eclipse.che.core.tracing.TracingInterceptor.invoke(TracingInterceptor.java:61)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes$StartRuntimeTask.run(WorkspaceRuntimes.java:873)
[...]

PR #1772, opened 2020-01-21
rh-che #1771 Overriding methods for namespace evaluation on 'Fabric8OpenShiftProjectFactory' level

ibuziuk:
### What does this PR do?
Overriding methods for namespace evaluation on 'Fabric8OpenShiftProjectFactory' level

https://github.com/redhat-developer/rh-che/issues/1771

N/A


ibuziuk:
cc @meltos

ibuziuk:
@amisevsk thanks for review
@meltos can you comment on that and provide some guidance what is expected to be overridden on our end?

metlos:
The override in the legacy case is needed for handling the legacy workspaces that do not have their namespace stored in their attributes. In that case, we need a fallback method that should figure out where such workspaces are/should be running. In case of rh-che, this has always been the same.

PR #1786, opened 2020-02-03
Overriding 'evaluatePlaceholders' to prevent workspace startup failures for users whose login != <namespace>-che

ibuziuk:
# What does this PR do?
Overriding 'evaluatePlaceholders' to prevent workspace startup failures for users whose login != <namespace>-che

N/A 
issue identified during tests on prod-preview
workspace

Can be tested on prod-preview / dev-clsuter

### DIFF
--- a/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
+++ b/plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
@@ -24,9 +24,12 @@ import org.eclipse.che.workspace.infrastructure.openshift.OpenShiftClientConfigF
 import org.eclipse.che.workspace.infrastructure.openshift.OpenShiftClientFactory;
 import org.eclipse.che.workspace.infrastructure.openshift.project.OpenShiftProject;
 import org.eclipse.che.workspace.infrastructure.openshift.project.OpenShiftProjectFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** @author Sergii Leshchenko */
 public class Fabric8OpenShiftProjectFactory extends OpenShiftProjectFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(Fabric8OpenShiftProjectFactory.class);
 
   private final OpenShiftClientFactory clientFactory;
   private final Fabric8WorkspaceEnvironmentProvider envProvider;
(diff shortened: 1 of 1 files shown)


plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
```diff
@@ -69,6 +72,26 @@ public String evaluateLegacyNamespaceName(NamespaceResolutionContext resolutionC
...
+    try {
+      placeholder = evaluateNamespace();
+    } catch (InfrastructureException e) {
+      LOG.error("Failed to evaluate placeholder for '*-che' namespace: {}", e.getMessage());
+    }
+    return placeholder;
+  }
+
   private String evaluateNamespace() throws InfrastructureException {
     Subject currentSubject = EnvironmentContext.getCurrent().getSubject();
```
sleshchenko: It makes sense to check is the current user is not ANONYMOUS, if so - please log error along with stacktrace.
It could help investigate the issue with `Service Account does not have permissions to clean up workspaces resources`

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
```diff
@@ -69,6 +72,26 @@ public String evaluateLegacyNamespaceName(NamespaceResolutionContext resolutionC
...
+  }
+
+  private String evalPlaceholders() {
+    String placeholder = null;
+    try {
+      placeholder = evaluateNamespace();
+    } catch (InfrastructureException e) {
+      LOG.error("Failed to evaluate placeholder for '*-che' namespace: {}", e.getMessage());
+    }
+    return placeholder;
```
sleshchenko: I'm not sure that returning null here won't cause NPE above in upstream code.

ibuziuk: technically this should never happen, I can return emty string though if you think this would be better

plugins/fabric8-multi-tenant-manager/src/main/java/com/redhat/che/multitenant/Fabric8OpenShiftProjectFactory.java
ibuziuk: I think it is done on the deeper level

```throw new InfrastructureException(
          "The anonymous subject is used, and won't be able to perform this action");```

ibuziuk:
@sleshchenko merging to master, but please comment if you think smth. extra should be done

PR #1795, opened 2020-02-13
Fix event obtaining in compatibility test

Katka92:
### What does this PR do?
Compatibility is still failing on weird issues. Verify that it was a networking problem and the script is working correctly by running this test instead of PR check here.


Katka92:
[test]

Katka92:
[test]

PR #1803, opened 2020-02-19
Triggering periodic tests every 3h instead of 1h to reduce the load on the clusters

ibuziuk:
### What does this PR do?
Triggering periodic tests every 3h instead of 1h to reduce the load on the clusters

N/A

N/A


ScrewTSW:
Did we see any significant effects of those jobs? We used to run them every 30 minutes for months before the migration to crw-jenkins.

ibuziuk:
yes, we have a rise of req metrics, that's is why I want to make it less frequent - 8 times per day is enough

Katka92:
I just want to ensure that we are all on the same page. So @ibuziuk you want to reduce the time of just `workspace-startup` tests (the ones that do daily report to the slack channel). And the periodic jobs (I mean typescript ones) and route tests are OK to be as they are now, right?
Just to be clear - typescript periodic tests are running every 4 hours, route tests once an hour, rollout test once a day and compatibility test once a day. All jobs expect route tests are creating workspaces.

ibuziuk:
> I just want to ensure that we are all on the same page. So @ibuziuk you want to reduce the time of just workspace-startup tests (the ones that do daily report to the slack channel).

correct

ScrewTSW:
@ibuziuk could we hold off this change until tomorrow? We are seeing some workspace startup times issues with che 7.9.0 happy path tests, we want to use the current granularity (once every hour) to observe if we will see same issues on prod-preview once the 7.9.0RC is deployed

ibuziuk:
@ScrewTSW not following, we are not going to promote 7.9.0 RC on prod-preview today / tomorrow

Issue #1775, opened 2020-01-27
Remove specific project for compatibility test


Katka92:
The compatibility test should use the same project as a PR when the compatibility check reports its results. 
Currently, the nightly run of compatibility check uses `compatibility-check` project on devcluster. We should get rid of it so there will be only one project for the one branch (we will also save resources by that).

Katka92:
Done.

PR #1805, opened 2020-02-21
Update to 7.10.0

RhCheGithubBot:
Tracking changes for fixing compatibility with upstream 7.10.0-SNAPSHOT. This PR was created automatically by Jenkins from job devtools-rh-che-rh-che-compatibility-test-dev.rdu2c.fabric8.io

PR #1806, opened 2020-02-21
Fix compatibility - remove test part

Katka92:
### What does this PR do?
The compatibility check is currently failing on the part that is running tests. In this PR, the logic for deploying Rh-Che and running tests is deleted. Instead, the test will comment `[test]` to the PR to trigger a PR check.

It solves the issue https://github.com/redhat-developer/rh-che/issues/1775


Katka92:
The script works as expected, it comments to the PR https://github.com/redhat-developer/rh-che/pull/1805 and tests were executed.

.ci/functional_tests_utils.sh
```diff
@@ -76,6 +76,18 @@ function installDependencies() {
   installNodejs
 }
 
+function installDependenciesForCompatibilityCheck() {
+  installEpelRelease
+  installYQ
+  installJQ
+  installGit  
+  # Getting dependencies ready
+  yum install --assumeyes -d1 \
```
ScrewTSW: Why do we need Java and Maven for the script if we comment `[test]` and exit?

Katka92: Because we are using that to change version here https://github.com/redhat-developer/rh-che/pull/1806/files#diff-946ba133a72dce6ae10a7d1b4f89575dR139

ScrewTSW: would it be possible to do it from bash using `sed` or `yq`?

ScrewTSW: Ok, let's keep Java and Maven there for now. Resolving discussion.

PR #1807, opened 2020-02-21
[DO NOT MERGE] Remove compabitility namespace

Katka92:
### What does this PR do?
Replacing PR check with the script to remove `compatibility-test` namespace as it will not be used anymore. It will not be cleaned as a part of cleanup job, because it does not start with `prcheck-`. 
This PR is not supposed to be merged.


Katka92:
Project `compatibility-test` was deleted from dev cluster successfully. Closing.

PR #1808, opened 2020-02-24
Prevent command execution echoing into console/logs, add logging for rollout tests

ScrewTSW:
### What does this PR do?
Disable command execution echoing into console

hotfix

manually

### DIFF
--- a/.ci/cico_build_func_dep.sh
+++ b/.ci/cico_build_func_dep.sh
@@ -6,6 +6,8 @@
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
 
+set +x
+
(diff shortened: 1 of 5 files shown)


Katka92:
I'm -1 for adding `set +x` to the scripts. I agree that the output would be simpler, but I'm afraid that it will make investigation of test failures in some cases more difficult/impossible.

ScrewTSW:
> 
> 
> I'm -1 for adding `set +x` to the scripts. I agree that the output would be simpler, but I'm afraid that it will make investigation of test failures in some cases more difficult/impossible.

I think it's better practice to introduce manual logging for those purposes. If we see some failures that are difficult to investigate, we can always add those logging lines into our scripts.

The main purpose for adding the `set +x` is to increase security and prevent us from leaking secrets. Besides that if I remember it correctly, it should be the default behavior anyways

PR #1809, opened 2020-02-24
Fix echoing username in tests.

Katka92:
### What does this PR do?
There was a typo in a variable, so the username of a test user was not printed correctly.


Katka92:
[test]

PR #1810, opened 2020-02-26
Added --nogpgcheck to yum install of zabbix-sender

ScrewTSW:
### What does this PR do?
Hopefully fixes route tests issue with installation of zabbix-sender

hotfix

manually

### DIFF
--- a/functional-tests/devscripts/route_tests.sh
+++ b/functional-tests/devscripts/route_tests.sh
@@ -46,7 +46,7 @@ function checkPrereq {
 		
 	if $SEND_TO_ZABBIX; then
 		rpm -ivh https://repo.zabbix.com/zabbix/3.0/rhel/7/x86_64/zabbix-release-3.0-1.el7.noarch.rpm
-		yum install -y zabbix-sender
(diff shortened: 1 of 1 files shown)


functional-tests/devscripts/route_tests.sh
```diff
@@ -46,7 +46,7 @@ function checkPrereq {
 		
 	if $SEND_TO_ZABBIX; then
 		rpm -ivh https://repo.zabbix.com/zabbix/3.0/rhel/7/x86_64/zabbix-release-3.0-1.el7.noarch.rpm
-		yum install -y zabbix-sender
+		yum install -y --nogpgcheck zabbix-sender
```
ibuziuk: could you clarify what is the fix is about in the description

ScrewTSW: There is an issue, where `zabbix-sender` is complaining that the package's GPG key is not installed. This broke our route tests. I'm adding the `--nogpgcheck` flag to YUM to avoid this problem. It's been broken with the recent versions of zabbix-sender.

PR #1794, opened 2020-02-12
Try to customize UD with product.json. Remove all not needed files.

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?

1. Add custom product.json file.
2. Remove all not needed file anymore
3. Clean up pom.xml

PR #1814, opened 2020-02-27
Patch dashboard

Katka92:
### What does this PR do?
There is a method `Dashboard.waitPage` which is used across the tests. It checked Dashboard readiness by checking items on the left menu. This menu can be configurable, and because `Organization` and `Factories` are going to be removed [1], this behavior will cause test failures.

[1] https://github.com/redhat-developer/rh-che/pull/1794

EDIT: The second commit is just a cleanup of unsed files.

Running locally from docker image against problematic PR.


ibuziuk:
PR with workaround on Hosted Che side - https://github.com/redhat-developer/rh-che/pull/1814 (should be removed once the issue is fixed in the upstream)

ibuziuk:
@Katka92 looks like still failing on error-highlighting

ibuziuk:
[test]

Issue #1798, opened 2020-02-14
Migrating Hosted Che e2e CI jobs from dev-cluster to OSD 4


ibuziuk:
dev-cluster is going to be deprecated in the short run. Need to migrate all the CI jobs to the OSD 4 which is going to be provisioned by the end of the month

ibuziuk:
blocked due to missing osd 4 cluster.

ibuziuk:
Removing the blocked label:
- cluster is provisioned
- redirect url for preview has been added
- SA has been created and added to vault

ibuziuk:
Closing as done

PR #1834, opened 2020-03-17
Migrating Hosted Che e2e CI jobs from dev-cluster to OSD 4

ibuziuk:
### What does this PR do?
Migrating Hosted Che e2e CI jobs from dev-cluster to OSD 4

https://github.com/redhat-developer/rh-che/issues/1798

test via CI jobs

### DIFF
--- a/.ci/cico_build_deploy_test_rhche.sh
+++ b/.ci/cico_build_deploy_test_rhche.sh
@@ -70,7 +70,7 @@ if ./dev-scripts/deploy_custom_rh-che.sh -o "${RH_CHE_AUTOMATION_DEV_CLUSTER_SA_
 
 then
   echo "Che successfully deployed."
-  export RH_CHE_AUTOMATION_SERVER_DEPLOYMENT_URL=https://rhche-$PROJECT_NAMESPACE.devtools-dev.ext.devshift.net/
+  export RH_CHE_AUTOMATION_SERVER_DEPLOYMENT_URL=https://rhche-$PROJECT_NAMESPACE.apps.che-dev.x6e0.p1.openshiftapps.com/
 else
   echo "Custom che deployment failed. Error code:$?"
   exit 4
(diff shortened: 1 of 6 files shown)


.ci/cico_build_deploy_test_rhche.sh
```diff
@@ -56,7 +56,7 @@ else
   echo "Switching to project failed, probably not exists [$?]. Creating..."
   oc new-project "${PROJECT_NAMESPACE}" --display-name="PR ${RH_PULL_REQUEST_ID} - Automated Deployment" > /dev/null 2>&1
 fi
-oc policy add-role-to-user edit Katka92 ScrewTSW rhopp tomgeorge ibuziuk amisevsk davidfestal skabashnyuk -n $PROJECT_NAMESPACE
+oc policy add-role-to-user edit ibuziuk -n $PROJECT_NAMESPACE
```
ScrewTSW: One thing I don't understand is why are we being removed from the adminlist?

.ci/cico_rollout_test.sh
```diff
@@ -114,7 +114,7 @@ oc login "${DEV_CLUSTER_URL}" --insecure-skip-tls-verify \
 echo "setting project"
 oc new-project "${PROJECT}" --display-name="PR ${RH_PULL_REQUEST_ID} - Automated Deployment" > /dev/null 2>&1
 echo "adding roles to project"
-oc policy add-role-to-user edit Katka92 ScrewTSW rhopp tomgeorge ibuziuk amisevsk davidfestal skabashnyuk -n $PROJECT
+oc policy add-role-to-user edit ibuziuk -n $PROJECT
```
ScrewTSW: Same question here.

Katka92:
It seems that token used for log in to the OC cluster is not set properly:
```
error: The token provided is invalid or expired.
```

.ci/cico_build_deploy_test_rhche.sh
ibuziuk: because I want to test this first with the minimal users, and othes would be added in a separate PR

ibuziuk:
@Katka92 I think it is expected since the token is from the old cluster atm

.ci/cico_build_deploy_test_rhche.sh
ScrewTSW: 👍

ibuziuk:
[test]

PR #1843, opened 2020-03-24
Fixing misconfigured mocha-pr-check.opts suite

ScrewTSW:
### What does this PR do?
Hotfix for rh-che prcheck test logic.
We were using an incorrect Reporter class which was causing issues with API calls because of different approach to getting ActiveToken of a user account.
This switches the suite to use our downstream implementation.

hotfix

manually against deployment on de-cluster

### DIFF
--- a/e2e-saas/mocha-pr-check.opts
+++ b/e2e-saas/mocha-pr-check.opts
@@ -1,5 +1,5 @@
 --timeout 120000
---reporter 'node_modules/e2e/dist/driver/CheReporter.js'
+--reporter './dist/utils/RhCheReporter.js'
 -u tdd
 --spec ./dist/tests/login/Login.spec
 --spec ./dist/tests/devfiles/JavaMaven.spec


e2e-saas/mocha-pr-check.opts
```diff
@@ -1,5 +1,5 @@
 --timeout 120000
---reporter 'node_modules/e2e/dist/driver/CheReporter.js'
```
ibuziuk: why the upstream reporter works for 7.9.1 ?

ScrewTSW: Because we weren't using any API requests in the older versions

ScrewTSW:
> I don't think we want to push this change to master, as the RhCheReporter is not present in 7.9.2. Please push those changes directly to 7.10.0 branch. Thank you

Good point. The change has been pushed there already.

PR #1845, opened 2020-03-25
Adding the original admins back to the prcheck script

ScrewTSW:
### What does this PR do?
Adds original admins back to the dev cluster edit rights list

hotfix

prcheck

### DIFF
--- a/.ci/cico_build_deploy_test_rhche.sh
+++ b/.ci/cico_build_deploy_test_rhche.sh
@@ -56,7 +56,7 @@ else
   echo "Switching to project failed, probably not exists [$?]. Creating..."
   oc new-project "${PROJECT_NAMESPACE}" --display-name="PR ${RH_PULL_REQUEST_ID} - Automated Deployment" > /dev/null 2>&1
 fi
-oc policy add-role-to-user edit ibuziuk -n $PROJECT_NAMESPACE
+oc policy add-role-to-user edit Katka92 ScrewTSW rhopp tomgeorge ibuziuk amisevsk davidfestal skabashnyuk -n $PROJECT_NAMESPACE
 
 start=$(date +%s)
 echo "Running deploy script for che version:${CHE_VERSION}"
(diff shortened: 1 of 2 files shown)


.ci/cico_build_deploy_test_rhche.sh
```diff
@@ -56,7 +56,7 @@ else
   echo "Switching to project failed, probably not exists [$?]. Creating..."
   oc new-project "${PROJECT_NAMESPACE}" --display-name="PR ${RH_PULL_REQUEST_ID} - Automated Deployment" > /dev/null 2>&1
 fi
-oc policy add-role-to-user edit ibuziuk -n $PROJECT_NAMESPACE
+oc policy add-role-to-user edit Katka92 ScrewTSW rhopp tomgeorge ibuziuk amisevsk davidfestal skabashnyuk -n $PROJECT_NAMESPACE
```
ibuziuk: are you sure that all the users are provisioned against the new cluster?
Could you add only Hosted Che team members for now.

.ci/cico_rollout_test.sh
```diff
@@ -114,7 +114,7 @@ oc login "${DEV_CLUSTER_URL}" --insecure-skip-tls-verify \
 echo "setting project"
 oc new-project "${PROJECT}" --display-name="PR ${RH_PULL_REQUEST_ID} - Automated Deployment" > /dev/null 2>&1
 echo "adding roles to project"
-oc policy add-role-to-user edit ibuziuk -n $PROJECT
+oc policy add-role-to-user edit Katka92 ScrewTSW rhopp tomgeorge ibuziuk amisevsk davidfestal skabashnyuk -n $PROJECT
```
ibuziuk: In general, this code should be refactored and list of users should be defined in a separate file to avoid hardcoding the value

ScrewTSW: I could create an issue for myself and deal with it later

.ci/cico_build_deploy_test_rhche.sh
ScrewTSW: I can remove Angel and David, but I would leave Sergii there

ibuziuk: could you add only people that **need** access for now
I would, in general, prefer that anyone who needs access will submit a separate PR. why do you want to add everyone from the previous list if you even do not know if a) you do not know if people are provisioned on the cluster b) people need this access

.ci/cico_rollout_test.sh
ibuziuk: up-to-you

ibuziuk:
Closing to save some resources on osd dev cluster, feel free to open another PR if you need access

PR #1869, opened 2020-04-20
Add video recording to tests

Katka92:
### What does this PR do?
**Adding logic for building a test image.**
Currently, if there is already a test image pushed with the corresponding version, the image is not rebuilt in the PR check flow. Adding logic that will check, if there are some changes made in rh-che/dockerfiles/e2e-saas folder. If so, the test image will not be reused but rebuilt.

**Add video recording to tests.**
In previous versions, there were no videos, but every second the screenshot was taken and a screencast was saved in artifacts. Now, in upstream, they removed that logic and added video recording directly to the docker entrypoint. We need to adapt and to set the same recording in our entrypint.

### DIFF
--- a/dockerfiles/e2e-saas/docker-entrypoint.sh
+++ b/dockerfiles/e2e-saas/docker-entrypoint.sh
@@ -5,6 +5,10 @@ kill_ffmpeg(){
   echo "In method kill_ffmpeg()"
   echo "Killing ffmpeg with PID=$ffmpeg_pid"
   kill -2 "$ffmpeg_pid"
+  set +e
+  wait "$ffmpeg_pid"
+  echo "After a wait, ffmpeg process was successfully killed."  
+  set -e
   return 0
 }
 
@@ -113,6 +117,14 @@ export TS_SELENIUM_BASE_URL=$URL
 export TS_SELENIUM_MULTIUSER=true
 export TS_SELENIUM_LOG_LEVEL=$DEBUG_LEVEL
 
+echo "Running Xvfb"
+
+export DISPLAY=:20
+export SCREEN="0"
+
+/usr/bin/Xvfb $DISPLAY -screen $SCREEN 1920x1080x16 +extension RANDR > /dev/null 2>&1 &
+x11vnc -display $DISPLAY -N -forever > /dev/null 2>&1 &
+
 # Launch selenium server
 /usr/bin/supervisord --configuration /etc/supervisord.conf & \
 export TS_SELENIUM_REMOTE_DRIVER_URL=http://localhost:4444/wd/hub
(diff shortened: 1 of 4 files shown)


Katka92:
[test]

Katka92:
[test]

Katka92:
[test]

Katka92:
[test]

functional-tests/devscripts/run_tests.sh
```diff
@@ -97,7 +97,7 @@ if [[ "$PR_CHECK_BUILD" == "true" ]]; then
   #check version in rh-che pom.xml to use correct version of tests.
   getMavenVersion # Checking the maven version for debugging https://github.com/redhat-developer/rh-che/issues/1716
   version=$(getVersionFromPom)
-  if [[ -z "${version}" ]]; then
+  if [[ -z "${version}" || $version = "" ]]; then
```
ScrewTSW, line 100: Is this really a necessary change?

dockerfiles/e2e-saas/docker-entrypoint.sh
```diff
@@ -165,5 +174,17 @@ fi
 echo "Installing upstream dependency."
 npm --silent i ../../e2e/
 
+# create a folder and record a video
+ffmpeg_path="./report/ffmpeg_report"
```
ibuziuk, line 178: would be nice to extract this into the meaningful functions e.g. strartRecording / stopRecording

functional-tests/devscripts/run_tests.sh
ibuziuk: `if [ -z "${version}"]; then` should do the trick

Katka92: Tests are failing if the version is empty. It's probably not caught by the `-z` switch.

Katka92: I see, I'll remove it.

ibuziuk: you need to have a single `[` in the if statement

dockerfiles/e2e-saas/docker-entrypoint.sh
Katka92: Done.

functional-tests/devscripts/run_tests.sh
Katka92: Done.

Katka92:
@ibuziuk @ScrewTSW I've added changes you suggested. You can see the video here [1]. Please note that first ~30 seconds there is a black screen - it is expected as a test need some time to prepare all the stuff (e.g. build the code). 

[1] http://artifacts.ci.centos.org/devtools/rhche/devtools-rh-che-rh-che-prcheck-dev.rdu2c.fabric8.io/2550/e2e_report/report/ffmpeg_report/output.mp4

PR #1883, opened 2020-05-05
Add pre-release test suite

Katka92:
### What does this PR do?
Adding variable ACCOUNT_ENV to docker entrypoint. Adding a pre-release test suite.

https://github.com/eclipse/che/issues/16549

Localy.


dockerfiles/e2e-saas/docker-entrypoint.sh
```diff
@@ -59,13 +59,23 @@ length=${#USERNAME}
...
+  else
+    preview=""
+  fi
+else 
+  if [ "$ACCOUNT_ENV" == "prod" ]; then
+    preview=""
+  else 
+    preview="prod-preview."
+  fi
 fi
```
ibuziuk, line 79: this logic is not straight forward for me. I believe we can do a better job of determining staging vs prod environment + would be nice to extract it to a separate function

Katka92: I can definitely separate that to a function. Do you have any idea how to improve this logic? I'm not aware of any other check.

ibuziuk: first `ACCOUNT_ENV variable is not set, retrieving env from URL` is it really true? according to the logic preview vs prod is retrieved based on the username?

dockerfiles/e2e-saas/docker-entrypoint.sh
```diff
@@ -59,13 +59,23 @@ length=${#USERNAME}
...
 
 response=$(curl -s -g -X GET --header 'Accept: application/json' "https://api.${preview}openshift.io/api/users?filter[username]=$USERNAME")
```
ibuziuk, line 81: in general, I'm not a big fan of the manipulations with `${preview}`

ibuziuk: I believe we can simply determine URLs based on the `ACCOUNT_ENV` and avoid using extra preview env var

dockerfiles/e2e-saas/docker-entrypoint.sh
Katka92: Yeah, it is retrieved from username. I'll fix that. Theoretically, we can remove that logic and just go with the mandatory variable `ACCOUNT_ENV` that would specify the env. But that would require changes in more scripts. If you want, I can do that as a part of this PR.

PR #1890, opened 2020-05-12
adapt to upstream changes when docker images are used instead of maven artifact to deliver che-dashboard && che-workspace-loader

skabashnyuk:
### What does this PR do?
adapt to upstream changes when docker images are used instead of maven artifact to deliver che-dashboard && che-workspace-loader

https://github.com/eclipse/che/issues/16799


ibuziuk:
@Katka92 could the test failure be related to the recent UD changes - https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-ide-dashboard-war/src/assets/branding/product.json#L16 ?

dockerfiles/che-fabric8/Dockerfile
```diff
@@ -3,6 +3,12 @@
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
+ARG CHE_DASHBOARD_VERSION=next
+ARG CHE_WORKSPACE_LOADER_VERSION=next
```
ibuziuk, line 7: I believe we should not fall back on the next version in future, but since this approach is used in the upstream this should be ok for now

ibuziuk:
@Katka92 could you please take a look at the failing tests, is it related to the discussion we had yesterday regarding  https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-ide-dashboard-war/src/assets/branding/product.json#L16 ?

Katka92:
@ibuziuk Yes, it seems that's the problem we discussed. The `new workspace page` is not like it used to be, but switched to `Get Started` page with two tabs. I would like to discuss that on today's meeting.

Katka92:
@skabashnyuk I wrote a fix but I can not push to your branch. I verified, that it works here: https://github.com/redhat-developer/rh-che/pull/1895. Could you please merge your PR to `upstream-check-7.13.0-SNAPSHOT`? I'll merge my fix for tests after that. Thank you!

PR #1898, opened 2020-05-25
Installing latest version of Docker during the CI build. Scripts Refactoring

ibuziuk:
### What does this PR do?
Installing latest version of Docker during the CI build. Scripts Refactoring

N/A

PR check CI

### DIFF
--- a/.ci/cico_build.sh
+++ b/.ci/cico_build.sh
@@ -30,8 +30,12 @@ if [[ "$DeveloperBuild" != "true" && "$PR_CHECK_BUILD" != "true" && "$USE_CHE_LA
 
   set -x
   yum -y update
-  yum -y install centos-release-scl java-1.8.0-openjdk-devel git patch bzip2 golang docker
+  yum -y install centos-release-scl java-1.8.0-openjdk-devel git patch bzip2 golang
   yum -y install rh-maven33 rh-nodejs8
+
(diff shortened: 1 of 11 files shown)


.ci/cico_build.sh
```diff
@@ -30,9 +30,13 @@ if [[ "$DeveloperBuild" != "true" && "$PR_CHECK_BUILD" != "true" && "$USE_CHE_LA
 
   set -x
   yum -y update
-  yum -y install centos-release-scl java-1.8.0-openjdk-devel git patch bzip2 golang docker
+  yum -y install centos-release-scl java-1.8.0-openjdk-devel git patch bzip2 golang
```
Katka92, line 33: This command can be replaced by `yum -y install java-1.8.0-openjdk-devel patch bzip2 golang`. The `patch` and `golang` is used in cico_utils.sh too, but it does not have its own method. Maybe it would be worth creating a method for that in cico_utils.sh and reusing here?

and using those methods from cico_utils.sh:
```
installScl
installGit
```

ibuziuk: yeah, I believe it is a good idea for further refactoring ;)
I just scratched the surface in this PR

Katka92: Ok :)

PR #1901, opened 2020-05-25
Remove build of functional test image

Katka92:
### What does this PR do?
Exclude the building of functional tests image from build job. This job is triggered when the master build succeeds and builds all test images.

Currently, the script is broken.
https://ci.centos.org/job/devtools-rh-che-build-func-dep/183/consoleFull


Katka92:
Merging without waiting for tests to pass - tests are not touching the script that is modified in this PR.

PR #1958, opened 2020-07-30
Adding sum up of all QE jobs and images

Katka92:
### What does this PR do?
Sum up all jobs and images that Hosted QE is using.

https://github.com/eclipse/che/issues/17349


Katka92:
@ibuziuk I've investigated why we didn't see the rollout project. The flow is the following:
 - if the test fails, the rollout project remains there to allow QE&Devs to investigate
 - if the test passes, it removes the project to save the resources

PR #1982, opened 2020-09-07
Set default value for storage type to 'ephemeral'

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?
Set default value for storage type to 'ephemeral'


openshift/rh-che.config.yaml
```diff
@@ -39,6 +39,7 @@ data:
   CHE_SYSTEM_SUPER__PRIVILEGED__MODE: "true"
   CHE_WEBSOCKET_ENDPOINT: "wss://che.prod-preview.openshift.io/api/websocket"
   CHE_WEBSOCKET_ENDPOINT__MINOR: "wss://che.prod-preview.openshift.io/api/websocket-minor"
+  CHE_WORKSPACE_STORAGE_PREFERRED__TYPE: "ephemeral"
```
sleshchenko, line 42: Should it replace https://github.com/redhat-developer/rh-che/pull/1982/files#diff-89824f4a1e5e379f4f8378527ac0aab8R44 ?

vparfonov: I think so, but need to check with @akurinnoy

ibuziuk: yeah, as I understand `CHE_WORKSPACE_PERSIST__VOLUMES_DEFAULT` is deprecated and does not work anymore (worth removing it from upstream also)

akurinnoy: `CHE_WORKSPACE_PERSIST__VOLUMES_DEFAULT` can be safely removed as UD doesn't rely on it anymore.

vparfonov: PR updated

ibuziuk (approved):
@vparfonov thanks, I also sent MR for updating values on prod-preview cm. So, let's merge it once cm on the preview is applied in order to trigger the rollout update

PR #1985, opened 2020-09-08
Adding info about preserving jobs for Toolchain switch

Katka92:
### What does this PR do?
Add column to the README to indicate if we plan to move the job to Toolchain.

https://github.com/eclipse/che/issues/17516


Katka92:
@ibuziuk could you please take a look at the jobs - would you like to preserve some more jobs than I selected?

Katka92:
Rh-che was not deployed on the dev cluster. Reruning tests.
[test]

ibuziuk (approved):

PR #1997, opened 2020-10-09
Removed unnecessary data from product.json

akurinnoy:
Signed-off-by: Oleksii Kurinnyi <okurinny@redhat.com>

### What does this PR do?

In accordance to https://github.com/eclipse/che/issues/17741 `product.json` should contain only values that will should overwrite default branding data. This PR gets rid of those fields in `product.json` that are the same as branding defaults in upstream.


sleshchenko (approved):
Happy to see that we have less to maintain

ibuziuk (approved):

PR #2032, opened 2021-01-12
che #18770 Updating Hosted Che to the latest '7.24.1' upstream version

ibuziuk:
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?
Updating Hosted Che to the latest '7.24.1' upstream version

### What issues does this PR fix or reference?
Related issue - https://github.com/eclipse/che/issues/18770

### How have you tested this PR?
PR check CI

PR #2034, opened 2021-01-12
Stabilize setting of a secret in PR check

Katka92:
### What does this PR do?
The deployment of https://github.com/redhat-developer/rh-che/pull/2032 does not have all secrets set up. Debugging the flow.

### DIFF
--- a/dev-scripts/deploy_custom_rh-che.sh
+++ b/dev-scripts/deploy_custom_rh-che.sh
@@ -453,8 +453,7 @@ fi
 sleep 3
 echo "checking if the pod has quay-dev-deployer secret"
 pod=$(oc get pod -l app=rhche -o name)
-secret_contain=$(oc get -o json ${pod} | jq .spec.imagePullSecrets[] | grep quay-dev-deployer )
-if [ ${#secret_contain} > 0 ]; then
+if oc get -o json ${pod} | jq .spec.imagePullSecrets[] | grep -q quay-dev-deployer; then
   echo "Secret is present in the rhche pod, there should not be problem with pulling images."
 else
   echo "Secret is not present in the rhche pod, removing the pod"
--- a/dev-scripts/deploy_custom_rh-che.sh
+++ b/dev-scripts/deploy_custom_rh-che.sh
@@ -441,11 +441,24 @@ if ! (echo "$CHE_APP_CONFIG_YAML" | oc process -f - | oc apply -f - > /dev/null 
   exit 5
 fi
 
+
 # APPLY SECRET IF REQUESTED
 if [ "$RH_CHE_APPLY_SECRET" == "true" ]; then
   echo "$RH_CHE_OC_SECRET" | oc apply -f -
   oc secrets link default quay-dev-deployer --for=pull
   oc secrets link rhche quay-dev-deployer --for=pull
+fi
+
+# CHECK IF THE SECRET IS SET PROPERLY, RE-SET IF NEEDED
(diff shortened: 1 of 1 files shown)


Katka92:
[test]

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -441,13 +441,26 @@ if ! (echo "$CHE_APP_CONFIG_YAML" | oc process -f - | oc apply -f - > /dev/null
...
+# CHECK IF THE SECRET IS SET PROPERLY, RE-SET IF NEEDED
+sleep 3
+echo "checking if the pod has quay-dev-deployer secret"
+pod=$(oc get pod -l app=rhche -o name)
+secret_contain=$(oc get -o json ${pod} | jq .spec.imagePullSecrets[] | grep quay-dev-deployer )
+if [ ${#secret_contain} > 0 ]; then
+  echo "Secret is present in the rhche pod, there should not be problem with pulling images."
+else
+  echo "Secret is not present in the rhche pod, removing the pod"
+  oc delete $(oc get pod -l app=rhche -o name)
```
ibuziuk, line 461: should we delete pod or dc here? or the idea is that pod will be re-created automatically?

Katka92: Yes, the idea is just to re-create pod. From how I understand the flow, the dc should be configured fine.

Katka92: But it's hard to test that as this issue is happening just from time to time.

dev-scripts/deploy_custom_rh-che.sh
```diff
@@ -441,13 +441,26 @@ if ! (echo "$CHE_APP_CONFIG_YAML" | oc process -f - | oc apply -f - > /dev/null
...
   echo "$RH_CHE_OC_SECRET" | oc apply -f -
   oc secrets link default quay-dev-deployer --for=pull
   oc secrets link rhche quay-dev-deployer --for=pull
 fi
 
+# CHECK IF THE SECRET IS SET PROPERLY, RE-SET IF NEEDED
+sleep 3
+echo "checking if the pod has quay-dev-deployer secret"
+pod=$(oc get pod -l app=rhche -o name)
+secret_contain=$(oc get -o json ${pod} | jq .spec.imagePullSecrets[] | grep quay-dev-deployer )
```
ScrewTSW, line 456: This can be simplified to
```
if oc get -o json ${pod} | jq .spec.imagePullSecrets[] | grep -q quay-dev-deployer; then
```

dev-scripts/deploy_custom_rh-che.sh
ibuziuk: since it is hard to verify properly, I would postpone merging the PR unless we would have issues with https://github.com/redhat-developer/rh-che/pull/2039

Katka92:
Heh, haven't seen this issue during a test run for a while.
image
[test]

dev-scripts/deploy_custom_rh-che.sh
Katka92: @ibuziuk This should fix the issue. If the issue is not appearing, the workaround will just not be triggered. If the problem appears, it will redeploy the pod and fix the issue. I don't see any harm that could be caused by merging that now - it can only bring benefits if the issue will come back again.

ibuziuk (approved):

